Issue #324, opened 2017-07-21
Proxy Policy: add more load balancing strategies


kevinswiber:
Several options are available for load balancing algorithms.

- Round robin
- Weighted round robin
- Random
- Least connections
- Custom

Extra credit: Consistent Hashing with Bounded Loads (https://arxiv.org/pdf/1608.01350.pdf)

kevinswiber:
This will give birth to new feature requests, as well: health checks, dynamic routing, service discovery, etc.

altsang:
added to feathub - http://feathub.com/ExpressGateway/express-gateway/+30

PR #326, opened 2017-07-21
Adding basic round-robin load balancer

kevinswiber:
This is mostly here as a strawman to start discussions regarding load balancing.  See #324 for more details.

### DIFF
--- a/lib/policies/proxy/round-robin.js
+++ b/lib/policies/proxy/round-robin.js
@@ -1,0 +1,30 @@
+const logger = require('../../logger').gateway;
+
+module.exports = class RoundRobin {
+  constructor (proxy, endpoints) {
+    this._proxy = proxy;
+
+    this.endpoints = endpoints;
+    this.endpointIndex = 0;
+    this.endpointMaxIndex = this.endpoints.length - 1;
+  }
+
+  proxy (req, res) {
+    const options = {
+      target: this.nextTarget()
+    };
+
+    logger.debug(`proxying to ${options.target}, ${req.method} ${req.url}`);
+    return this._proxy.web(req, res, options);
+  }
+
+  nextTarget () {
+    const target = this.endpoints[this.endpointIndex++];
+
+    if (this.endpointIndex > this.endpointMaxIndex) {
+      this.endpointIndex = 0;
+    }
+
+    return target;
+  }
+};
--- a/lib/policies/proxy/index.js
+++ b/lib/policies/proxy/index.js
@@ -1,5 +1,3 @@
-'use strict';
-
 module.exports = {
   policy: require('./proxy')
 };
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -1,27 +1,44 @@
-'use strict';
-
 const httpProxy = require('http-proxy');
-const lodash = require('lodash');
-const logger = require('../../logger').gateway;
 
 const ConfigurationError = require('../../errors').ConfigurationError;
+const logger = require('../../logger').gateway;
+const RoundRobin = require('./round-robin');
+
+const strategies = {
+  'round-robin': RoundRobin
+};
+
+const createStrategy = (strategy, proxy, endpointUrls) => {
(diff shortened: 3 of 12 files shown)


kevinswiber:
@geek You might be interested in this, too.

lib/policies/load-balancer/balancer.js
```diff
@@ -0,0 +1,55 @@
...
+
+const createStrategy = (name, proxy, endpoints) => {
+  const Strategy = strategies[name];
+  return new Strategy(proxy, endpoints);
+};
+
+module.exports = function (params, config) {
+  const endpoints = params.serviceEndpoints
+    .map(name => {
+      const endpoint = config.gatewayConfig.serviceEndpoints[name];
```
geek: Another option would be that any config endpoint with `urls` instead of `url` gets round-robin treatment

```
serviceEndpoints:
  frontend:
    urls:
      - 'http://server1'
      - 'http://server2'
````

Comment (DrMegavolt) [inline lib/policies/load-balancer/balancer.js]: I would say this should be only in a way @geek is suggesting. Because next time when we want to introduce consul\etcd support we will just define some `key` for service endpoint that will be used for lookup. this key will be resolved to  `urls` array.

Comment (kevinswiber) [inline lib/policies/load-balancer/balancer.js]: Should we also add a `method` property (or whatever) to the serviceEndpoint schema?

```
serviceEndpoints:
  frontend:
    urls:
      - 'http://server1'
      - 'http://server2'
    method: 'least-conn'
```

Then, I assume, our proxy policy can wire everything up behind the scenes.

Comment (DrMegavolt) [inline lib/policies/load-balancer/balancer.js]: yes, it should be good. 

HAProxy uses `algorithm`
http://cbonte.github.io/haproxy-dconv/configuration-1.4.html#4.2-balance

`strategy` or `algorithm` seems to be good. `method` is confusing with `HTTP method`

**Comment by kevinswiber** [comment]
@DrMegavolt @geek 

Latest changes support the following config:

```yaml
http:
  port: 8080
admin:
  port: 9876
  hostname: localhost
apiEndpoints:
  api:
    host: localhost
    paths: '/ip'
serviceEndpoints:
  httpbin:
    strategy: 'round-robin'
    urls:
      - 'https://service1'
      - 'https://service2'
policies:
  - proxy
pipelines:
  - name: default
    apiEndpoints:
      - api
    policies:
      - proxy:
          - action:
              serviceEndpoint: httpbin 
              changeOrigin: true
```

The `strategy` property is optional in the `serviceEndpoint` schema.  It defaults to `round-robin`.

lib/policies/proxy/round-robin.js
```diff
@@ -0,0 +1,30 @@
+const logger = require('../../logger').gateway;
+
+module.exports = class RoundRobin {
+  constructor (proxy, endpoints) {
```
DrMegavolt: For this specific strategy I believe proxy in Strategy is not needed. 
having `nextTarget` interface should be enough 
same for 'sticky' type

is proxy here for future least-conn type to count requests\responses?

kevinswiber: Hmm.  Don't know.  Just felt right to delegate proxying responsibility to the load balancer, so it can make whatever decisions it needs to or return an error of some kind if it must without bouncing it back to the caller (proxy.js in this case).

kevinswiber:
@DrMegavolt Maybe one more peek here before merging.

kevinswiber:
Latest round of changes:

```yaml
http:
  port: 8080
admin:
  port: 9876
  hostname: localhost
apiEndpoints:
  api:
    host: localhost
    paths: '/ip'
serviceEndpoints:
  httpbin:
    urls:
      - 'https://service1'
      - 'https://service2'
policies:
  - proxy
pipelines:
  - name: default
    apiEndpoints:
      - api
    policies:
      - proxy:
          - action:
              serviceEndpoint: httpbin 
              strategy: 'round-robin'
              changeOrigin: true
```

DrMegavolt:
@kevinswiber 👍 go

PR #329, opened 2017-07-23
corrected coveralls

altsang:


altsang:
@DrMegavolt  - when you get a chance, fixed the issue with Coveralls, please ignore appveyor had the integration turned on before finishing it
cc: @IrfanBaqui

Issue #335, opened 2017-07-25
Proxy Policy: expose more options from http-proxy.


kevinswiber:
See here: https://github.com/nodejitsu/node-http-proxy#options

Specifically, I need `xfwd`, but a number of these might be useful.

altsang:
added to feathub - http://feathub.com/ExpressGateway/express-gateway/+28
closing here

kevinswiber:
Reopening.

PR #337, opened 2017-07-25
Feature/ws simple PoC 

DrMegavolt:
Ignores all pipelines\policies since there is no way to make them work with ws

### DIFF
--- a/lib/config/gateway.config.yml
+++ b/lib/config/gateway.config.yml
@@ -6,6 +6,11 @@ https:
 admin:
   port: 9876
   hostname: localhost
+
+ws:
+  target: 'ws://localhost:10017'
+
+
 apiEndpoints:
   api:
     host: '*'
--- a/lib/gateway/index.js
+++ b/lib/gateway/index.js
@@ -1,6 +1,7 @@
 const express = require('express');
 const log = require('../logger').gateway;
 const servers = require('./server');
+const wsServers = require('./ws-server');
 const pipelines = require('./pipelines');
 let config = require('../config');
 let policies = require('../policies');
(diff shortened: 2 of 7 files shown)


kevinswiber:
The WebSocket server, in the majority of cases, is going to need to be linked to an HTTP(S) server.  I think we can add options on the `http` and `https` objects in config, `enableWebSockets` or something.

For the implementation, we should just need to plug into the proxy policy.  It's a little sloppy, but here's how I accomplished that in a hurried way yesterday: https://github.com/ExpressGateway/express-gateway/blob/e16b2cd43059af21afc146dc6373c5a1dd211425/lib/policies/proxy/proxy.js

DrMegavolt:
I link it to http and https servers. 
I would imaging it to be ws + wss sections in gateway.config and the config itself is just `http-proxy` parameters like target.  The only reason to have separate section is that it can be many options
https://www.npmjs.com/package/http-proxy#options



I saw you implementation. Main point for me is using pipelines. I would expect other policies to work. And they will not. Having it separate and independent  of pipelines make sense for me. 

and even for load balancing implementation for proxy ws is similar but different 
https://github.com/nodejitsu/node-http-proxy/blob/master/examples/balancer/simple-balancer-with-websockets.js

kevinswiber:
They are linked to the HTTP/HTTPS servers, but not in the config.

Why not treat this like a normal `serviceEndpoint` instead of a special `ws` config?

There are plenty of proxy options we should expose to config, but that's #335.  Does it differ for WebSockets?

kevinswiber:
We should be able to process a pipeline inside an upgrade handler, right?  I think the pipeline bootstrap returns the router.  We'll need to make sure it gets wired up on config changes, as well.

PR #350, opened 2017-07-28
CLI: header and verbose support

DrMegavolt:


bin/eg-generator.js
```diff
@@ -8,7 +8,14 @@ module.exports = class EgGenerator extends Generator {
     this._configuration = null;
     this.eg = this.env.eg;
     this.argv = this.env.argv;
-    this.admin = require('../lib/admin')(config.gatewayConfig.admin);
+    if (this.argv) {
+      let cliConfig = config.systemConfig.cli || {};
```
kevinswiber: Could we look for something like an `EG_ADMIN_URL` environment variable and fall-back to `config.gatewayConfig.admin` if it doesn't exist?

DrMegavolt: I'll add EG_ADMIN_URL, but we should not add `config.gatewayConfig.admin`
'cuz config may be different, for example if admin API is protected by EG.
Also `config.cli` has additional property `protocol`, that doesn't make sense for REST admin

kevinswiber: Okie dokie.  :+1:

lib/admin/index.js
```diff
@@ -1,13 +1,15 @@
 // TODO: this is EG admin client; internal utility
-module.exports = function (adminConfig) {
-  let host = adminConfig.hostname || 'localhost';
-  let port = adminConfig.port || 9876;
-  adminConfig.baseUrl = adminConfig.baseUrl || `http://${host}:${port}/`;
+module.exports = function ({cliConfig, headers, verbose}) {
+  let host = cliConfig.hostname || 'localhost';
+  let port = cliConfig.port || 9876;
+  let protocol = cliConfig.protocol || 'http';
+  let baseUrl = cliConfig.baseUrl || `${protocol}://${host}:${port}/`;
```
kevinswiber: Could we just do a single `url` parameter in the `cliConfig` instead of `protocol`, `host`, and `port`?

DrMegavolt:
@kevinswiber
 Lookup priority
1) EG_ADMIN_URL 
2) `cli.url`
3) http://localhost:9876 default

Issue #217, opened 2017-06-27
Core: rm bluebird dependency


DrMegavolt:
it has very big uninformative stack traces that complicate debugging. 
 custom 
and it is not possible to switch to another lib once start using. a lot of wrappers around standard then\catch

PR #358, opened 2017-08-01
rm bluebird that caused significant perfromance degradation

DrMegavolt:
It is quite strange to have raw proxy perf of 1300 rps and significant drop to 200 if key-auth is enabled 
after spending some time in chrome debugger it appears that most of the time is spend in bluebird deepest and darkest stacktraces

before 200rps
screen shot 2017-07-30 at 1 20 22 pm

after 800 rps
screen shot 2017-08-01 at 12 17 34 pm

just removing redis bluebird promise wrap it is possible to make performance great again :)
How to:

`node --inspect lib/index.js`

connect chrome (type inspect in address bar) and on the inspect page select your instance to debug

start CPU profile capture
```bash
for i in $(eval echo {1..100})
 do
curl http://localhost:8080/users -H "Authorization: apikey 2JWGGh1vNC0DPvcyQ8EbRb:2lPw6PoplLteXU8xYJZmRp"
done && wait
```
review profile. In my case it was 60ms overall promise chain, and only 14ms was in the user service
strategy\key-auth was taken 2-3ms


sachingill:
I don't have full perspective about the problem, however,  starting Node version  6.11.1 , promise native support is provided by Nodejs itself.  Can't we do away with third party promise libraries?

DrMegavolt:
@sachingill yes  
https://github.com/ExpressGateway/express-gateway/issues/217
it is almost done. except for tests and this PR,  there is only one place left that has no trivial drop in replacement (usage of Promise.some ) other places can easily be converted to standard promises

lib/db.js
```diff
@@ -1,7 +1,8 @@
-
-let Promise = require('bluebird');
 const logger = require('./logger').db;
 let config = require('./config');
+const redisCommands = require('redis-commands');
```
kevinswiber: Instead of adding a new dependency, can we just add this as a constant?

I ran this command:

```shell
curl -s https://raw.githubusercontent.com/NodeRedis/redis-commands/master/commands.json | json -M -a key | awk -v ORS=", " -v q="\"" '{ print q $1 q }' | sed 's/, $//'| awk '{ print "[",$0,"]" }' | json
```
It generates a JSON array that we can then shove into a JavaScript module.

```js
const REDIS_COMMANDS = [
  "append",
  "asking",
  "auth",
  "bgrewriteaof",
  "bgsave",
  "bitcount",
  "bitfield",
  "bitop",
  "bitpos",
  "blpop",
  "brpop",
  "brpoplpush",
  "client",
  "cluster",
  "command",
  "config",
  "dbsize",
  "debug",
  "decr",
  "decrby",
  "del",
  "discard",
  "dump",
  "echo",
  "eval",
  "evalsha",
  "exec",
  "exists",
  "expire",
  "expireat",
  "flushall",
  "flushdb",
  "geoadd",
  "geodist",
  "geohash",
  "geopos",
  "georadius",
  "georadiusbymember",
  "get",
  "getbit",
  "getrange",
  "getset",
  "hdel",
  "hexists",
  "hget",
  "hgetall",
  "hincrby",
  "hincrbyfloat",
  "hkeys",
  "hlen",
  "hmget",
  "hmset",
  "host:",
  "hscan",
  "hset",
  "hsetnx",
  "hstrlen",
  "hvals",
  "incr",
  "incrby",
  "incrbyfloat",
  "info",
  "keys",
  "lastsave",
  "latency",
  "lindex",
  "linsert",
  "llen",
  "lpop",
  "lpush",
  "lpushx",
  "lrange",
  "lrem",
  "lset",
  "ltrim",
  "memory",
  "mget",
  "migrate",
  "module",
  "monitor",
  "move",
  "mset",
  "msetnx",
  "multi",
  "object",
  "persist",
  "pexpire",
  "pexpireat",
  "pfadd",
  "pfcount",
  "pfdebug",
  "pfmerge",
  "pfselftest",
  "ping",
  "post",
  "psetex",
  "psubscribe",
  "psync",
  "pttl",
  "publish",
  "pubsub",
  "punsubscribe",
  "quit",
  "randomkey",
  "readonly",
  "readwrite",
  "rename",
  "renamenx",
  "replconf",
  "restore",
  "restore-asking",
  "role",
  "rpop",
  "rpoplpush",
  "rpush",
  "rpushx",
  "sadd",
  "save",
  "scan",
  "scard",
  "script",
  "sdiff",
  "sdiffstore",
  "select",
  "set",
  "setbit",
  "setex",
  "setnx",
  "setrange",
  "shutdown",
  "sinter",
  "sinterstore",
  "sismember",
  "slaveof",
  "slowlog",
  "smembers",
  "smove",
  "sort",
  "spop",
  "srandmember",
  "srem",
  "sscan",
  "strlen",
  "subscribe",
  "substr",
  "sunion",
  "sunionstore",
  "swapdb",
  "sync",
  "time",
  "touch",
  "ttl",
  "type",
  "unlink",
  "unsubscribe",
  "unwatch",
  "wait",
  "watch",
  "zadd",
  "zcard",
  "zcount",
  "zincrby",
  "zinterstore",
  "zlexcount",
  "zrange",
  "zrangebylex",
  "zrangebyscore",
  "zrank",
  "zrem",
  "zremrangebylex",
  "zremrangebyrank",
  "zremrangebyscore",
  "zrevrange",
  "zrevrangebylex",
  "zrevrangebyscore",
  "zrevrank",
  "zscan",
  "zscore",
  "zunionstore"
];

Object.freeze(REDIS_COMMANDS)

module.exports = REDIS_COMMANDS;
```

DrMegavolt: @kevinswiber  we already have this dependency, it is not some random module. it is used inside redis js client 
https://github.com/NodeRedis/node_redis/blob/master/lib/commands.js#L3

if we copy this json into the project we will have headache maintaining list up to date with new versions of node-redis

sachingill:
I agree with @DrMegavolt . Also there is no point storing redis commands internally, as in future Redis can be replaced with some other db.

lib/db.js
IrfanBaqui: I agree with having it as a dependency. Constants would require maintenance.

IrfanBaqui:
LGTM

lib/db.js
kevinswiber: Approving since we've already got a higher level dependency.  👍

Issue #367, opened 2017-08-04
https - error in capturing key/cert file path


larryschirmer:
When opening an https connection I noticed that express-gateway/lib/gateway/server.js is having issues properly capturing the file path in the .yml. 

I made some minor changes to `createTlsServer()` which seem to have resolved the issue that I would like to propose here:

```javascript
function createTlsServer(httpsConfig, app) {
	let defaultCert = null;
	let sniCerts = [];

	let domainCount = httpsConfig.tls.length;
	let domains = [];
	httpsConfig.tls.forEach(domainObj => {
		domains = [...domains, Object.getOwnPropertyNames(domainObj)];
	});

	for (let i = 0; i < domainCount; i++) {
		let domain = domains[i].toString();

		let certPaths = httpsConfig.tls[i][domain];

		let cert;
		if (domain === 'default') {
			cert = defaultCert = {};
		} else {
			cert = {};
			sniCerts.push([domain, cert]);
		}

		cert.key = fs.readFileSync(path.resolve(certPaths.key), 'utf-8');
		cert.cert = fs.readFileSync(path.resolve(certPaths.cert), 'utf-8');
		if (certPaths.ca && certPaths.ca.length) {
			cert.ca = certPaths.ca.map(ca => fs.readFileSync(path.resolve(ca), 'utf-8'));
		}
	}

	// see possible options https://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener
	let options = Object.assign({}, httpsConfig.options);

	if (defaultCert) {
		options.key = defaultCert.key;
		options.cert = defaultCert.cert;
		options.ca = defaultCert.ca;
	}

	if (sniCerts.length > 0) {
		options.SNICallback = (servername, cb) => {
			for (let [domain, cert] of sniCerts) {
				if (minimatch(servername, domain)) {
					logger.debug(`sni: using cert for ${domain}`);
					cb(null, tls.createSecureContext(cert));
					return;
				}
			}
			if (defaultCert) {
				logger.debug('sni: using default cert');
				cb(null, tls.createSecureContext(defaultCert));
			} else {
				logger.error('sni: no cert!');
				cb(new Error('cannot start TLS SNI - no cert configured'));
			}
		};
	}

	return https.createServer(options, app);
}
```

kevinswiber:
Thanks, @larryschirmer!  Can you turn this into a Pull Request?

larryschirmer:
@kevinswiber 
[...]

PR #368, opened 2017-08-05
Fix error in capturing key/cert file path - connects to #367

larryschirmer:
When opening an https connection I noticed that express-gateway/lib/gateway/server.js is having issues properly capturing the file path in the .yml.

The for in loop is trying to pull properties from `httpsConfig.tls` when the object is defined as an array

```javascript
//httpsConfig
{
    "port": 9443,
    "tls": [
        {
            "localhost": {
                "key": "localhost.key.pem",
                "cert": "localhost.cert.pem"
            }
        },
        {
            "default": {
                "key": "localhost.key.pem",
                "cert": "localhost.cert.pem"
            }
        }
    ]
}
``` 

The issue was resolved by looping through the `tls` array and assigning each domain object to `certPaths`.

### DIFF
--- a/lib/gateway/server.js
+++ b/lib/gateway/server.js
@@ -19,10 +19,10 @@ module.exports.bootstrap = function (app) {
 function createTlsServer (httpsConfig, app) {
   let defaultCert = null;
   let sniCerts = [];
-  
+
   let domainCount = httpsConfig.tls.length;
   let domainNames = [];
-	
+
   httpsConfig.tls.forEach(domainObj => {
     domainNames = [...domainNames, Object.getOwnPropertyNames(domainObj)];
   });
--- a/lib/gateway/server.js
+++ b/lib/gateway/server.js
@@ -19,10 +19,17 @@ module.exports.bootstrap = function (app) {
 function createTlsServer (httpsConfig, app) {
   let defaultCert = null;
   let sniCerts = [];
+  
+  let domainCount = httpsConfig.tls.length;
+  let domainNames = [];
+	
+  httpsConfig.tls.forEach(domainObj => {
+    domainNames = [...domainNames, Object.getOwnPropertyNames(domainObj)];
+  });
 
-  for (let el in httpsConfig.tls) {
-    let domain = el;
-    let certPaths = httpsConfig.tls[el];
+  for (let i = 0; i < domainCount; i++) {
+    let domain = domainNames[i].toString();
+    let certPaths = httpsConfig.tls[i][domain];
     let cert;
     if (domain === 'default') {
       cert = defaultCert = {};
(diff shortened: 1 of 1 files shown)


lib/gateway/server.js
```diff
@@ -20,9 +20,16 @@ function createTlsServer (httpsConfig, app) {
   let defaultCert = null;
   let sniCerts = [];
 
-  for (let el in httpsConfig.tls) {
-    let domain = el;
-    let certPaths = httpsConfig.tls[el];
+  let domainCount = httpsConfig.tls.length;
+  let domainNames = [];
+
+  httpsConfig.tls.forEach(domainObj => {
```
kevinswiber: Can this be collapsed to something like:

```js
const domainNames = httpsConfig.tls.map(domainObj => Object.getOwnPropertyNames(domainObj));
```

@DrMegavolt I think you worked on this in the past.  What behavior should we expect if no `tls` configuration exists?  Should we log a warning?  Exit with a config validation error?  Thoughts?

larryschirmer: @kevinswiber great code condensing suggestion. 

CircleCI is showing quite a few errors and warnings with my pull request. Is this all caused by a disruption from this change?

Also, if no `tls` option is set in `gateway.config.yml` for https i.e.
```yml
https:
  port: 9443
```
then `npm start` silently fails to begin listening on the specified port allowing the process to continue with no error output. 

Should an error be thrown in this file to make a developer aware of the potential bug?

lib/gateway/server.js
```diff
@@ -20,9 +20,14 @@ function createTlsServer (httpsConfig, app) {
   let defaultCert = null;
   let sniCerts = [];
 
-  for (let el in httpsConfig.tls) {
-    let domain = el;
-    let certPaths = httpsConfig.tls[el];
+  let domainCount = httpsConfig.tls.length;
+  const domainNames = httpsConfig.tls
```
DrMegavolt: @kevinswiber @larryschirmer 

this fails in Circle CI if not tls provided in https section. so additional check is needed. 

It seems there is a variant TLS-PSK that does not need a certificate. And it seems to be not supported by node.js https://github.com/nodejs/node/pull/6701

So until then certificates are required. And I would imaging throwing an error and preventing start would be a good option.

larryschirmer:
@DrMegavolt @kevinswiber 
I would like to close this pull request, and replace it with a different pull request that addresses an easier error in the documentation. See latest comment in [issue 367](https://github.com/ExpressGateway/express-gateway/issues/367). Is there additional work that should be preserved from this pr before I close it?

kevinswiber:
@larryschirmer We can close this.  No worries.  Thanks for taking another stab at it!

Issue #345, opened 2017-07-26
add in more e2e integration tests


altsang:
string together - a good covered set of policies and different options, multiple pipelines, to try to get functional coverage of all capabilities

kevinswiber:
e2e tests:

- key-auth
- basic-auth
- oauth2 authorization code grant type

maintenance:

- split into separate npm script (e.g., `npm run test:e2e`)

PR #377, opened 2017-08-14
key-auth+cli e2e test connect #345

DrMegavolt:
connect #345

### DIFF
--- a/test/common/gateway.helper.js
+++ b/test/common/gateway.helper.js
@@ -9,10 +9,14 @@ const { generateBackendServer } =
 let gatewayPort = null;
 let adminPort = null;
 let backendPort = null;
-module.exports.setGatewayConfig = function ({gatewayConfigPath, newConfig}) {
-  fs.writeFileSync(gatewayConfigPath, yaml.dump(newConfig));
+
+// Set gateway.config or system.config yml files
+module.exports.setYmlConfig = function ({ymlConfigPath, newConfig}) {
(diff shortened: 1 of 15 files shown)


test/common/gateway.helper.js
```diff
@@ -0,0 +1,109 @@
...
+        reject(err);
+      });
+    }
+  });
+};
+
+module.exports.startGatewayInstance = function ({dirInfo, gatewayConfig}) {
+  return this.findOpenPortNumbers(4)
+      .then(ports => {
+        console.log(ports);
```
kevinswiber: Should be able to get rid of the console.log statement.

test/common/gateway.helper.js
```diff
@@ -0,0 +1,109 @@
...
+          let gatewayProcess = fork(modulePath, [], {
+            cwd: dirInfo.basePath,
+            env: childEnv
+          });
+
+          gatewayProcess.on('error', err => {
+            reject(err);
+          });
+          let count = 0;
+          let interval = setInterval(() => {
```
kevinswiber: :+1:

kevinswiber:
@DrMegavolt Can you refactor the OAuth2 Authorization Code test to take advantage of the new helpers you built and perhaps throw in the retry counter you added?

DrMegavolt:
@kevinswiber auth-code and hot-reload now uses refactored `find ports` method.
Create Gateway is bit different in each case, probably my helper will need revision once we have more use cases, for now I left original code

Issue #73, opened 2017-05-09
Plugin Framework: install plugin via `npm install` - user manually adds to system.config.yml to plugins: section


IrfanBaqui:

DrMegavolt:
Spec proposal for plugin framework
https://docs.google.com/document/d/1jSDul2n_xbeKNtnek69M79-geur6aTWShAcBZ9evD0E/edit

PR #381, opened 2017-08-23
Feature/plugins v3spec part1 connect #73

DrMegavolt:
Final v1.0 version


lib/plugins.js
```diff
@@ -0,0 +1,68 @@
...
+    cliExtensions: extract(loadedPlugins, 'cliExtensions')
+  };
+};
+
+class PluginManifest {
+  constructor ({settings, config}) {
+    this.logger = logger;
+    this.settings = settings || {};
+    this.config = config;
+    this.policies = [];
```
kevinswiber: Instead of exposing the underlying data structure, what do you think about registration functions?  `manifest.registerPolicy` instead of `manifest.policies.push`.

DrMegavolt: 👍

lib/plugins.js
```diff
@@ -0,0 +1,68 @@
...
+  return {
+    policies: extract(loadedPlugins, 'policies'),
+    conditions: extract(loadedPlugins, 'conditions'),
+    gatewayExtensions: extract(loadedPlugins, 'gatewayExtensions'),
+    adminExtensions: extract(loadedPlugins, 'adminExtensions'),
+    cliExtensions: extract(loadedPlugins, 'cliExtensions')
+  };
+};
+
+class PluginManifest {
```
kevinswiber: I think the concept of "manifest" is a little strange in this context.  I think of a manifest as a file living outside the code that contains metadata about the contents of the project, usually written in a config format like XML, JSON, or INI.  I get that it's serving a similar purpose, but I look at this in code and think, "Gee, is this telling me I can actually put this manifest in a separate config file?"  This might be a bias based on my past experiences.  Not sure.  Also, feels a tad more weird when we start assigning event handlers.  Just thought I'd toss this out there.  Would it make more sense to just call this... `Plugin`?

kevinswiber:
Just to clarify, this PR doesn't include support for the following extensions:

- Custom database entities
- Services (or an internal service registry of any kind)
- Consumer Management (custom auth methods)

Is that correct?

DrMegavolt:
and no CLI extensions as well

lib/plugins.js
DrMegavolt: I agree that it looks weird to common approach like that  
https://developer.chrome.com/extensions/manifest

the thing is that what it passes is not a plugin, because plugin is entire package itself. 
It is something like container/ plugin manager / context. 

I cannot name it pluginOptions because it will be confused with parameters and settings

lib/config/system.config.yml
```diff
@@ -8,6 +8,9 @@ db:
 cli:
   url: http://localhost:9876
 
+plugins:
+  # eg-example-plugin:
```
altsang: `express-gateway-plugin-example` would be the package name, the actual plugin name would be `example` in the `system.config.yml` ?

for example:
`express-gateway-plugin-oauth2` would be the npm module the entry  within `system.config.yml` - `oauth2`

DrMegavolt: ok, make sense. and if user needs to install non convention package they will have 'package' word for that?

kevinswiber: This section basically acts as a whitelist, yeah?

If so, why do we need a whitelist if the package has already been explicitly installed?

DrMegavolt: I like concept installed vs enabled. Same idea as behind policies - have many registered(installed), but use only whitelisted(enabled). 
It seems to be standard pattern for plugins. Because installation is typically complex process and enable\disable is 1 sec operation.

Atom\VS Code\Sublime all have that. Kong - same idea.

DrMegavolt: Even ExpressJS - installed middleware, unless it is used in code doesn't exist

kevinswiber: Ah, right.  So it's not really a "whitelist" of allowed plugins; it's just a signal to enable.  We could have discovery logic for finding a package listed here.

1. Try wrapping the package name in `express-gateway-${name}-plugin`.
2. If that fails, try `${name}`.
3. Error.

DrMegavolt: At this point I've implemented the logic 
read settings for ${name} 
if contains `package` -> require('package')
else  ensure  `name` is prefixed with `express-gateway-plugin-` -> require(name)

So if you really want different name, np use `package` property

kevinswiber: Got it.  Thanks.

lib/plugins.js
```diff
@@ -0,0 +1,79 @@
...
+  registerCondition (condition) {
+    this.conditions.push(condition);
+  }
+  registerGatewayExtension (gatewayExtension) {
+    this.gatewayExtensions.push(gatewayExtension);
+  }
+  registerAdminExtension (adminExtension) {
+    this.adminExtensions.push(adminExtension);
+  }
+  registerCliExtension (cliExtension) {
```
kevinswiber: For acronyms in JavaScript identifiers, I use all capital letters as the convention, taken from some built-ins like `encodeURIComponent`.  Would you be willing to switch this to `registerCLIExtension`?

DrMegavolt:
@altsang @kevinswiber please take a look once again. 

The main difference for last commit it is now possible to provide entire config to `gateway` entity itself. so it will be completely decoupled from `config` module. 

it allows test code without having real config file loaded first
(Still some tests rely on config automatically loaded by gateway, so backward compatibility is preserved - gateway will do config loading if no config object provided)

Another difference is that previously it was a trick to load fake test policies, now it is possible to just use plugins for that. Some tests where refactored with this approach

kevinswiber:
Hmm.  So I'm actually running into an issue trying to load plugins.  The `require` call fails.  Agh!  Trying to figure it out now.

lib/config/system.config.yml
altsang: 👍  sounds good to me, the `plugins:` is so that there is a known declaration and exposure to what global params are in place 

simliarly in `policies:` you can have policies installed but unless you declare it, it can't be referenced in a pipeline

kevinswiber:
Just found this.  I think we should use it.  Looks like it can be an issue for people developing plugins.  Let's circumvent it.

https://www.npmjs.com/package/parent-require

@DrMegavolt

altsang:
Jared's stuff is solid 👍

DrMegavolt:
parent-require is integrated

Issue #401, opened 2017-09-07
rm node-uuid@1.4.8: deprecation warning


DrMegavolt:
Deprecation warning when installing EG
node-uuid@1.4.8: Use uuid module instead

PR #403, opened 2017-09-11
chore: replace node-uuid with uuid

XVincentX:
Connect #401  

The following pull request will replace `node-uuid` with `uuid` library - which is the suggested by the old one as well.

I also took the liberty to replace the `let lib = require("lib")` with `const lib = require("lib")` - so the Javascript engine will stop us to eventually replace the imported libraries.

Note: it seems like the library is still used internally by some other library; if you look at `package-lock` you can see `uuid-base62` - do you guys want me to look into this as well?


DrMegavolt:
Thank you for you PR

PR #404, opened 2017-09-11
Plugins CLI and API

### LINKED ISSUE #396: Plugin Framework: eg cli extension to do plugin installation and manipulate `system.config.yml` and `gateway.config.yml`

kevinswiber:
Fixes #396.

Ready for review.

- [X] please use "Connect #<Github Issue>" anywhere within the description
- [X] branch naming prefix
- [x] tests
- [x] build passes
- [ ] squash commits and/or rebase


lib/plugin-installer.js
```diff
@@ -0,0 +1,274 @@
...
+        ? systemConfig.plugins[name] : {};
+
+    return existingPluginOptions;
+  }
+
+  get pluginKey () {
+    let name = this.pluginManifest.name || this.packageName;
+    let abbreviatedPackageName = null;
+
+    if (this.packageName.startsWith(PluginInstaller.PACKAGE_PREFIX)) {
```
DrMegavolt: I was trying to install `ExpressGateway/express-gateway-plugin-example#master` that should work 
some flag needed to avoid prefixing

lib/plugin-installer.js
```diff
@@ -0,0 +1,274 @@
...
+  updateConfigurationFiles ({ packageName, pluginManifest,
+    pluginOptions, enablePlugin, addPoliciesToWhitelist }) {
+    // WARNING (kevinswiber): Updating YAML while maintaining presentation
+    // style is not easy.  We're using the YAWN library here, which has
+    // a decent approach given the current state of available YAML parsers,
+    // but it's far from perfect.  Take a look at existing YAWN issues
+    // before making any optimizations.  If any section of this code looks
+    // ugly or inefficient, it may be that way for a reason (or maybe not).
+    //
+    // ¯\_(ツ)_/¯
```
DrMegavolt: 👍

lib/plugin-installer.js
kevinswiber: I think this bit of code is doing its job just fine.  There's no code path to add a non-aliased plugin to the config.  That was because YAWN can't handle the empty object.  We can either add `package: <packagename>` or `disabled: false` to make it work... at least until we fix YAWN.

DrMegavolt: add `package: name` only when there is package
and don't add this property at all if name is the same as package, or I'm missing something ?

kevinswiber:
```yaml
example:

```

That won't parse in YAWN.

```yaml
example:
  disabled: false
```

```yaml
example:
  package: express-gateway-plugin-example
```

Those will parse in YAWN.

altsang: _sigh_ 
let's use `package:` for now to be explicit even when installing by convention and we can deal with the ramifications later

kevinswiber: Addressed in the latest commit.

bin/generators/plugins/configure.js
```diff
@@ -0,0 +1,142 @@
...
+              'or number.');
+
+            return false;
+          }
+
+          if (schema.required && !input) {
+            return false;
+          }
+
+          if (type === 'number' && isNaN(input)) {
```
DrMegavolt: integer is also valid JSON schema type as well as array (ip addresses etc.). 
I'm not even speaking for options like "minimum" and validation of array items 

So it may be ok for first version, but I would prefer using some validator like  https://www.npmjs.com/package/ajv or any other

DrMegavolt:
general question what should we do with options that are already  entered in system.config.yml ? 
it seems to be some edge case, (manual edit, or old plugin version obsolete properties)

however it may make sense to at least output as console.log entire config if it exists during `eg plugin configure`

bin/generators/plugins/install.js
```diff
@@ -0,0 +1,209 @@
...
+
+        self.pluginOptions[key] = value;
+      });
+    }
+
+    if (hasInvalidProperty) {
+      return;
+    }
+
+    return this.installer.runNPMInstallation({
```
DrMegavolt: ../express-gateway/bin/index.js plugin install ExpressGateway/express-gateway-plugin-example
does not save in the package.json for npm 4.6 most likely it will save for npm5.4 

can you add --save option here ?

kevinswiber: Done-zo.

bin/generators/plugins/configure.js
kevinswiber: Yes, we can increase JSON Schema support on a future iteration.

kevinswiber:
> general question what should we do with options that are already entered in system.config.yml ?
> it seems to be some edge case, (manual edit, or old plugin version obsolete properties)
> 
> however it may make sense to at least output as console.log entire config if it exists during eg plugin configure

On an update, the CLI will look at the options being offered by the new plugin version.  When it prompts for those values, if an existing value is in system.config.yml, it'll default to that value.

This model works fine when the options are the same or when new options are added.

One thing that isn't really supported is when options are removed.  We don't remove those on behalf of the user, and I'm not sure we should, either.

The other edge case is if an option remains in the new version but the type has changed (say from string to number).  The existing value wouldn't validate.  But the user would get prompted for it anyway, and the validation provided by the CLI should match that of the new version.

bin/generators/plugins/configure.js
```diff
@@ -0,0 +1,142 @@
...
+  constructor (args, opts) {
+    super(args, opts);
+
+    this.installer = null;
+    this.pluginOptions = null;
+
+    this.configureCommand({
+      command: 'configure <package>',
+      description: 'Configure a plugin',
+      builder: yargs =>
```
XVincentX: Nit: `(yargs) =>`

bin/generators/plugins/configure.js
```diff
@@ -0,0 +1,142 @@
...
+            return false;
+          }
+
+          return true;
+        }
+      };
+    });
+
+    return this.prompt(pluginQuestions)
+    .then(answers => {
```
XVincentX: Nit: this should have another level of indentation

bin/generators/plugins/configure.js
```diff
@@ -0,0 +1,142 @@
...
+      this.pluginOptions = {};
+
+      const keys = pluginQuestions.map(opt => opt.name);
+      const self = this;
+      keys.forEach(key => {
```
XVincentX: Nip: new line

bin/generators/plugins/install.js
```diff
@@ -0,0 +1,209 @@
...
+    if (argv.o) {
+      optionValues = Array.isArray(argv.o) ? argv.o : [argv.o];
+    }
+
+    let hasInvalidProperty = false;
+
+    if (optionValues.length) {
+      this.pluginOptions = {};
+
+      const self = this;
```
XVincentX: Is this still required? Arrow function should be automatically bound to the correct `this`, therefore you should be able to use that directly

lib/plugin-installer.js
```diff
@@ -0,0 +1,272 @@
...
+    //
+    // https://github.com/mohsen1/yawn-yaml/issues
+
+    return new Promise((resolve, reject) => {
```
XVincentX: I do not see any asynchronous flow here. If I am correct (the method is huge, I might me missing it) why would you return a `Promise`?

lib/rest/routes/plugins.js
```diff
@@ -0,0 +1,202 @@
...
+
+    const pluginManifest = getPluginManifest(packageName);
+    const version = getPackageVersion(packageName);
+
+    pluginManifest.options = pluginManifest.options || {};
+
+    const optionKeys = Object.keys(plugin);
+    const reserved = ['state', 'disabled', 'package'];
+    const options = optionKeys
+      .filter(optionKey => {
```
XVincentX: Nit: if you're returning immediately (which is your case) you can avoid the `{}` and the `return` keyword

bin/generators/plugins/install.js
kevinswiber: Nope.  Just an old habit.

bin/generators/plugins/configure.js
kevinswiber: :+1:

bin/generators/plugins/configure.js
kevinswiber: Agree to disagree.  😃

lib/rest/swagger-doc.yaml
```diff
@@ -1,740 +1,931 @@
...
+                description: Enable debugging
+              value: true
+    PluginInstallResponse:
+      allOf:
+        - '$ref': '#/components/schemas/Plugin'
+        - example:
+            name: custom-plugin
+            description: An example plugin
+            package: express-gateway-plugin-custom
+            version: 1.1.0
```
DrMegavolt: 1.2.0 ?

lib/rest/swagger-doc.yaml
```diff
@@ -1,740 +1,931 @@
...
+              type: boolean
+              default: true
+    PluginResponse:
+      allOf:
+        - '$ref': '#/components/schemas/Plugin'
+        - example:
+            name: custom-plugin
+            description: An example plugin
+            package: express-gateway-plugin-custom
+            version: 1.1.0
```
DrMegavolt: same here 1.2.0 ?

test/e2e/cli-plugin-install.e2e.test.js
```diff
@@ -0,0 +1,84 @@
...
+  systemConfigPath: null,
+  gatewayConfigPath: null
+};
+
+describe('E2E: eg plugins install', () => {
+  before(() => {
+    return util.promisify(tmp.dir)()
+      .then(temp => {
+        tempPath = temp;
+        return util.promisify(cpr)(gatewayDirectory, tempPath);
```
XVincentX: I'd save and store the promisified instances on top of the file.

test/e2e/cli-plugin-install.e2e.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+let tempPath = null;
+let configPath = null;
+
+const config = {
+  systemConfigPath: null,
+  gatewayConfigPath: null
+};
+
+describe('E2E: eg plugins install', () => {
```
XVincentX: Testing frameworks are usually relying on `bind` a lot to inject a `this` object you can use down in the chain. If you're using arrow functions, that won't work. I usually stick with `function() {}` here.

test/e2e/cli-plugin-install.e2e.test.js
```diff
@@ -0,0 +1,84 @@
...
+        config.systemConfigPath = path.join(configPath, 'system.config.yml');
+        config.gatewayConfigPath = path.join(configPath, 'gateway.config.yml');
+      })
+      .then(() => {
+        return runCLICommand({
+          cliArgs: ['plugins', 'install', pluginDirectory, '-n', '-g',
+            '-o', '"foo=bar"',
+            '-o', '"baz=4444"'],
+          adminPort: 0,
+          configDirectoryPath: configPath,
```
XVincentX: You can use `return Promise.all` to pass internal data between promises. It will avoid you having tons of global objects on top of the file.

bin/generators/plugins/configure.js
XVincentX: As long we stick with the same convention, I'm fine with that. I saw some inconsistencies around, though.

lib/plugin-installer.js
```diff
@@ -0,0 +1,272 @@
...
+        '--save'
+      ];
+
+      const installOpts = {
+        cwd: cwd || process.cwd(),
+        env: env || process.env,
+        stdio: ['ignore', 'pipe', 'inherit']
+      };
+
+      const npmInstall = spawn('npm', installArgs, installOpts);
```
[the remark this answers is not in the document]
XVincentX: Note to self: propose [appveyor](https://www.appveyor.com/) to check if Windows is ok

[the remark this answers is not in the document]
kevinswiber: Should work fine in Windows.  We have Appveyor.  I just checked it out, though, and it hasn't run in two months.  😮

test/e2e/cli-plugin-install.e2e.test.js
kevinswiber: I know, but I'm not using `this` in this particular test.  `function` is just sooo many more characters.  :wink:

lib/rest/swagger-doc.yaml
kevinswiber: This is the plugin version, not the configuration format version.

lib/rest/swagger-doc.yaml
kevinswiber: This is the plugin version, not the configuration format version.

lib/plugin-installer.js
```diff
@@ -0,0 +1,272 @@
...
+  static get PACKAGE_PREFIX () {
+    return 'express-gateway-plugin-';
+  }
+
+  static create (options) {
+    return new PluginInstaller(options);
+  }
+
+  runNPMInstallation ({ packageSpecifier, cwd, env }) {
+    return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
kevinswiber: Sure, take a look.  Let me know if you find anything.  I don't see why this would be frowned upon.  The only events we'd need to communicate are based on success or failure, which surface through the Promise API quite nicely.

lib/plugin-installer.js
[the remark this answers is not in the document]
kevinswiber: @XVincentX I think `npm` is a pretty hefty dependency to take on.  I'm fine with shelling out to it.  If it gets to a point where `npm` is no longer shipped with Node as the default package manager, then I think we should consider this.

test/e2e/cli-plugin-install.e2e.test.js
XVincentX: Keyboard strokes are expansive these days 😱

Issue #409, opened 2017-09-14
Plugin Framework: replace AdminExtensions and GatewayExtensions in favor of Routes


altsang:
Sometimes custom routes will need to be added as helper functions/endpoints

- Routes can be defined for both the Admin or Gateway instance within Express Gateway
- in doing so, plugin folder structure will have `routes`
- Routes can be registered through the PluginContext using the `.registerAdminRoutes` and `registerGatewayRoutes` methods
-  Adding a route is intended to be used for helper endpoints like in the case of OAuth2 for accessing and validating tokens through `/token`
- Routes can contain calls out to Express middleware to support processing like `body-parser`

Notes: all routes defined are always _GLOBAL_ to the Express instance per how Express works and always executed use with caution

PR #410, opened 2017-09-14
renamed extension to routes

DrMegavolt:
"Connect #409"

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -43,8 +43,8 @@ module.exports.load = function ({config}) {
   return {
     policies: extract(loadedPlugins, 'policies'),
     conditions: extract(loadedPlugins, 'conditions'),
-    gatewayExtensions: extract(loadedPlugins, 'gatewayExtensions'),
-    adminExtensions: extract(loadedPlugins, 'adminExtensions'),
+    gatewayRoutes: extract(loadedPlugins, 'gatewayRoutes'),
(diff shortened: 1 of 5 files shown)


lib/plugins.js
```diff
@@ -67,11 +67,11 @@ class PluginContext {
   registerCondition (condition) {
     this.conditions.push(condition);
   }
-  registerGatewayExtension (gatewayExtension) {
-    this.gatewayExtensions.push(gatewayExtension);
+  registerGatewayRoute (gatewayRoute) {
```
kevinswiber: Should this be plural?  We can register more than one route at a time through this, right?  @altsang @DrMegavolt

DrMegavolt: @kevinswiber this should be singular because it accepts only one non array parameter
but the naming of argument could be 
gatewayRoutesDeclaration

lib/plugins.js
```diff
@@ -67,11 +67,11 @@ class PluginContext {
   registerCondition (condition) {
     this.conditions.push(condition);
   }
-  registerGatewayRoute (gatewayRoute) {
-    this.gatewayRoutes.push(gatewayRoute);
+  registerGatewayRoute (gatewayRoutesDeclaraion) {
+    this.gatewayRoutes.push(gatewayRoutesDeclaraion);
```
kevinswiber: `gatewayRoutesDelaraion` -> `gatewayRoutesDeclaration` . Missing the `t`.

PR #412, opened 2017-09-14
Fixing plugin installation on Windows: spawning npm.cmd instead of npm.

kevinswiber:


kevinswiber:
Waiting for one more `yawn-yaml` fix (https://github.com/mohsen1/yawn-yaml/pull/12).  Then we'll update the dependency and should be good to go!

kevinswiber:
This is good to go.  AppVeyor runs two builds, one for Node 6 and one for Node 8.  The Node 8 one failed because of our fragile hot reload test.

Issue #385, opened 2017-08-29
Admin: extend API to include policy declaration


altsang:

DrMegavolt:
As I understood it is to add policy to whitelist 
and #387 is for actual config

PR #418, opened 2017-09-18
Feature/admin pipelines, endpoints, whitelist

DrMegavolt:
- connect #416 connect #417 (duplicates connect #388) connect #385 connect #389

pipelines, apiEndpoints, serviceEndpoints
/pipelines, /api-enpoints, /service-endpoints  
GET /  -> list of all entities
GET /:name -> entity by object key
PUT /:name with body -> create\update entity with this name as key
DELETE /:name -> remove entity 

whitelist
GET /policies/whitelist -> list of all allowed policies
PUT /policies/whitelist/:name - add to the list
DELETE /policies/whitelist/:name - rm from list


lib/config/config.js
```diff
@@ -103,6 +104,30 @@ class Config {
       gatewayWatcher.isReady = true;
     });
   }
+
+  // this function suppose to preserve comments.
+  // due to fragile nature of YAWN it may not work properly.
+  updateGently (modifier) {
+    return new Promise((resolve, reject) => {
+      fs.readFile(this.gatewayConfigPath, 'utf8', (err, data) => {
```
stanislaw-glogowski: Can we use util.promisify for all callback like methods?

lib/config/config.js
```diff
@@ -103,6 +104,30 @@ class Config {
...
+        if (err) {
+          return reject(err);
+        }
+        let json = yamlOrJson.load(data);
```
stanislaw-glogowski: There should be `const` instead of `let`. (`let` is a signal that the variable may be reassigned)

lib/config/config.js
```diff
@@ -1,4 +1,5 @@
 const fs = require('fs');
+// const YAWN = require('yawn-yaml/cjs');
```
XVincentX: Leftover comment?

lib/rest/routes/api-endpoints.js
```diff
@@ -0,0 +1,34 @@
...
+const express = require('express');
+
+module.exports = function ({config}) {
+  let router = express.Router();
+  router.put('/:name', function (req, res, next) {
+    config.updateGently(json => {
+      json.apiEndpoints[req.params.name] = req.body;
+      return json;
+    })
+    .then(() => {
```
XVincentX: Nitpick:
`.then(() => res.json({status: 'ok'}));`

lib/rest/routes/api-endpoints.js
```diff
@@ -0,0 +1,34 @@
...
+  });
+  router.delete('/:name', function (req, res, next) {
+    config.updateGently(json => {
+      delete json.apiEndpoints[req.params.name];
+      return json;
+    })
+    .then(() => {
+      res.json({status: 'ok'});
+    })
+    .catch(next);
```
XVincentX: 👍

lib/rest/routes/pipelines.js
```diff
@@ -0,0 +1,34 @@
...
+    config.updateGently(json => {
+      delete json.pipelines[req.params.name];
+      return json;
+    })
+    .then((json) => {
+      res.json({status: 'ok'});
+    })
+    .catch(next);
+  });
+  return router;
```
XVincentX: This seems to be almost the same code as `api-endpoints.js` (and other files as well), the only difference is the target json object. It can be probably generalized and reused in all the places you need.

lib/config/config.js
```diff
@@ -103,6 +104,30 @@ class Config {
...
+        // let yawn = new YAWN(data);
+        // yawn.json = modifier(yawn.json);
+        // let text = yaml.yaml;
+        fs.writeFile(this.gatewayConfigPath, text, (err) => {
+          if (err) {
+            return reject(err);
+          }
+          resolve();
+        });
+      });
```
XVincentX: I would refactor this piece of code to take advantage of the `util.promisify` and chain the promises instead of going into a callback hell.

Something like this:

```javascript
const fs = require("fs");
const readFile = util.promisify(fs.readFile);
const writeFile = util.promisify(fs.writeFile);

Pseudo code (it might be a bit different)
function updateGently(modifier) {
  return readFile(this.gatewayConfigPath, 'utf8').then((data) => {

    let json = yamlOrJson.load(data);
    let result = modifier(json);
    let text = yamlOrJson.dump(result);
    return writeFile(this.gatewayConfigPath, text);
  });
}
```

This feels more readable to me but more importantly - easier to reason about.

lib/config/config.js
XVincentX: `eslint` is usually able to catch these things - is it configured in CircleCI?

XVincentX:
I can see double commits with the same message 🤔 
image

My only "concern" is the code duplication - it can probably be extracted and reused across the routes when the action we want to propose are almost the same.

I can live with that, anyway ;)

lib/config/config.js
stanislaw-glogowski: Can we consider switching to babel? async / await syntax really helps in working with promises.

XVincentX: Personally speaking I would refrain to introduce babel and a transpilation step for node programs. `Promises` are good enough and can handle all the use cases without sacrificing the readability, it's just matter of reorganizing the flows a bit. We can discuss that separately.

kevinswiber:
> I can see double commits with the same message

It's of little consequence.  We "Squash & Merge" PRs by default.

lib/rest/routes/pipelines.js
DrMegavolt: no generalization, intentionally copy pasted. 
The reason is further differences: 
validation is one example
additional methods to manipulate policies 

Having some generic monster would greatly affect maintainability

lib/config/config.js
```diff
@@ -103,6 +107,25 @@ class Config {
...
     });
   }
+
+  // this function can preserve comments and formatting.
+  // due to fragile nature of YAWN it may not work properly in that mode.
+  // safe way is to use it in rewrite file mode
+  update (modifier) {
+    return readFile(this.gatewayConfigPath, 'utf8').then(data => {
+      let text;
+      if (this.systemConfig && this.systemConfig.preserveStructureOnUpdates) {
```
DrMegavolt: @kevinswiber seems to be a compromise on yawn with comments vs nuke it all approach? 
by default it will be nuke it all, and if people complain it will be a doc section describing this flag
(obviously name can be better)

lib/rest/routes/api-endpoints.js
```diff
@@ -0,0 +1,30 @@
+const express = require('express');
+
+module.exports = function ({config}) {
+  let router = express.Router();
+  router.put('/:name', function (req, res, next) {
+    let isUpdate;
+    config.update(json => {
+      isUpdate = req.params.name in json.apiEndpoints;
+      json.apiEndpoints[req.params.name] = req.body;
```
XVincentX: I think you guys already discussed this - so bear with me if I am repetitive - but this might be harmful as fundamentally I can put everything I want without having any sort of validation.

lib/rest/routes/pipelines.js
```diff
@@ -0,0 +1,30 @@
...
+      json.pipelines[req.params.name] = req.body;
+      return json;
+    })
+    .then(() => res.status(isUpdate ? 204 : 201).send())
+    .catch(next);
+  });
+  router.get('/:name', function (req, res, next) {
+    res.json(config.gatewayConfig.pipelines[req.params.name]);
+  });
+  router.get('/', function (req, res, next) {
```
XVincentX: Is the `next` parameter needed for all these routes? This can probably be safely removed.

lib/config/config.js
```diff
@@ -103,6 +107,25 @@ class Config {
...
+        let yawn = new YAWN(data);
+        yawn.json = modifier(yawn.json);
+        text = yawn.yaml;
+      } else {
+        let json = yamlOrJson.load(data);
+        let result = modifier(json);
+        text = yamlOrJson.dump(result);
+      }
+      return writeFile(this.gatewayConfigPath, text);
```
XVincentX: I like this now so much 👍

lib/rest/routes/api-endpoints.js
DrMegavolt: I agree as for general idea, why it is not in this PR 

I cannot do it for pipeline policies, because it depends on introducing schema for policies in plugins. 

Here the only thing that is required is 'host' field. but for `paths` it is quite complex and I would like to move it after approach for pipelines validation is settled. (+ some intentionally undocumented fields do exist) 

and by the way UI is going to add some new fields like fieldId to each entity that has nothing to do with EG, but needed for UI. so its cannot be strict schema

lib/rest/routes/pipelines.js
DrMegavolt: and what is the problem with having next ? 
(req,res), or (req,res,next) are well known signatures for middleware 

and it seems that there is no rule in standard js that says it is not good
https://standardjs.com/rules.html

I remember having that rule enabled in the project but I didn't like  the experience, because in some cases implementation change and you need add\remove arguments constantly. 

but I will agree if this will promoted as ESlint rule

lib/rest/routes/api-endpoints.js
XVincentX: It makes perfectly sense - thanks a lot for the clarification 👍

lib/rest/routes/pipelines.js
XVincentX: The general rule I found among the project (although it's not codified) is

route: `fn(req, res)`
middleware: `fn(req, res, next)`

but it's a nitpick, it's absolutely not blocking anything.

lib/admin/config/api-endpoints.js
```diff
@@ -0,0 +1,27 @@
+module.exports = function (client) {
+  let baseUrl = `api-endpoints/`;
```
XVincentX: This can be `const`

Issue #419, opened 2017-09-18
JWT Plugin: verification


altsang:
_From @altsang on September 10, 2017 21:34_

see #30 for part 1 of the JWT use case

altsang:
EPIC: 
- for planning poker purposes - please assign a point size
- this is needs to be broken down into engineering user stories - please prefix with `JWT Plugin:`

### Background Info
- per our plugin naming convention, this will be the first iteration of jwt plugin -  `express-gateway-plugin-jwt`
- JWT RFC - https://tools.ietf.org/html/rfc7519
- this is an auth mechanism to EG, to keep consistent with the ecosystem and the interfaces already adopted, we should use `Passport` based middleware (i.e. `passport-jwt`
- library for generating JWT - `jsonwebtoken`
- Passport JWT overview - https://blog.jscrambler.com/implementing-jwt-using-passport/
- super helpful debugger - https://jwt.io

### New JWT Policy and Plugin Registration
The plugin will introduce a new auth policy - JWT.
- plugin is registered under `system.config.yml` in the `plugins:` section 
- system wide parameters are specified underneath its entry
  * querystring param that contain JWT - `jwt` = default
  * claim used containing the JWT key, `iss` = default
  * registered claims that can be verified
    - `iss`-  in the case of EG being the provider, the parameter specified above is used to identify the key
    - `sub` - TBD, should understand the use case for this claim further
    - `aud` - TBD, however, if `sub` is used, then `aud` must also be verified
    - `exp` - EG can verify this claim
    - `nbf` - EG can verify this claim
    - `jti` -  an extra layer of security to prevent replay, probably not necessary for 1st iteration of JWT
- a JWT policy can be declared in the `gateway.config.yml` in the `policies:` section which will allow the JWT policy to be used within a pipeline

### New Credential Type
The plugin will extend EG credential management by introducing a new credential type known as JWT - (i.e. `jwt`, just like `basic`, `oauth2` etc)

### JWT Credential Type 
The new JWT credential type will need to contain the following:
- keyId - auto generated (please refer to how `key-auth` works in EG for more info on how keys are generated in pairs - id and secret)
- verification algorithm: `HS256`, `RS256`, or `ES256`
- rsa_public key: for `RS256`/`ES256`
- keySecret:  auto generated (note: if the algo is `H256`/`ES256`)

### Admin API and CLI extensions
- you need to be create a `jwt` credential type via CLI
- parameters:
  * which `algo` to use (see above)
[...]

Issue #30, opened 2017-04-24
JWT Plugin:  3rd part introspection capabilities


jimhw:
A developer familiar with Express middleware should be able to take any middleware that extracts the parameters that are necessary to construct a configured instance of it to be passed dynamically when instantiating that middleware. 

Those parameters should go into EG config files in a standardized format.

altsang:
we'll use JWT as the documented example of taking an off the shelf middleware for `passport-jwt`and turn into an EG policy

altsang:
the JWT use case can be broken into two parts:
- JWT introspection and verification (an authoritative source other than EG)
- EG credential management introducing a new token type - JWT,  validating, generating and maintaining these tokens as part of the existing token service

this user story covers #1

XVincentX:
Introspection capabilities have been implemented, I'm closing this.

PR #420, opened 2017-09-19
Remove bluebird dependency

### LINKED ISSUE #217: Core: rm bluebird dependency

stanislaw-glogowski:
connect #217


stanislaw-glogowski:
Same tests fail on master branch

lib/services/tokens/token.dao.js
```diff
@@ -72,27 +69,31 @@ dao.find = function (tokenObj, options) {
...
-            if (!token) {
-              return Promise.reject(new Error());
-            }
+      const getTokenPromise = new Promise((resolve, reject) => {
+        activeTokenIds.reverse();
 
-            isEqual = Object.keys(tokenObj).every((key) => tokenObj[key] === token[key]);
-            return isEqual ? token : Promise.reject(new Error());
-          });
+        const next = () => {
```
DrMegavolt: I believe you have correct assumption that typically there are not many active tokens, most likely 0-1. So this approach should work

lib/services/tokens/token.dao.js
```diff
@@ -57,7 +54,7 @@ dao.find = function (tokenObj, options) {
 
   return db.hgetallAsync(consumerTokensKey)
     .then((tokenIds) => {
-      let tokenPromises, activeTokenIds, getTokenPromise;
+      let activeTokenIds;
```
XVincentX: We need a style guide, or we will keep changing these lines of code according to the tastes of the person that's working on it.

lib/services/tokens/token.dao.js
```diff
@@ -72,27 +69,31 @@ dao.find = function (tokenObj, options) {
...
-          });
+        const next = () => {
+          if (!activeTokenIds.length) {
+            reject(new Error());
+            return;
+          }
+          db.hgetallAsync(tokenNamespace.concat(':', activeTokenIds.pop()))
+            .then((token) => {
+              if (!token || !Object.keys(tokenObj).every((key) => tokenObj[key] === token[key])) {
+                throw new Error();
```
XVincentX: I'd put an error message here.

lib/services/tokens/token.dao.js
```diff
@@ -155,11 +156,11 @@ dao.revoke = function (token, options) {
...
-        .multi()
-        .hset(tokenNamespace.concat(':', token.id), 'archived', 'true')
-        .hdel(consumerTokensNamespace.concat(':', token.consumerId), token.id)
-        .hset(consumerTokensExpiredNamespace.concat(':', token.consumerId), token.id, 'true')
-        .execAsync();
+    .multi()
+    .hset(tokenNamespace.concat(':', token.id), 'archived', 'true')
+    .hdel(consumerTokensNamespace.concat(':', token.consumerId), token.id)
+    .hset(consumerTokensExpiredNamespace.concat(':', token.consumerId), token.id, 'true')
+    .execAsync();
```
XVincentX: dtto for the style guide comment

lib/services/utils.js
```diff
@@ -1,51 +1,57 @@
...
-  return bcrypt.genSalt(config.systemConfig.crypto.saltRounds)
-  .then(function (salt) {
-    return bcrypt.hashAsync(password, salt);
-  })
-  .catch(() => Promise.reject(new Error('password hash failed'))); // TODO: replace with server error
+
+  return bcrypt
+    .genSalt(config.systemConfig.crypto.saltRounds)
+    .then((salt) => hashAsync(password, salt))
+    .catch(() => Promise.reject(new Error('password hash failed'))); // TODO: replace with server error
```
XVincentX: This feels really weird (not your thing - I just was having a look)

lib/services/tokens/token.dao.js
stanislaw-glogowski: I didn't want to change logic / style of existing code. I only removed `bluebird` promise library.

lib/services/tokens/token.dao.js
stanislaw-glogowski: new Error() was here before. Take a look what happened with that error next (`catch(() => null)`). In my opinion whole file needs to be rewritten.

DrMegavolt:
@stanislaw-glogowski @XVincentX while there is a general code style\architecture problem in this module please make as less changes as possible. 

This is far more complicated than  Error() vs Error(message) or  const vs let. Once you get good understanding of what this service is doing in combination with other consumer management parts we can start discussing changes 

And the way to understand is JWT implementation #419 #30

PR #426, opened 2017-09-22
Clears default gateway template

### LINKED ISSUE #425: Bug: generator for basic express gateway is not correct
the generator for the basic EG server should not have any service or API endpoints defined

it should contain one pipeline called `default`
in the default pipeline should be the proxy policy, if the proxy policy requires a service endpoint or requires input parameters - then comment it out

stanislaw-glogowski:
fixes #425

### DIFF
--- a/bin/generators/gateway/templates/basic/config/gateway.config.yml
+++ b/bin/generators/gateway/templates/basic/config/gateway.config.yml
@@ -4,11 +4,9 @@ admin:
   port: 9876
   hostname: localhost
 apiEndpoints:
-  api:
-    host: '*'
+  # see: http://www.express-gateway.io/docs/configuration/gateway.config.yml/apiEndpoints
 serviceEndpoints:
-  httpbin:
-    url: 'https://httpbin.org'
+  # see: http://www.express-gateway.io/docs/configuration/gateway.config.yml/serviceEndpoints
 policies:
   - basic-auth
   - cors
@@ -23,7 +21,5 @@ pipelines:
     apiEndpoints:
       - api
     policies:
-      - proxy:
-          - action:
-              serviceEndpoint: httpbin
-              changeOrigin: true
+      # - proxy: # see: http://www.express-gateway.io/docs/policies/proxy
+


altsang:
will all of these deletions - the gateway still comes up if `npm start`?

kevinswiber:
I'd prefer to see this as a new template named `bare` or something.  For the basic or default gateway, I'd like the experience to include `cd gateway && npm start` to get a little instant gratification.  Thoughts, @altsang?

kevinswiber:
Also, we need to test this configuration with the latest PRs related to updating config via the Admin API.  IIRC, our YAML-manipulator (YAWN) crashes on empty objects.

stanislaw-glogowski:
@altsang it doesn't start. Instead of it throws `Configuration Error('No api Endpoints found')`.

lib/gateway/pipelines.js
```diff
@@ -140,40 +140,41 @@ function configurePipeline (pipelinePoliciesConfig, config) {
   validatePipelinePolicies(pipelinePoliciesConfig, config);
 
   pipelinePoliciesConfig.forEach(policyConfig => {
-    let policyName = Object.keys(policyConfig)[0];
-    let policySteps = policyConfig[policyName];
+    if (policyConfig && typeof policyConfig === 'object') {
```
DrMegavolt: I suggest to negate if and return instead of putting entire code inside block

lib/gateway/pipelines.js
```diff
@@ -190,12 +191,16 @@ function validateGatewayConfig (config) {
   if (!gatewayConfig.pipelines) {
     if (gatewayConfig.pipeline) {
       gatewayConfig.pipelines = Array.isArray(gatewayConfig.pipeline) ? gatewayConfig.pipeline : [ gatewayConfig.pipeline ];
-    } else throw new ConfigurationError('No pipelines found');
+    } else {
+      gatewayConfig.pipeline = [];
```
DrMegavolt: what if we merge line 191 to this in something like 
gatewayConfig.pipelines = gatewayConfig.pipelines || gatewayConfig.pipeline || [] ? 

it will reduce nesting, and later in the code  always use plural form

lib/gateway/pipelines.js
```diff
@@ -190,12 +191,16 @@ function validateGatewayConfig (config) {
...
+    }
   }
   if (!gatewayConfig.apiEndpoints) {
     if (gatewayConfig.apiEndpoint) {
       gatewayConfig.apiEndpoints = Array.isArray(gatewayConfig.apiEndpoint) ? gatewayConfig.apiEndpoint : [ gatewayConfig.apiEndpoint ];
-    } else throw new ConfigurationError('No apiEndpoints found');
+    } else {
+      gatewayConfig.apiEndpoints = [];
```
DrMegavolt: same suggestion as for pipelines
gatewayConfig.apiEndpoints = gatewayConfig.apiEndpoints || gatewayConfig.apiEndpoint || {} ?

lib/gateway/pipelines.js
DrMegavolt: and pipelines is not an array, it is an object 
name: pipelineDefinition

lib/gateway/pipelines.js
```diff
@@ -223,14 +228,16 @@ function validatePipelinePolicies (policies, config) {
   let policiesConfig = config.gatewayConfig.policies || [];
 
   policies.forEach(policyObj => {
-    let policyNames = Object.keys(policyObj);
+    if (policyObj && typeof policyObj === 'object') {
```
DrMegavolt: how this can be ? 
if there is no policies just do not call  policies.forEach, why this check in foreach ?

DrMegavolt: if the idea is to protect from some invalid policy configuration please add some validation logic

lib/gateway/pipelines.js
DrMegavolt: and if there is no policies in pipeline do not enter forEach at all, not sure if the line is needed at all. 
if configuration is invalid it should at least fail (current behavior), preferably with useful description

altsang:
@kevinswiber bare is fine - but what would basic have that would be instant gratification? what we have now? , if so - a little weird that the only difference between it and getting started is the fact there's a key-auth entry....

PR #427, opened 2017-09-22
Fixes logger path

### LINKED ISSUE #421: Bug: logger.warn() is not a function
Getting the following error with 1.2.0

```javascript
TypeError: logger.warn is not a function
    at authService.authenticateCredential.then.catch.err (/home/val/Workspace/meanIT/LunchBadger/scopes/node_modules/express-gateway/lib/policies/key-auth/keyauth.js:37:14)
    at <anonymous>
```

Offending code is [here](https://github.com/ExpressGateway/express-gateway/blob/733d2dcd84b0af50e0814ba8b796a0aaa16bb5e9/lib/policies/key-auth/keyauth.js#L37) attempting to call into [this file](https://github.com/ExpressGateway/express-gateway/blob/733d2dcd84b0af50e0814ba8b796a0aaa16bb5e9/lib/policies/log/log.js) which doesn't have a `.warn()` function.

Here's my config file. I get this error when sending a request to an endpoint which my credential doesn't have the proper scope:

```
http:
  port: 8080
admin:
  port: 9876
  hostname: localhost
apiEndpoints:
  ip:
    host: localhost
    paths: '/ip'
  get:
    host: localhost
    paths: '/get'
    scopes:
      - 'admin'
serviceEndpoints:
  httpbin:
    url: 'https://httpbin.org'
policies:
  - basic-auth
  - cors
  - expression
  - key-auth
  - log
  - oauth2
  - proxy
  - rate-limit
pipelines:
  default:
    apiEndpoints:
      - ip
      - get
    policies:
      - key-auth:
      - proxy:
          - action:
              serviceEndpoint: httpbin
              changeOrigin: true
```

I have a patch to remove the offending code here: https://github.com/ExpressGateway/express-gateway/compare/master...vkarpov15-patch-1 but the checklist for contributing a PR is far too long and I didn't want to break it.

stanislaw-glogowski:
fixes #421

### DIFF
--- a/lib/policies/key-auth/keyauth.js
+++ b/lib/policies/key-auth/keyauth.js
@@ -1,7 +1,7 @@
 const passport = require('passport');
 const LocalAPIKeyStrategy = require('./passport-apikey-strategy');
 let services = require('../../services/index');
-let logger = require('../log').policy;
+let logger = require('../../logger').policy;
 let authService = services.auth;
 const credentialType = 'key-auth';
 passport.use(new LocalAPIKeyStrategy({passReqToCallback: true}, (req, apikey, done) => {


XVincentX:
My two cents: such an error almost means that this part is totally untested. We should fix that in a way or another.

Issue #383, opened 2017-08-25
Delete Credential  CLI command (and REST api ) is missing 


DrMegavolt:
@ahwing210 found that no delete credential is implemented

altsang:
@awhing210 - you can deactivate credentials, which will preserve the complete history for audit purposes and there is an archive that happens as well - if we filtered deactivated credentials from the `list` command  - would this work for you?  Trying to understand if you really need a hard delete and why? thx!

altsang:
make `list` command by default not to show deactivated credentials, add option to show deactivated, active, or both

XVincentX:
This has been done in #428 and will land in EG 1.3.

Issue #260, opened 2017-07-11
CLI: add ability to do operations against Apps with friendly names instead of AppID GUID


altsang:

stanislaw-glogowski:
### CLI options proposal:
https://docs.google.com/document/d/17i_Blpwes_iPfiTlc96phCoEomRZWiUGOogmEM82Hp0/edit?usp=sharing

XVincentX:
This is being paused because of #499

XVincentX:
In order to provide the same feature in the same way search by user has been provided, we would need to add a new key in Redis that has been missing so far for all the applications. 

Although we could add it and provide the features, all legacy users won't be able to search by application name for the legacy.

1. Implement a different (and probably inefficient system — get all apps and manually filtering them)
2. Provide another migration script to take all the applications and index them properly.

@altsang @DrMegavolt

DrMegavolt:
yes the new key is what is expected. I believe original PR already has it. but in general it works same way as for users. 
1) no, just another relation field should work 
2) yes, a migration to build such index

PR #428, opened 2017-09-22
Adds list credentials modes

stanislaw-glogowski:
Connected #383

CLI:
```bash
$ eg credentials list [options]
# Options:
# -c --consumerId <consumerId> - Consumer ID: can be User ID or username or app ID
# -f --filter <filter> - Comma separated list of credential state (active, archived), default: active
```

List active credentials for consumer (default):
```bash
$ eg credentials list -c <consumerId>
```
List archived credentials for consumer:
```bash
$ eg credentials list --filter active -c <consumerId>
```
List active and archived credentials for consumer:
```bash
$ eg credentials list -f active,archived -c <consumerId>
# or
$ eg credentials list -f active -f archived -c <consumerId>
```

### DIFF
--- a/bin/generators/credentials/list.js
+++ b/bin/generators/credentials/list.js
@@ -6,32 +6,24 @@ module.exports = class extends eg.Generator {
 
     this.configureCommand({
       command: 'list [options]',
-      description: 'List all active credentials for Consumer (User or App)',
+      description: 'List Consumer (User or App) credentials',
       builder: yargs =>
         yargs
           .usage(`Usage: $0 ${process.argv[2]} list [options]`)
           .example(`$0 ${process.argv[2]} list -c 7498d1a9-7f90-4438-a9b7-0ba4c6022353`)
-          .string(['c'])
-          .boolean(['a', 'd'])
+          .string(['c', 'i'])
           .describe('c', 'Consumer ID: can be User ID or username or app ID')
-          .describe('a', 'List activated and deactivated credentials')
-          .describe('d', 'List only deactivated credentials')
+          .describe('i', 'Comma separated list of credential state (active, archive), default: active')
           .alias('c', 'consumerId').nargs('c', 1)
-          .alias('a', 'all')
-          .alias('d', 'deactivated')
-          .group(['c', 'a'], 'Options:')
+          .alias('i', 'include')
+          .group(['c', 'i'], 'Options:')
     });
   }
 
   prompting () {
-    const {consumerId, all, deactivated} = this.argv;
-    let mode = 'activeOnly';
-    if (all) {
-      mode = 'both';
-    } else if (deactivated) {
-      mode = 'deactivatedOnly';
-    }
-    return this.admin.credentials.list(consumerId, mode)
+    const {consumerId, include} = this.argv;
+
+    return this.admin.credentials.list(consumerId, include)
       .then(data => {
         const credentials = data.credentials;
         if (!credentials || !credentials.length) {
--- a/bin/generators/credentials/list.js
+++ b/bin/generators/credentials/list.js
@@ -26,7 +26,8 @@ module.exports = class extends eg.Generator {
 
     return this.admin.credentials.list(consumerId, filter)
       .then(data => {
-        const credentials = { data };
+        const { credentials } = data;
+
         if (!credentials || !credentials.length) {
           this.log.error(`Consumer ${consumerId} has no credentials`);
           return;
--- a/bin/generators/credentials/list.js
+++ b/bin/generators/credentials/list.js
@@ -16,7 +16,8 @@ module.exports = class extends eg.Generator {
           .describe('f', 'Comma separated list of credential state (active, archived), default: active')
           .alias('c', 'consumerId').nargs('c', 1).required('c')
           .alias('f', 'filter')
-          .group(['c', 'i'], 'Options:')
+          .coerce('f', (val) => val.toLowerCase())
+          .group(['c'], 'Options:')
     });
   }
 
--- a/bin/generators/credentials/list.js
+++ b/bin/generators/credentials/list.js
@@ -6,24 +6,32 @@ module.exports = class extends eg.Generator {
 
     this.configureCommand({
       command: 'list [options]',
-      description: 'List all credentials for Consumer (User or App)',
+      description: 'List all active credentials for Consumer (User or App)',
       builder: yargs =>
         yargs
           .usage(`Usage: $0 ${process.argv[2]} list [options]`)
           .example(`$0 ${process.argv[2]} list -c 7498d1a9-7f90-4438-a9b7-0ba4c6022353`)
           .string(['c'])
-          .boolean(['a'])
+          .boolean(['a', 'd'])
           .describe('c', 'Consumer ID: can be User ID or username or app ID')
-          .describe('a', 'List all activated and deactivated credentials')
+          .describe('a', 'List activated and deactivated credentials')
+          .describe('d', 'List only deactivated credentials')
           .alias('c', 'consumerId').nargs('c', 1)
           .alias('a', 'all')
+          .alias('d', 'deactivated')
           .group(['c', 'a'], 'Options:')
     });
   }
 
   prompting () {
-    const {consumerId, all} = this.argv;
-    return this.admin.credentials.list(consumerId, all)
+    const {consumerId, all, deactivated} = this.argv;
+    let mode = 'activeOnly';
+    if (all) {
+      mode = 'both';
+    } else if (deactivated) {
+      mode = 'deactivatedOnly';
+    }
+    return this.admin.credentials.list(consumerId, mode)
       .then(data => {
         const credentials = data.credentials;
         if (!credentials || !credentials.length) {
--- a/bin/generators/credentials/list.js
+++ b/bin/generators/credentials/list.js
@@ -23,34 +23,36 @@ module.exports = class extends eg.Generator {
           .describe('c', 'Consumer ID: can be User ID or username or app ID')
 
           .array('f')
-          .describe('f', 'Comma separated list of credential state (active, archived), default: active')
+          .describe('f', 'List of credential state (active, archived), default: active')
           .alias('f', 'filter')
+          .default('f', ['active'])
           .coerce('f', (filters) =>
-            filters.reduce((array, filter) => {
+            filters.map((filter) => {
               if (filterTypes.indexOf(filter) === -1) {
                 throw new Error(`Unrecognised filter type: ${filter}`);
               }
               return filtersTable[filter];
-            }, [])
+            })
           )
     });
   }
 
   prompting () {
     const { consumerId, filter } = this.argv;
-
     return this.admin.credentials.list(consumerId)
       .then(data => {
-        let { credentials } = data;
+        const { credentials } = data;
 
         if (!credentials || !credentials.length) {
           this.log.error(`Consumer ${consumerId} has no credentials`);
           return;
         }
 
-        filter.forEach(f => { credentials = credentials.filter(f); });
+        const filteredCredentials = [];
 
-        credentials.forEach(credential => {
+        filter.forEach(f => { filteredCredentials.push(...credentials.filter(f)); });
+
+        filteredCredentials.forEach(credential => {
           if (this.argv.q) {
             this.stdout(credential.id);
           } else {
--- a/bin/generators/credentials/list.js
+++ b/bin/generators/credentials/list.js
@@ -1,5 +1,12 @@
 const eg = require('../../eg');
 
+const filtersTable = {
+  active: (consumer) => consumer.isActive,
+  archived: (consumer) => !filtersTable.active(consumer)
+};
+
+const filterTypes = Object.keys(filtersTable);
+
 module.exports = class extends eg.Generator {
   constructor (args, opts) {
     super(args, opts);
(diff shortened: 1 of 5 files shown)


altsang:
consumerId - would that be the appid and the username (in the case of a user?)

stanislaw-glogowski:
@altsang  I think this PR is not connected with #260. All changes was done in `service` not in `dao`.

bin/generators/credentials/list.js
```diff
@@ -6,33 +6,45 @@ module.exports = class extends eg.Generator {
 
     this.configureCommand({
       command: 'list [options]',
-      description: 'List all credentials for Consumer (User or App)',
+      description: 'List all active credentials for Consumer (User or App)',
```
DrMegavolt: why active ? you can list inactive as well. previous description seems to be better

bin/generators/credentials/list.js
```diff
@@ -6,33 +6,45 @@ module.exports = class extends eg.Generator {
...
+          .alias('a', 'all')
+          .alias('d', 'deactivated')
+          .group(['c', 'a'], 'Options:')
     });
   }
 
   prompting () {
-    return this.admin.credentials.list(this.argv.consumerId)
+    const {consumerId, all, deactivated} = this.argv;
+    let mode = 'activeOnly';
```
DrMegavolt: this will be sent using url, so it is better to have to have lower case params. 

I suggest:
active
disabled or archived

lib/services/credentials/credential.service.js
```diff
@@ -143,8 +143,21 @@ s.getCredential = function (id, type, options) {
...
-  return credentialDao.getAllCredentials(consumerId);
+s.getCredentials = function (consumerId, options = {}) {
+  return credentialDao
+    .getAllCredentials(consumerId)
+    .then((credentials) => {
+      switch (options.mode) {
+        case 'deactivatedOnly':
+          credentials = credentials.filter((credential) => !credential.isActive);
+          break;
+        case 'both':
```
DrMegavolt: `both` is tricky to extend. Imagine you will need some new "mode" parameter, what "both" will mean in this case ? 

`url?include=archived `
`url?include=active, archived `
and default with no params will be `active`

test/cli/credentials/list.test.js
```diff
@@ -1,58 +1,187 @@
...
-        adminHelper.admin.credentials.create(user.username, 'key-auth', {}),
-        adminHelper.admin.credentials.create(user.username, 'basic-auth', {password: 'test'}),
-        adminHelper.admin.credentials.create(user.username, 'oauth2', {secret: 'eg'})
-      ]);
-    })
-    .then(([keyCred1Res, keyCred2Res, basicCredRes, oauth2CredRes]) => {
-      keyCred1 = keyCred1Res;
+
+    return adminHelper
+      .admin
```
DrMegavolt: just curios how it is better from `adminHelper.admin.users.create({  ` ?

test/cli/credentials/list.test.js
```diff
@@ -1,58 +1,187 @@
...
     env.hijack(namespace, generator => {
       generator.once('run', () => {
         generator.log.error = message => {
           done(new Error(message));
         };
         generator.stdout = msg => {
-          let crd = JSON.parse(msg);
-          output[crd.type] = (crd.keyId || crd.secret || crd.password);
+          const crd = JSON.parse(msg);
+          types[crd.type] = (types[crd.type] || 0) + 1;
```
DrMegavolt: didn't get the magic with `|| 0 +1`

stanislaw-glogowski: try add 1 to undefined ... you will get NaN

test/cli/credentials/list.test.js
stanislaw-glogowski: ?

lib/services/credentials/credential.service.js
stanislaw-glogowski: What about CLI? -a [--all] for active, archived and --archived for only archived, what do you think @DrMegavolt

stanislaw-glogowski: Or -i [--include]  active, archived

test/cli/credentials/list.test.js
DrMegavolt: I see new lines, just trying to understand style preference

lib/services/credentials/credential.service.js
DrMegavolt: I like -i approach

test/cli/credentials/list.test.js
DrMegavolt: ok, I get it, you are trying to count number of repeats and after that put key-auth keys separately
as well as this sorting      `createdKeyAuthKeys.deactivated.sort();` and deep equal 

I can live with that. 

From my understanding there is simpler way 
this line `output[crd.type] = (crd.keyId || crd.secret || crd.password);`
could be changed to 
output[crd.type +  (crd.keyId || crd.secret || crd.password)] = true;

so it will be object with props:
"basic-authsomemagichere"
"key-authbase62guid" 

so no sorting, NaN check, all unique 

up to you

lib/services/credentials/credential.service.js
```diff
@@ -143,8 +143,21 @@ s.getCredential = function (id, type, options) {
...
+          break;
+        default: // active only by default
+          credentials = credentials.filter((credential) => credential.isActive);
```
XVincentX: Careful with overriding variables. I'd rather do `return credentials.filter(...`

lib/services/credentials/credential.service.js
```diff
@@ -143,8 +143,26 @@ s.getCredential = function (id, type, options) {
...
-  return credentialDao.getAllCredentials(consumerId);
+s.getCredentials = function (consumerId, options = {}) {
+  return credentialDao
+    .getAllCredentials(consumerId)
+    .then((credentials) => {
+      const include = Array.isArray(options.include) ? options.include : [];
+      include.sort();
+      let filter;
+      switch (include.join(',')) {
+        case 'active,archive':
```
DrMegavolt: ha-ha let's hope nobody will pass arguments in different order 🥇 (I see the sort but no test for different order of arguments)

I assumed that it would output active always and if include is present as "archived" 
it will output everything. because include is kind of add to something you already have.

Now I see the task says to have 3 way, so `filter` may be better word than include. sorry for confusion

test/cli/credentials/list.test.js
```diff
@@ -1,58 +1,187 @@
...
+        createdKeyAuthKeys.all.sort();
+
+        assert.deepEqual(types, createdTypes.all);
+        assert.deepEqual(keyAuthKeys, createdKeyAuthKeys.all);
         done();
       });
     });
 
-    env.argv = program.parse('credentials list -c ' + user.username);
+    env.argv = program.parse('credentials list -i archive,active -c ' + username);
```
DrMegavolt: instead of credentials list -i archive,active -c  username
it should be 
credentials list -i archive -i active -c  username

lib/services/credentials/credential.service.js
DrMegavolt: for the code part, I suggest something explicit: 
```js
const showActive = 'active' in params.filter
const showArchived = 'archived' in params.filter
if (showActive && !showArchived) {
    return credentials.fillter(c=>c.isActive)
}
if (!showActive && showArchived) {
    return credentials.fillter(c=>!c.isActive)
}
return credentials //default

```

test/cli/credentials/list.test.js
stanislaw-glogowski: The problem is that object returned from `adminHelper.admin.credentials.create` does not contain same fields as this one from cli. 

Example (`basic-auth` credential):

`adminHelper.admin.credentials.create`:
```json
{ 
  "isActive": true,
  "createdAt": "Wed Sep 27 2017 15:43:01 GMT+0200 (CEST)",
  "updatedAt": "Wed Sep 27 2017 15:43:01 GMT+0200 (CEST)",
  "id": "6qg7msNHt8xxNCjEnvoXUX"
}
```

CLI:
```json
{ 
  "createdAt": "Wed Sep 27 2017 15:43:01 GMT+0200 (CEST)",
  "isActive": true,
  "password": "$2a$10$mc38mMk1ZtU6qlJbsJ.O7.iCJiefQP3l.NVzr1gcWeqh6map6snLi",
  "updatedAt": "Wed Sep 27 2017 15:43:01 GMT+0200 (CEST)",
  "type": "basic-auth"
}
```

DrMegavolt: ok, I don't mind against this part of the code. 

Important part that you found a bug or even 2. 
1) crd.type should always be filled, for create, get etc. operations
should be one line in  credential service (REST API for create) (Swagger reference promises "type" field everywhere)
 
2) and most likely there is not need for password hash in cli info output 
CLI needs password output when it autogenerates passwords, but after that hash should not be revealed 

I would prefer fix here, but feel free to register  tickets.

stanislaw-glogowski: Done #468

test/cli/credentials/list.test.js
stanislaw-glogowski: In my opinion it's better to read.

lib/rest/routes/credentials.js
```diff
@@ -98,13 +98,22 @@ module.exports = function () {
...
 
   router.get('/:consumerId', function (req, res, next) {
-    credentialSrv.getAllCredentials(req.params.consumerId)
-    .then(credentials => {
-      res.json({credentials});
-    })
-    .catch(err => {
-      next(err);
-    });
+    const {params, query} = req;
```
XVincentX: Nit: I would not de-structure this — it biases the view that this stuff is coming from the request.

lib/services/credentials/credential.service.js
```diff
@@ -144,8 +144,24 @@ s.getCredential = function (id, type, options) {
...
+      }
+      const showActive = filter.indexOf('active') !== -1;
+      const showArchived = filter.indexOf('archived') !== -1;
+      if (showActive && !showArchived) {
+        return credentials.filter(c => c.isActive);
+      }
+      if (!showActive && showArchived) {
+        return credentials.filter(c => !c.isActive);
+      }
+      return credentials;
```
XVincentX: Revise this code — it's confusing.

test/cli/credentials/list.test.js
XVincentX: I agree

lib/rest/routes/credentials.js
```diff
@@ -98,13 +98,22 @@ module.exports = function () {
...
+    let filter;
+    try {
+      filter = query.filter
+        .split(',')
+        .map(value => value.trim().toLowerCase())
+        .filter(value => !!value);
```
XVincentX: Note to self: investigate what this filter is about

lib/rest/routes/credentials.js
```diff
@@ -98,13 +98,22 @@ module.exports = function () {
...
+    } catch (err) {
+      filter = [];
```
XVincentX: Maybe we should just abort the operation?

bin/generators/credentials/list.js
```diff
@@ -6,33 +6,37 @@ module.exports = class extends eg.Generator {
...
           .usage(`Usage: $0 ${process.argv[2]} list [options]`)
           .example(`$0 ${process.argv[2]} list -c 7498d1a9-7f90-4438-a9b7-0ba4c6022353`)
-          .string(['c'])
+          .string(['c', 'f'])
           .describe('c', 'Consumer ID: can be User ID or username or app ID')
+          .describe('f', 'Comma separated list of credential state (active, archived), default: active')
           .alias('c', 'consumerId').nargs('c', 1)
-          .group(['c'], 'Options:')
+          .alias('f', 'filter')
+          .group(['c', 'i'], 'Options:')
```
XVincentX: Note to self: what's the `i` option for

bin/generators/credentials/list.js
```diff
@@ -1,41 +1,43 @@
...
           .example(`$0 ${process.argv[2]} list -c 7498d1a9-7f90-4438-a9b7-0ba4c6022353`)
-          .string(['c'])
+          .string(['c', 'f'])
           .describe('c', 'Consumer ID: can be User ID or username or app ID')
-          .alias('c', 'consumerId').nargs('c', 1)
-          .group(['c'], 'Options:')
+          .describe('f', 'Comma separated list of credential state (active, archived), default: active')
+          .alias('c', 'consumerId').nargs('c', 1).required('c')
+          .alias('f', 'filter')
+          .group(['c', 'i'], 'Options:')
```
XVincentX: @DrMegavolt Do you know what's this `i` option is about? I can't find any reference.

bin/generators/credentials/list.js
DrMegavolt: see comment in slack, feel free to implement differently

XVincentX:
@kevinswiber @DrMegavolt 

I've been thinking about this during the weekend and I was wondering whether we should simply _NOT_ include any filtering capabilities on the server, and just rely on the client (CLI, in our case) to do that.

As out WebAPI is already consumer based, I do not think we're gonna never return that much data to blown up the client.

Let me know what you guys think, and I can finalize the PR according to the decision we'll take.

XVincentX:
Very good! I'll change the code right now!

XVincentX:
@DrMegavolt Done in fe4edac 44a8719 and ae30752

PR #430, opened 2017-09-22
Dockerfile builds an Alpine-based image with a basic express-gateway

rlisagor:
The Dockerfile takes the version and scaffold type as a build argument, so it can be used to generate other instance types.

Two script targets were also added to `package.json` to automatically create images for `basic` and `getting-started`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,17 @@
-FROM node:latest
+FROM node:alpine
+
+ARG VERSION
+ARG TYPE=basic
+
+ENV NODE_ENV production
 
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
-COPY package.json /usr/src/app/
-RUN npm install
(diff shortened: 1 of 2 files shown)


kevinswiber:
Thanks, @rlisagor!

Dockerfile
```diff
@@ -1,11 +1,17 @@
...
 
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
-COPY package.json /usr/src/app/
-RUN npm install
-COPY . /usr/src/app
+RUN npm install express-gateway@$VERSION && \
+    ./node_modules/.bin/eg gateway create -n gateway -d . -t $TYPE && \
+    npm cache clean --force
```
XVincentX: @rlisagor Why the cache clean?

rlisagor: This is to reduce the size of the resulting image as much as possible.

Issue #361, opened 2017-08-02
[bugs] CLI: eg credentials list -c <userId>


ahwing210:
Hi EG Team,

I follow the [getting started guide](http://www.express-gateway.io/getting-started/) and try some other CLI commands.
The CLI command "**eg credentials list**" can show the user credentials with "**User username**".
However, it seems that the CLI command "**eg credentials list**" cannot show the user credentials with "**User Id**". 

Please find below screenshot for trying..

1. eg credentials list -c 795872c9-6c01-410c-9dcd-395b353befd9
**# not successful**
2. eg credentials list -c bob
**# successful**

![image](https://user-images.githubusercontent.com/15073442/28861032-9e3e9352-7791-11e7-9e7c-b9e953548a2e.png)

DrMegavolt:
@ahwing210 Thank you for finding issue. 
Indeed new credential for user  will be linked to whatever is provided as consumer: user id or username. and this should not be that way. 

if used as `eg users create -c username ....` then 'eg credentials list -c username' will work  
if used as `eg users create -c userid ....` then 'eg credentials list -c userid' will work 

while quick way to fix exists,  it maybe not that simple to fix root cause, can take some time 

for now in key-auth scenario  please use userId

DrMegavolt:
Is related to https://docs.google.com/document/d/15s3njiBfbfiwvwPuylcYfJlG4fvm_QJDJ54IKakyUiI/edit#heading=h.igsojut9cpcj

altsang:
@XVincentX - please assume new - please write up spec first

XVincentX:
Guys — I've been looking at the code and here are my considerations:

Both `username` and `id` are unique for an `user`. This means that one implies the other and viceversa. That's redundant.

If this affirmations holds with the passing of the time - and I think it will - I personally do not see any reason to have this duplicate value. 

I'd rather promote `username` as the `id` of the object and simply remove the `uuid` generated by the gateway, so the only way to look for an user would be by `id` - that is the username.

Regardless I'm right on the uniqueness of the things or not, I would prefer that the gateway itself (and the exposed webapi) would only deal with the `id`. All these byproducts (such as get by username, filtering things and so on and so forth) should be implemented in the CLI - in the same way we did with #428 : [filtering](https://github.com/ExpressGateway/express-gateway/blob/master/bin/generators/credentials/list.js#L3) is happening on the CLI.

What's your idea?

DrMegavolt:
[...]

PR #431, opened 2017-09-25
Feature/friendly app names

stanislaw-glogowski:
Connected #260 

The following pull request changes application service and dao. 
 - ~~when application has `name` property, additional `unique name` validator is added in insert method~~
- ~~`name` can be use as a friendly application `id`~~
- `name` and `userId` as a option in CLI app list:
```bash
$ eg apps list [options]
# options:
# -n --name - Application name
# -u --userId - Application User ID
```

### DIFF
--- a/lib/admin/client.js
+++ b/lib/admin/client.js
@@ -1,5 +1,4 @@
-const _ = require('lodash');
-const superagent = require('superagent');
+let superagent = require('superagent');
 
 module.exports = function ({baseUrl, verbose, headers}) {
   if (baseUrl[baseUrl.length - 1] !== '/') { // ensure trailing slash
@@ -32,14 +31,13 @@ module.exports = function ({baseUrl, verbose, headers}) {
   let client = new Proxy(superagent, {
     get: function (request, name) {
       if (methods.indexOf(name) >= 0) {
-        const method = name === 'del' ? 'delete' : name;
-        return (url, query = {}) => {
-          const queryString = _.map(query, (value, key) => `${key}=${encodeURIComponent(value)}`).join('&');
-          const currentUrl = baseUrl + url + (queryString ? `?${queryString}` : '');
-          debug('>', method.toUpperCase(), currentUrl);
-          const relativeUrlRequest = new Proxy(request(method, currentUrl), wrapHandler);
+        let method = name === 'del' ? 'delete' : name;
+        return (url) => {
+          debug('>', method.toUpperCase(), baseUrl + url);
+          let relativeUrlRequest = new Proxy(request(method, baseUrl + url), wrapHandler);
           if (headers) {
-            return relativeUrlRequest.set(prepareHeaders(headers));
+            let headersToAdd = prepareHeaders(headers);
+            return relativeUrlRequest.set(headersToAdd);
           }
           return relativeUrlRequest;
         };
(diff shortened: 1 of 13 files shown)


DrMegavolt:
btw, I like removing magic dao and s objects. 
the only comment is next time please make this refactoring as separate commit. 
now I see entire file changed = hard to see real changes. 

if it was first refactor and then changes required by tasks I would be able to compare between intermediate commits, to see real changes vs refactoring

test/cli/apps/list.test.js
```diff
@@ -52,21 +64,108 @@ describe('eg apps list', () => {
...
       generator.once('end', () => {
-        assert.ok(output[app1.name]);
-        assert.ok(output[app1.id]);
-        assert.ok(output[app2.name]);
-        assert.ok(output[app2.id]);
+        const filteredApps = apps.filter((app) => (
+          app.userId === user1.id &&
+          app.name === 'app1'
+        ));
+        assert.equal(_.size(output), filteredApps.length);
```
DrMegavolt: How it is possible to have for the same user app with the same name ? 


and in general `_.size(output)` could be Object.keys(output).length with no dependencies 
https://github.com/lodash/lodash/blob/master/size.js#L40

lib/admin/client.js
```diff
@@ -31,13 +32,14 @@ module.exports = function ({baseUrl, verbose, headers}) {
   let client = new Proxy(superagent, {
     get: function (request, name) {
       if (methods.indexOf(name) >= 0) {
-        let method = name === 'del' ? 'delete' : name;
-        return (url) => {
-          debug('>', method.toUpperCase(), baseUrl + url);
-          let relativeUrlRequest = new Proxy(request(method, baseUrl + url), wrapHandler);
+        const method = name === 'del' ? 'delete' : name;
+        return (url, query = {}) => {
+          const queryString = _.map(query, (value, key) => `${key}=${encodeURIComponent(value)}`).join('&');
```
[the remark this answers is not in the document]
DrMegavolt: And btw, the reasons for this client are:
- set urls as relative path. 
- debug info in CURL style
- automatically include headers from config (main use case - authorization)

test/cli/apps/list.test.js
stanislaw-glogowski: @DrMegavolt the question is should we use lodash ...

DrMegavolt: I believe situation is the same as bluebird. there is only one function (_.omit) usage across the code (3 places) that has no native analog. and can be replaced as small module - https://www.npmjs.com/package/lodash.omit

and some weird ` _.map` in test usage that can be just changed to standard map\for..in\for..of

DrMegavolt: update: lodash.omit module is just an export from ldoash. still has lodash as dependency, but the idea is still the same - it can be replaced

test/cli/apps/update.test.js
```diff
@@ -37,37 +36,6 @@ describe('eg apps update', () => {
       app1 = createdApp;
     });
   });
-  it('updates an app from prompts', done => {
```
DrMegavolt: why it was removed ?

stanislaw-glogowski:
@kevinswiber , @DrMegavolt - can we close that PR without merge, I'll start it again on current, updated master branch. 
In my opinion this task depends not only on #260 but also #361

DrMegavolt:
it is your PR, feel free to close it any time. 

btw, if you have latest master locally
`git rebase master` (executed from current branch) will rebase your code on top of it.

you have 17 commits, you can squash them to make rebase simpler 
`get rebase -i HEAD~17`
it will be text editor to squash everything

PR #439, opened 2017-09-28
rm lodash direct dependency

DrMegavolt:
Same idea as bluebird. 
Those libs had sense in jQuery world. now you can complete all operations in native js

BTW  _.omit -  is deprecated in lodash 5.0 due to high performance cost
connect #437

### DIFF
--- a/lib/policies/oauth2/oauth2.js
+++ b/lib/policies/oauth2/oauth2.js
@@ -1,11 +1,10 @@
 'use strict';
-
+const omitedProps = ['tokenDecrypted'];
 const passport = require('passport');
 const LocalStrategy = require('passport-local').Strategy;
 const ClientPasswordStrategy = require('passport-oauth2-client-password').Strategy;
 const BearerStrategy = require('passport-http-bearer').Strategy;
 
-const _ = require('lodash');
 const services = require('../../services/index');
 const authService = services.auth;
 
@@ -84,7 +83,7 @@ function authenticateToken (req, accessToken, done) {
 
             delete req.headers['authorization'];
             setHeaders(req, consumer, 'consumer');
-            setHeaders(req, _.omit(token, ['tokenDecrypted']), 'token');
+            setHeaders(req, token, 'token');
             return done(null, consumer);
           }
 
@@ -96,7 +95,7 @@ function authenticateToken (req, accessToken, done) {
 
               delete req.headers['authorization'];
               setHeaders(req, consumer, 'consumer');
-              setHeaders(req, _.omit(token, ['tokenDecrypted']), 'token');
+              setHeaders(req, token, 'token');
 
               req.egContext.consumer = consumer;
               req.egContext.token = token;
@@ -172,6 +171,13 @@ module.exports = function () {
 
 function setHeaders (req, headerObject, namespace) {
   for (let key in headerObject) {
+    // TODO: this is proxy concern to set HEADERS,
+    // in this policy we should only declare information that somehow can be sent downstream
+    // (most reasonable way is indeed use heders, but it is not the only option)
+    // some properties should not be proxied due to security reasons
+    if (omitedProps.some(p => p === key)) {
+      continue;
+    }
     const val = headerObject[key];
     if (val) {
       key = key.replace(/([A-Z])/g, (g) => `-${g[0]}`); // convert camelcase to hyphens
--- a/lib/services/consumers/user.service.js
+++ b/lib/services/consumers/user.service.js
@@ -3,7 +3,6 @@
 const userDao = require('./user.dao.js');
 const applicationService = require('./application.service.js');
 const credentialService = require('../credentials/credential.service.js');
-const _ = require('lodash');
 const config = require('../../config');
 const utils = require('../utils');
 const uuidv4 = require('uuid/v4');
@@ -36,7 +35,10 @@ s.get = function (userId, options) {
     }
 
     user.isActive = user.isActive === 'true';
-    return (options && options.includePassword) ? user : _.omit(user, ['password']);
+    if (!options || !options.includePassword) {
+      delete user.password;
+    }
+    return user;
   });
 };
 
@@ -66,8 +68,10 @@ s.update = function (userId, _props) {
   }
   return this.get(userId) // validate user exists
   .then(user => {
-    return !user ? false // user does not exist
-    : validateUpdateToUserProperties(_.omit(_props, ['username']))
+    if (!user) { return false; } // user does not exist
+
+    delete _props.username;
+    return validateUpdateToUserProperties(_props)
     .then(function (updatedUserProperties) {
       if (updatedUserProperties) {
         utils.appendUpdatedAt(updatedUserProperties);
@@ -119,17 +123,20 @@ s.remove = function (userId) {
 
 function validateAndCreateUser (_user) {
   let user;
-  if (!_user && !_user.username) {
+  const username = _user.username;
+  if (!_user && !username) {
     return Promise.reject(new Error('invalid user object')); // TODO: replace with validation error
   }
 
-  return s.find(_user.username) // Ensure username is unique
+  return s.find(username) // Ensure username is unique
   .then(function (exists) {
-    return !exists ? validateNewUserProperties(_.omit(_user, ['username']))
-      : Promise.reject(new Error('username already exists')); // TODO: replace with validation error
+    if (exists) {
+      throw new Error('username already exists');
+    }
+    return validateNewUserProperties(_user);
   })
   .then(function (newUser) {
-    const baseUserProps = { isActive: 'true', username: _user.username, id: uuidv4() };
+    const baseUserProps = { isActive: 'true', username, id: uuidv4() };
     if (newUser) {
       user = Object.assign(newUser, baseUserProps);
     } else user = baseUserProps;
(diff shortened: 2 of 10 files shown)


DrMegavolt:
connect #437

lib/policies/oauth2/oauth2.js
```diff
@@ -172,6 +171,9 @@ module.exports = function () {
 
 function setHeaders (req, headerObject, namespace) {
   for (let key in headerObject) {
+    if (key === 'tokenDecrypted') {
```
XVincentX: I do not really like to couple this specific logic here. `setHeaders` should just set the headers. Providing the right ones should be a caller responsibility.

lib/services/credentials/credential.service.js
```diff
@@ -95,7 +94,7 @@ s.insertCredential = function (id, type, credentialDetails) {
 
       return Promise.all([validateNewCredentialScopes(credentialConfig, credentialDetails),
         validateAndHashPassword(credentialConfig, credentialDetails),
-        validateNewCredentialProperties(credentialConfig, _.omit(credentialDetails, ['scopes']))
+        validateNewCredentialProperties(credentialConfig, credentialDetails)
```
XVincentX: I guess it's just ok to pass this additional data to this function

test/services/authorization-codes.test.js
```diff
@@ -2,23 +2,15 @@ let mock = require('mock-require');
...
-    db.flushdbAsync()
-    .then(function (didSucceed) {
-      if (!didSucceed) {
-        console.log('Failed to flush the database');
-      }
-      done();
-    })
-    .catch(done);
+  before(function () {
+    return db.flushdbAsync();
```
XVincentX: 👍 👍 👍 

You can also do

`before(db.flushdbAsync)`

test/services/authorization-codes.test.js
```diff
@@ -35,7 +27,11 @@ describe('Authorization Code Tests', function () {
       should.exist(code.id);
       code.id.length.should.be.greaterThan(15);
       should.ok(new Date(code.expiresAt) > Date.now());
-      _.omit(code, [ 'id', 'createdAt', 'expiresAt' ]).should.deepEqual(newCode);
+      let filteredCode = Object.assign({}, code);
+      delete filteredCode.id;
+      delete filteredCode.createdAt;
+      delete filteredCode.expiresAt;
+      filteredCode.should.deepEqual(newCode);
```
XVincentX: If I understand correctly, the aim of this test is to make sure some properties and values are contained the returned object and not really to be deep equal
We can probably use [should.have.properties](http://shouldjs.github.io/#assertion-properties) instead of removing the properties by hand.

test/services/tokens.test.js
```diff
@@ -329,15 +328,14 @@ describe('Access Token tests', function () {
...
           tokens.length.should.eql(tokenObjs.length * 2);
           tokens.forEach(tokenObj => {
-            tokenObj.prop.should.be.oneOf(_.map(tokenObjs, 'prop'));
+            tokenObj.prop.should.be.oneOf(tokenObjs.map(x => x.prop));
           });
-          _.map(tokens, 'archived').filter(val => val).length.should.eql(4);
+          tokens
+          .map(x => x.archived)
+          .filter(val => val).length.should.eql(4);
           done();
```
XVincentX: You can return the Promise directly and remove the `done` parameter

test/services/tokens.test.js
```diff
@@ -329,15 +328,14 @@ describe('Access Token tests', function () {
           should.exist(tokens);
           tokens.length.should.eql(tokenObjs.length * 2);
           tokens.forEach(tokenObj => {
-            tokenObj.prop.should.be.oneOf(_.map(tokenObjs, 'prop'));
+            tokenObj.prop.should.be.oneOf(tokenObjs.map(x => x.prop));
           });
-          _.map(tokens, 'archived').filter(val => val).length.should.eql(4);
+          tokens
+          .map(x => x.archived)
+          .filter(val => val).length.should.eql(4);
```
XVincentX: What's the aim behind `filter(val => val)`?

test/services/tokens.test.js
```diff
@@ -329,15 +328,14 @@ describe('Access Token tests', function () {
           should.exist(tokens);
           tokens.length.should.eql(tokenObjs.length * 2);
           tokens.forEach(tokenObj => {
-            tokenObj.prop.should.be.oneOf(_.map(tokenObjs, 'prop'));
+            tokenObj.prop.should.be.oneOf(tokenObjs.map(x => x.prop));
           });
-          _.map(tokens, 'archived').filter(val => val).length.should.eql(4);
+          tokens
+          .map(x => x.archived)
```
XVincentX: Given you're testing the length of the returned object, there's no need to use the `map` function.

test/services/users.test.js
```diff
@@ -37,28 +26,20 @@ describe('User service tests', function () {
...
-          username: user.username,
-          email: user.email,
-          isActive: true
-        };
         Object.keys(newUser).should.eql(expectedUserProps);
-        should.deepEqual(_.omit(newUser, ['id', 'createdAt', 'updatedAt']), expectedPartialUserObj);
-        db.hgetallAsync(redisConfig.namespace.concat('-', 'user').concat(':', newUser.id))
+        newUser.firstname.should.eql(user.firstname);
+        newUser.lastname.should.eql(user.lastname);
+        newUser.username.should.eql(user.username);
```
XVincentX: dtto - [Assertion properties](http://shouldjs.github.io/#assertion-properties) are probably a better tool for this.

lib/policies/oauth2/oauth2.js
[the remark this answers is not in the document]
DrMegavolt: the logic of this method is to transform headers for proxying. And some of the properties MUST NOT be transferred. So this place is the right one to protect from that part. 

I can create some const object to define those properties that need to be filtered out. 
will it work ?

test/services/tokens.test.js
DrMegavolt: haha, no idea, just copy paste 👍

test/services/tokens.test.js
DrMegavolt: 👍

test/services/users.test.js
DrMegavolt: 👍

test/services/authorization-codes.test.js
DrMegavolt: 👍

test/services/authorization-codes.test.js
DrMegavolt: 👍

lib/services/credentials/credential.service.js
DrMegavolt: yes, I added scopes filtering to the function itself, it was 5-6 places with _.omit scopes.

lib/policies/oauth2/oauth2.js
```diff
@@ -84,7 +83,7 @@ function authenticateToken (req, accessToken, done) {
 
             delete req.headers['authorization'];
             setHeaders(req, consumer, 'consumer');
-            setHeaders(req, _.omit(token, ['tokenDecrypted']), 'token');
```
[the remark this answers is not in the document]
DrMegavolt: see setHeaders, it is constant pattern all over the code. it is better to have it centralized

lib/services/consumers/user.service.js
```diff
@@ -119,17 +123,20 @@ s.remove = function (userId) {
...
 
-  return s.find(_user.username) // Ensure username is unique
+  return s.find(username) // Ensure username is unique
   .then(function (exists) {
-    return !exists ? validateNewUserProperties(_.omit(_user, ['username']))
-      : Promise.reject(new Error('username already exists')); // TODO: replace with validation error
+    if (exists) {
+      return Promise.reject(new Error('username already exists'));
+    }
+    return validateNewUserProperties(_user);
```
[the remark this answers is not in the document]
DrMegavolt: yes, it has filtering for username inside, same as setHeaders above.

test/services/authorization-codes.test.js
DrMegavolt: actually it doesn't work that way

test/services/tokens.test.js
DrMegavolt: actually it is required, it is checking only archived tokens, and ensures that the tokens are indeed archived.

lib/services/consumers/user.service.js
```diff
@@ -160,11 +167,15 @@ function validateUpdateToUserProperties (userProperties) {
 function validateNewUserProperties (userProperties) {
   let newUserProperties = {};
 
-  if (!Object.keys(userProperties).every(key => (typeof key === 'string' && !!config.models.users.properties[key]))) {
+  if (!Object.keys(userProperties).filter(key => key !== 'username').every(key => !!config.models.users.properties[key])) {
```
stanislaw-glogowski: You can do it in one loop
```js
Object.keys(userProperties).every(key => key ==='username' || !!config.models.users.properties[key])
```

DrMegavolt: filter returns array, every returns boolean. 
the code is inside if. it will require adding some another code like ....length

lib/policies/oauth2/oauth2.js
```diff
@@ -172,6 +171,13 @@ module.exports = function () {
 
 function setHeaders (req, headerObject, namespace) {
```
XVincentX: How about `filterHeaders` instead of `setHeaders`? That really gives a clear idea of what's happening.

test/services/authorization-codes.test.js
XVincentX: No? That's weird 🤔

PR #450, opened 2017-09-29
Use bcryptjs instead of bcrypt

agonbina:
The issue here is that bcrypt will try to build its binary and somehow this breaks when using the official `node:*` docker images. (possibly related issue: https://github.com/kelektiv/node.bcrypt.js/issues/473)


DrMegavolt:
@agonbina can you provide docker file ? 
we tested with node:slim and node:alpine it works. because it has prebuilt libraries for linux x64 https://github.com/kelektiv/node.bcrypt.js/releases

How it is possible you have it not building ?

XVincentX:
@agonbina 
In the worst case you can always use a full docker image to build/install the gateway and then, using the [multi stage feature](https://docs.docker.com/engine/userguide/eng-image/multistage-build/), run your instance using a very small docker image (such as alpine or slim).

Let us know if that solves the problem eventually.

DrMegavolt:
ok, use case found  (gitter with @srcnix )
there is no prebuilt version for Linux 32bit

DrMegavolt:
@agonbina could you have the following logic (or let us know, we can change it ourselves):

1) bcrypt C++ as *optional* dependency 
2) bcryptJS as dependency
3)
```js
let bcrypt; 
try {bcrypt = require('bcrypt');  }
catch (){
    bcrypt = require('bcryptjs'); 
}
```

It will work in C++ version for Linux64, MacOS, Windows and in JS mode for other linux versions   
and will not require complex docker build process

agonbina:
@DrMegavolt sorry for not following up. This is the best solution 👍

pablocarreraest:
@agonbina @DrMegavolt I try to install on node:8-alpine with no success so I move on with my life to use bcryptjs! is that the best way to solve?

DrMegavolt:
@pablocarreraest current version of Express Gateway is switching to js version if native cannot be installed. And this is ok, it happens on 32bit linux. 

However And build tools are not required for other systems, because brcypt has precompiled binaries shipped with package

So it should work with docker node:8-alpine

We are using this specific image to build official container
https://hub.docker.com/r/expressgateway/express-gateway/

PR #453, opened 2017-09-30
Provide plugin typings

### LINKED ISSUE #433: Provide a Typescript interface for plugin framework
Although we have a [great documentation](http://www.express-gateway.io/docs/plugins/plugin-development) for developing a plugin, it might be worth and lowering the barriers for people willing to write a plugin to provide a typescript definition - even people using javascript would benefit it, if using the right [IDE](https://code.visualstudio.com).

XVincentX:
Connects #433 closes #433 

This pull request provides basic Typings for the exposed plugin interface. People working with Typescript or even Javascript will receive a great help while writing code.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,11 +2,8 @@ environment:
   matrix:
     - nodejs_version: "6"
     - nodejs_version: "8"
-cache:
-  - node_modules
 install:
   - ps: Install-Product node $env:nodejs_version
-  - if exist node_modules npm rebuild
   - npm install
 build: off
 test_script:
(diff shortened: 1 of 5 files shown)


XVincentX:
Yeah - I was thinking about the JWT one

DrMegavolt:
I don't agree with JWT, because this is extremely popular and will be forked a lot. So having it TS is a barrier for that. 

There are some smaller examples like web sockets, advanced logger (with connectors to ELK etc.). They can be TS no problem, as well as ES6 style with import\export

XVincentX:
All right - it seems reasonable.

Issue #448, opened 2017-09-29
Common error response


DrMegavolt:
build a helper to express.js Response to send errors in standard format

DrMegavolt:
spec first, to agree on standard format of API output

DrMegavolt:
non relevant anymore

Issue #185, opened 2017-06-20
Core: Create central Error classes


IrfanBaqui:

DrMegavolt:
closing in favor of https://github.com/ExpressGateway/express-gateway/issues/448

Issue #63, opened 2017-05-05
Core: Define Errors library


IrfanBaqui:

DrMegavolt:
closing in favor of https://github.com/ExpressGateway/express-gateway/issues/448

PR #459, opened 2017-10-03
Feature/common error response

stanislaw-glogowski:
Connected #448 

The following pull request adds default error handle middleware into gateway app. 
It also extend req object with `egError` method:
```javascript
/**
 * send error 
 * @param code / error / errors
 * @param [errors]
 */
res.egError = (...args) => {
  next(new GatewayError(...args));
};
```
Error response format:
```
{
  "code": number,
  "errors": []string
}
```

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -1,56 +1,50 @@
 const statuses = require('statuses');
 
-class CustomError extends Error {
-  constructor (message) {
-    super();
-    this.message = message;
+class BaseError extends Error {
+  constructor (message = '') {
+    super(message);
     this.name = this.constructor.name;
     Error.captureStackTrace(this, this.constructor);
   }
 }
 
-class ConfigurationError extends CustomError {}
+class ConfigurationError extends BaseError {}
 
-class GatewayError extends CustomError {
-  constructor (code = 500, errors = null) {
+class GatewayError extends BaseError {
+  constructor (...args) {
     super();
-    this.code = 500;
-    this.error = null;
-    this.errors = [];
-    switch (typeof code) {
-      case 'number':
-        if (code >= 400 && code < 600) {
-          this.code = code;
-        }
-        break;
-      default:
-        errors = code;
-    }
-    if (Array.isArray(errors)) {
-      this.errors = errors.filter((error) => error && typeof error === 'string');
-    } else if (typeof errors === 'string') {
-      this.error = errors;
-    } else if (errors instanceof Error) {
-      this.error = errors.message;
-    }
-  }
 
-  get response () {
-    const {code, errors, error} = this;
-    const result = {code};
-    if (errors.length) {
-      Object.assign(result, {errors});
-    } else {
-      Object.assign(result, {
-        error: error || statuses[code]
-      });
+    let [code, errors] = args;
+    if (typeof code !== 'number' || code < 400 || code >= 600) {
+      errors = code;
+      code = 500;
     }
-    return result;
+
+    if (!Array.isArray(errors)) {
+      errors = [errors];
+    }
+
+    errors = errors
+      .map((error) => {
+        if (error instanceof Error) {
+          error = error.message;
+        }
+        return error;
+      })
+      .filter(error => error && typeof error === 'string');
+
+    if (!errors.length) {
+      errors.push(statuses[code]);
+    }
+
+    this.response = {
+      code,
+      errors
+    };
   }
 }
 
 module.exports = {
-  CustomError,
   ConfigurationError,
   GatewayError
 };
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -1,17 +1,22 @@
-'use strict';
-
-const util = require('util');
-
-function CustomError (message) {
-  Error.captureStackTrace(this, this.constructor);
-  this.name = this.constructor.name;
-  this.message = message;
+class CustomError extends Error {
+  constructor (message) {
+    super();
+    this.message = message;
+  }
 }
-util.inherits(CustomError, Error);
 
 class ConfigurationError extends CustomError {}
 
+class GatewayError extends CustomError {
+  constructor (code = 500, errors = []) {
+    super(code);
+    this.code = code;
+    this.errors = errors;
+  }
+}
+
 module.exports = {
   CustomError,
-  ConfigurationError
+  ConfigurationError,
+  GatewayError
 };
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -1,17 +1,51 @@
+const statuses = require('statuses');
+
 class CustomError extends Error {
   constructor (message) {
     super();
     this.message = message;
+    this.name = this.constructor.name;
+    Error.captureStackTrace(this, this.constructor);
   }
 }
 
 class ConfigurationError extends CustomError {}
 
 class GatewayError extends CustomError {
-  constructor (code = 500, errors = []) {
-    super(code);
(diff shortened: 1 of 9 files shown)


lib/errors.js
```diff
@@ -1,17 +1,56 @@
...
+    this.message = message;
+    this.name = this.constructor.name;
+    Error.captureStackTrace(this, this.constructor);
+  }
 }
-util.inherits(CustomError, Error);
 
 class ConfigurationError extends CustomError {}
 
+class GatewayError extends CustomError {
```
DrMegavolt: Why do we need this custom Error ?

lib/errors.js
```diff
@@ -1,17 +1,56 @@
...
+          this.code = code;
+        }
+        break;
+      default:
+        errors = code;
+    }
+    if (Array.isArray(errors)) {
+      this.errors = errors.filter((error) => error && typeof error === 'string');
+    } else if (typeof errors === 'string') {
+      this.error = errors;
```
DrMegavolt: can it be always `errors` as array. otherwise consumer will have to check 2 fields

lib/errors.js
stanislaw-glogowski: Idk, it was here before.

lib/errors.js
stanislaw-glogowski: I'm thinking about changing it to:
```
{
  "errno": number,
  "error": string, 
  "data": any // as an option for eg. validation 
}
```

lib/policies/rate-limit/rate-limit.js
```diff
@@ -11,6 +11,11 @@ module.exports = (params) => {
       }
     };
   }
+
+  params.handler = (req, res) => {
+    res.egError(429);
```
DrMegavolt: make sense, 
could you please add error message, as I understand result will be just {code:429} ?  

not sure if we need a header, up to you 
https://github.com/nfriedly/express-rate-limit/blob/master/lib/express-rate-limit.js#L25

stanislaw-glogowski: Result will be:
```
{
  "code": 429,
  "error": "Too Many Requests"
}
```

When message is empty, it gets default http status message from `statuses` lib.

lib/errors.js
```diff
@@ -1,17 +1,56 @@
...
-const util = require('util');
-
-function CustomError (message) {
-  Error.captureStackTrace(this, this.constructor);
-  this.name = this.constructor.name;
-  this.message = message;
+class CustomError extends Error {
+  constructor (message) {
+    super();
+    this.message = message;
```
XVincentX: I think you should feed the `super` with the `message` parameter instead of storing the message by yourself.

`super(message);`

lib/errors.js
```diff
@@ -1,17 +1,56 @@
...
+    Error.captureStackTrace(this, this.constructor);
+  }
 }
-util.inherits(CustomError, Error);
 
 class ConfigurationError extends CustomError {}
 
+class GatewayError extends CustomError {
+  constructor (code = 500, errors = null) {
+    super();
```
XVincentX: Shouldn't `super()` be called with a valid message?

lib/gateway/errors.js
```diff
@@ -0,0 +1,28 @@
...
+function errorHelper (req, res, next) {
+  res.egError = (code = 500, errors = []) => {
+    next(new GatewayError(code, errors));
+  };
+
+  next();
+}
+
+function notFoundHandler (req, res, next) {
+  res.egError(404);
```
XVincentX: This seems a code smell to me - I have a `next` parameter that's not being called - but in the reality is being called by the inner method 🤔

lib/errors.js
stanislaw-glogowski: In my opinion it should be fixed in #63, #185

lib/gateway/errors.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+function errorHandler (err, req, res, next) {
+  if (!(err instanceof GatewayError)) {
```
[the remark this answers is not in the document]
stanislaw-glogowski: This middleware handle errors and response json object. We need to create another middleware for errors logging.

Something like this one:
```js
function logErrors (err, req, res, next) {
  if (!(err instanceof GatewayError)) {
    logger.error(err);
  }
  next(err);
}
```

lib/gateway/errors.js
stanislaw-glogowski: Fixed, I removed next

lib/errors.js
```diff
@@ -1,17 +1,50 @@
...
+const statuses = require('statuses');
 
-const util = require('util');
-
-function CustomError (message) {
-  Error.captureStackTrace(this, this.constructor);
-  this.name = this.constructor.name;
-  this.message = message;
+class BaseError extends Error {
+  constructor (message = '') {
```
DrMegavolt: no need for default param

lib/gateway/index.js
```diff
@@ -95,6 +99,9 @@ function bootstrap ({plugins, config} = {}) {
     rootRouter(req, res, next);
   });
 
+  app.use(notFoundHandler);
```
DrMegavolt: do I understand correctly that it will always return 404  ignoring any error ? 
because it is just
```
function notFoundHandler (req, res) {
   res.egError(404);
 }
```

DrMegavolt:
It should be  tests for 422\404\500 use cases 
and it should also check that multiple error messages provided

lib/gateway/index.js
stanislaw-glogowski: No, notFoundHandler middleware will only be called when there is no error. To catch error you need to create middleware with `err, req, res, next` attributes.

DrMegavolt: ah, yep, thx

DrMegavolt:
I was looking for  some standards around object processing 
http://jsonapi.org/format/#error-objects

so they suggest to have `errors` as array of *objects* to provide more info, not just string
`{errors:[{...}] ...}` 

http://jsonapi.org/format/#document-top-level
and return 
```{
data: ...
errors: ...
meta: ....
}```
as result for general request

XVincentX:
I've been working with this in the past - so far it has been working in all the use cases I needed.
What do you guys think?


https://github.com/blongden/vnd.error

DrMegavolt:
vnd.error (some weird content-type)seems to impose HAL convention 

Most common API styles comparison 
https://blogs.mulesoft.com/dev/api-dev/api-best-practices-hypermedia-part-3/
there is @kevinswiber Siren vs HAL   https://github.com/kevinswiber/siren/issues/15 discussion

@kevinswiber we may need to have a vortex around that. to understand pros and cons

kevinswiber:
There's also this: https://tools.ietf.org/html/rfc7807

altsang:
sorry guys - jumping into this late - anytime I see someting "commodity" my first question is what module/ middleware are we using to leverage?

in this case - `statuses`? anything else?

stanislaw-glogowski:
@altsang in this case we use only `statuses` which is already in expressjs deps

stanislaw-glogowski:
@DrMegavolt @kevinswiber @altsang @XVincentX Maybe we can add somehow `error response model / template` to our gateway configuration? Any advice?

DrMegavolt:
Closing for now. We need to agree on approach first before implementation.

PR #456, opened 2017-10-01
Fix flaky test + add node 6 build

XVincentX:
The following PR will change the `circleci` file in order to leverage the new 2.0 service - we're now running the build in parallel for node 6 and node 8.

Rebase on top on #455 

Connect #463 closes #463

PR #461, opened 2017-10-04
fixed to use superagent from EG;

DrMegavolt:
Originally I was thinking to have individual dependency on super agent, but since it is not a separate package yet, it should use same version as EG


XVincentX:
This is now part of #456

Issue #422, opened 2017-09-21
Feature: programmatic startup/shutdown improvements


XVincentX:
While writing integration tests for plugin I wish I had 3 main features in the gateway:

1. A way to shutdown a gateway instance (mainly for testing purposes, when you want to cleanup the resources).

`gatewayInstance.close().then(()=>console.log("It's over"));`

2. A way to know that the gateway is ready to get requests; the `run` method might return a Promise

`gatewayInstance.run().then(()=>console.log("Time to proxy requests"));`

3. A way to feed the gateway configuration from the code directly, and not being constrained to give file paths or directories.

```javascript
const systemConfig = {};
const gatewayConfig = {}

gateway().load(systemConfig, gatewayConfig).run().then(...)
```

DrMegavolt:
and config from code is not a problem gateway({config})

kevinswiber:
One issue here is that configuration is a global singleton.  If you have multiple `gateway().load()` calls, the last config will win.

XVincentX:
@DrMegavolt 

I’ve been looking to the helpers you provided and I realized that they’re basically miming somebody starting the gateway from the command line - and then throw some requests and wait the gateway to respond correctly in order to make sure it’s ready.

My aim is more - let’s provide a builtin mechanism to start and stop the gateway, and I’ve started look into this. It turns out we can do that kind of easily.

I've been playing around with a branch here:
https://github.com/ExpressGateway/express-gateway/compare/feature/programmatic-startup-shutdown?expand=1

The idea is - instead of returning `this` from the `run` method - we could collect and return the inner promises and use these as a way to say “the gateway is ready” simply using the `then` method.

Same goes with an eventual `shutdown` method - just close all the applications and resolve the promises once.

Although we have that singleton problem right now, it can be tamed with a correct usage of `beforeEach`.

//cc @kevinswiber 

What do you guys think?

DrMegavolt:
@kevinswiber this is why config is no just a parameter for gateway and for admin 
https://github.com/ExpressGateway/express-gateway/blob/master/lib/gateway/index.js#L63

@XVincentX here is this part implemented for admin
https://github.com/ExpressGateway/express-gateway/blob/master/test/common/admin-helper.js
it is exactly as you described 

please do not join it with gateway. it is common you need only one of those 

for gateway 
check this part 
https://github.com/ExpressGateway/express-gateway/blob/master/test/common/gateway.helper.js#L24

[...]

PR #462, opened 2017-10-04
Feature/validation engine

stanislaw-glogowski:
Connected #447 

The following pull request adds json schema validation to policy, condition and plugins params.
It also creates new `GET /schemas/:type?/:name?` endpoint in admin rest api.

**Example**:
`GET /schemas/policy`

**Response**:
```json
{
  "query": {
    "type": "policy"
  },
  "schemas": [{
    "name": "proxy",
    "type": "policy",
    "schema":{
      "$id": "http://express-gateway.io/schemas/proxy.json",
      "type": "object",
      "properties": {
        "serviceEndpoint": {"$ref": "config.json#/gateway/serviceEndpoint"},
        "changeOrigin": {"type": "boolean"},
        "strategy": {"type": "string", "enum": ["round-robin"]}
      },
      "required": ["serviceEndpoint"]
    }
  }, {
    "name": "oauth2",
    "type": "policy",
    "schema": {}
  }],
  ...
}
```


lib/policies/proxy/index.js
```diff
@@ -1,3 +1,12 @@
 module.exports = {
-  policy: require('./proxy')
+  policy: require('./proxy'),
+  schema: {
+    type: 'object',
+    properties: {
+      serviceEndpoint: {type: 'string'},
```
DrMegavolt: I understand that this is work in progress, just a reminder, pls add all options from corresponding middleware

DrMegavolt:
one small note:
properties like jscode assume some code editor (maybe not now, but in future)
and log/expression have kind of template syntax in some properties

most likely it will be helpful for UI to understand that these are not just strings
cc @kolbinski

XVincentX:
I'll take over this PR.

lib/conditions/index.js
```diff
@@ -0,0 +1,50 @@
+const express = require('express');
+const logger = require('../logger').policy;
+const schemas = require('../schemas');
+const predefined = require('./predefined');
+const conditions = {};
+
+function register ({name, handler, schema}) {
+  let validate;
```
XVincentX: ToDo: Change this function so it does not throw an exception. If it's returning a boolean, it means it's caller responsibility whether to throw an exception or not.

lib/conditions/index.js
```diff
@@ -0,0 +1,50 @@
...
+  try {
+    validate = schemas.register('condition', name, schema);
+  } catch (err) {
+    logger.warn(`warning: invalid schema for ${name} condition`, err);
+    return null;
+  }
+
+  conditions[name] = (req, config) => {
+    try {
+      validate(config);
```
XVincentX: Note to self — again, do not throw an exception.

lib/conditions/predefined.js
```diff
@@ -0,0 +1,126 @@
...
+module.exports = [
+  {
+    name: 'always',
+    handler: () => true,
+    schema: {}
+  }, {
+    // Not sure if anyone would ever use this in real life, but it is a
+    // "legitimate" condition, and is useful during tests.
+    name: 'never',
+    handler: (_req) => false,
```
XVincentX: Nip: not required parameter.

lib/plugins.js
```diff
@@ -30,6 +31,10 @@ module.exports.load = function ({config}) {
       if (semver.lt(engineVersion, plugin.version)) { // TODO: versioning of plugins
         logger.warn(plugin.version, 'is higher than engine version:', engineVersion, 'trying to load');
       }
+
+      // register schema and validate settings
+      schemas.register('plugin', pluginName, plugin.schema)(settings);
```
XVincentX: Not handled exception.

lib/schemas/index.js
```diff
@@ -0,0 +1,70 @@
...
+  schemas: predefined
+});
+
+const buildKey = (type, name) => `${type}:${name}`;
+
+const registered = {};
+
+function register (type, name, schema) {
+  let compiled;
+  if (schema && typeof schema === 'object') {
```
XVincentX: Worthless type check.

lib/schemas/index.js
```diff
@@ -0,0 +1,70 @@
...
+    compiled = ajv.compile(schema);
+  } else {
+    schema = {};
+    compiled = () => true;  // empty validator
+  }
+
+  registered[buildKey(type, name)] = {
```
XVincentX: ajv has a built in caching mechanism. This can be removed.

XVincentX:
@DrMegavolt I've modified couple of things and it should now be ready (with real passing tests this time), but I need another pairs of eyes.

I've also updated #422 in order to remember we should destroy `schemas` during the shutdown phase — but this is going to be another story.

lib/policies/index.js
```diff
@@ -10,12 +12,32 @@ class Policies {
...
+
+    const validate = schemas.register('policy', name, schema);
+
+    policy.policy = (params, ...args) => {
+      const validationResult = validate(params);
+      if (validationResult.isValid) {
+        return action(params, ...args);
+      } else {
+        logger.warn(`warning: policy ${name} params validation failed`, validationResult.error);
+        return null;
```
DrMegavolt: Does this means that if I have incorrect schema I won't be able to use policy ? 
can it be  `return action(params, ...args);` here as well ?

lib/policies/key-auth/index.js
```diff
@@ -1,5 +1,6 @@
-'use strict';
-
 module.exports = {
-  policy: require('./keyauth')
+  policy: require('./keyauth'),
+  schema: {
+    // TODO: add schema
```
DrMegavolt: ##### Options Reference

* `apiKeyheader`:
  - 'Authorization'
  - the name of the header that should contain the API key
* `apiKeyHeaderScheme`:
  - 'apiKey'
  - the enforced scheme in the header
* `apiKeyHeader`:
  - 'Authorization'
  - name of the header that should contain api key
* `apiKeyHeaderScheme`:
  - 'apiKey'
  - enforce a schema in the header
* `disableHeaders`:
  - false
  - disable apikey lookup in headers
* `disableHeadersScheme`:
  - false
  - disable verification of scheme in header
* `apiKeyField`:
  - 'apiKey'
  - name of field to check in query parameter
* `disableQueryParam`:
  - false
  - set to true to disable api key lookup in query string

lib/policies/rate-limit/index.js
```diff
@@ -1,5 +1,15 @@
-'use strict';
-
 module.exports = {
-  policy: require('./rate-limit')
+  policy: require('./rate-limit'),
+  schema: {
+    type: 'object',
+    properties: {
+      rateLimitBy: {type: 'string'},
+      windowMs: {type: 'integer'},
```
DrMegavolt: btw, can it be some description here or link to the doc? I'm ok with current version, just thinking how UI can show  help info

lib/schemas/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+function find (type = null, name = null) {
+  if (type && name) {
+    const item = ajv.getSchema(buildKey(type, name));
+    if (item) {
+      return [{ type, name, schema: item.schema }];
+    }
+  } else if (type) {
+    return registeredKeys
+      .filter(item => item.type === type)
```
DrMegavolt: can we join L45 + L49 else case using condition: 
`.filter(item => !type || item.type === type)` ?

DrMegavolt:
Some comments, but in general is almost good to go

lib/schemas/index.js
XVincentX: Good idea. Addressed in 0638256ff00fb402b653f9bac0238dae84fc2ec4

lib/policies/rate-limit/index.js
XVincentX: Good point — JSON schema supports `title` and `description` property so we can definitely decorate those once we'll be able to forward this data to the UI.

lib/policies/index.js
XVincentX: It depends — the idea behind the schema validation should be preventing the user from using the plugin if the provided parameters are not correct, therefore throwing an exception when this happens isn't it? Maybe @kevinswiber knows more about that?

PR #470, opened 2017-10-07
Authenticating consumer no longer requires options to be passed, all…

srcnix:
Previously, authenticateToken() did not work when needing to validate consumer due to not passing options. 

As options are not required, don't make them required.

### DIFF
--- a/lib/services/auth.js
+++ b/lib/services/auth.js
@@ -105,8 +105,7 @@ s.authorizeCredential = function (id, authType, scopes) {
   });
 };
 
-s.validateConsumer = function (id, options) {
-  options = options || {};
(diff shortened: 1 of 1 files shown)


lib/services/auth.js
```diff
@@ -106,6 +106,7 @@ s.authorizeCredential = function (id, authType, scopes) {
 };
 
 s.validateConsumer = function (id, options) {
+  options = options || {};
```
XVincentX: I think you can probably do `function(id, options = {})`

srcnix: Good point. Need to remember this! Shall update now and push.

PR #473, opened 2017-10-12
Update dependencies

XVincentX:
This Pull Request will update all minor and patch dependencies.
No harms should happen (TM)


XVincentX:
Kaboom!

PR #485, opened 2017-10-25
Credential Service — minor refactor

XVincentX:
Connects #361 

This pull request will change several things in the credential service. As I'll be working on this soon due to #361 , it's good to have a better base.

1. Reduce promise nesting (which is always bad)
2. DRY some functions and reuse them where possible
3. Throw errors instead of rejecting promises (an exception in a promise *IS* a rejection)

This is easier to review disabling whitespace diffs: https://github.com/ExpressGateway/express-gateway/pull/485/files?w=1

### DIFF
--- a/test/services/auth.test.js
+++ b/test/services/auth.test.js
@@ -17,7 +17,7 @@ describe('Auth tests', function () {
   let _credential;
   let originalUserModelConfig;
 
-  before(function (done) {
+  before(() => {
     credentialModelConfig.oauth2 = {
       passwordKey: 'secret',
       autoGeneratePassword: true,
@@ -33,8 +33,8 @@ describe('Auth tests', function () {
       email: { isRequired: false, isMutable: true }
     };
 
-    db.flushdbAsync()
-      .then(function (didSucceed) {
+    return db.flushdbAsync()
+      .then(() => {
         user = {
           username: 'irfanbaqui',
           firstname: 'irfan',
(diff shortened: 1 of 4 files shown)


XVincentX:
@DrMegavolt Tests are passing — I'm missing permission to retry AppVeyor //cc @kevinswiber

test/services/auth.test.js
```diff
@@ -77,102 +77,102 @@ describe('Auth tests', function () {
...
-
-          // reset credential back to active status
-          credentialService
-          .activateCredential(user.username, 'oauth2')
-          .then(function (res) {
-            should.exist(res);
-            res.should.eql(true);
-            done();
-          });
+          done();
```
DrMegavolt: just remove the `done` callback as well as `catch`.

test/services/auth.test.js
```diff
@@ -77,102 +77,102 @@ describe('Auth tests', function () {
...
+              // reset credential back to active status
+              credentialService
+                .activateCredential(user.username, 'oauth2')
+                .then(function (res) {
+                  should.exist(res);
+                  res.should.eql(true);
+                  done();
+                });
+            })
+            .catch(function (err) {
```
DrMegavolt: same here, just return the promise and remove done

test/services/auth.test.js
```diff
@@ -182,97 +182,97 @@ describe('Auth tests', function () {
...
-        done();
-      })
-      .catch(function (err) {
-        should.not.exist(err);
-        done();
-      });
+          should.exist(authResponse);
+          Object.keys(authResponse.token).sort().should.eql(expectedTokenProps.sort());
+          Object.keys(authResponse.consumer).sort().should.eql(expectedConsumerProps.sort());
+          delete authResponse.token.expiresAt;
```
DrMegavolt: can we use should.haveProperties ?

test/services/auth.test.js
```diff
@@ -182,97 +182,97 @@ describe('Auth tests', function () {
...
-      .catch(function (err) {
-        should.not.exist(err);
-        done();
-      });
+        .then(authResponse => {
+          should.exist(authResponse);
+          authResponse.should.eql(false);
+          done();
+        })
+        .catch(function (err) {
```
DrMegavolt: rm done\catch

test/services/credential-service/credentials.test.js
```diff
@@ -32,16 +32,16 @@ describe('Credential service tests', function () {
...
-          console.log('Failed to flush the database');
-        }
-        done();
-      })
-      .catch(function (err) {
-        should.not.exist(err);
-        done();
-      });
+        .then(function (didSucceed) {
+          if (!didSucceed) {
```
DrMegavolt: not needed didSucceed is always true

test/services/credential-service/credentials.test.js
```diff
@@ -107,66 +107,62 @@ describe('Credential service tests', function () {
...
-        done();
-      })
-      .catch(done);
+        .insertCredential(username, 'basic-auth', _credential)
+        .then(function (newCredential) {
+          should.exist(newCredential);
+          newCredential.isActive.should.eql(true);
+          done();
+        })
+        .catch(done);
```
DrMegavolt: just return promise

test/services/credential-service/credentials.test.js
```diff
@@ -231,40 +227,36 @@ describe('Credential service tests', function () {
...
-      });
     });
 
     it('should delete a credential', function (done) {
       credentialService.insertCredential(user.username, 'oauth2')
-      .then(res => {
-        should.exist(res);
-        credentialService.removeCredential(user.username, 'oauth2')
-        .then(deleted => {
-          deleted.should.eql(true);
```
DrMegavolt: you removed this check and check for  `resAfterDelete`, why ? 
what if some code relies on it returning true\false\undefined ?

test/services/credential-service/credentials.test.js
```diff
@@ -289,16 +281,16 @@ describe('Credential service tests', function () {
       };
 
       db.flushdbAsync()
```
DrMegavolt: just return `db.flushdbAsync()` , then catch is redundant

test/services/credential-service/credentials.test.js
XVincentX: Good point, better to restore those.

XVincentX:
@DrMegavolt I've addressed all your comments and went even further. I looked over the two test files and I've reduced the promise nesting to the bare minimum, deleted all old `done` callbacks as well as using `should` where possible (to check rejection, for example).

Please have yet another look. Once this gets merged, we can move on with the user id/names PR.

PR #490, opened 2017-10-27
Credential insert by name

### LINKED ISSUE #361: [bugs] CLI: eg credentials list -c <userId>

XVincentX:
Connects #361 closes #361 
Attempt 2 after #488 failure.

- [x] Merge #485 first

Review without whitespace will make the job easier: https://github.com/ExpressGateway/express-gateway/pull/490/files?w=1

## What's up?

The following _PR|Monster|Dragoon_ will change the way credentials are stored and retrieved with regards of the consumers.

In particular:

1. On the CLI, whether you're doing `-c username` or `-c id`, now a consumer lookup will be performed, therefore all the operations will be performed on the same entity. The `id` will always be the source of truth for all the subsequent operations.

2. For basic auth, the credential id *IS* the key id. This is not limiting anything as the gateway today is supporting 1 basic-auth key per consumer. We can change this behaviour in the future eventually.

3. Update all the tests to use `user.id` instead of the username when looking up for the credentials.

# Warning

As this is fixing a wrong behavior, This might be a breaking change for existing users.

## What what?

If an user created a credential in this way:

`eg credentials create -c userName`

It won't be found anymore by the system, since usernames are now "transformed" into their IDs first, and then the credential is retrieved/stored/bla bla.

As I do not know how many people are relying on this, we need to discuss whether this is a breaking change or a simple bugfix (hopefully the latter)

### DIFF
--- a/lib/rest/routes/credentials.js
+++ b/lib/rest/routes/credentials.js
@@ -86,9 +86,8 @@ module.exports = function () {
 
         return credentialSrv
           .getCredentials(consumer.id)
-          .then(credentials => res.json({ credentials }))
-          .catch(next);
-      });
+          .then(credentials => res.json({ credentials }));
+      }).catch(next);
   });
 
   return router;
--- a/lib/rest/routes/credentials.js
+++ b/lib/rest/routes/credentials.js
@@ -81,7 +81,7 @@ module.exports = function () {
     findConsumer(req.params.consumerId)
       .then((consumer) => {
         if (!consumer) {
-          return res.status(422).json(new Error('Consumer Not Found: id:' + req.body.consumerId));
+          return res.status(404).json(new Error('Consumer Not Found: id:' + req.body.consumerId));
         }
 
         return credentialSrv
--- a/lib/rest/routes/credentials.js
+++ b/lib/rest/routes/credentials.js
@@ -13,16 +13,11 @@ module.exports = function () {
           return res.status(422).json(new Error('Consumer Not Found: id:' + req.body.consumerId));
         }
 
-        return credentialSrv.insertCredential(req.body.consumerId, req.body.type, req.body.credential)
-          .then(data => {
-            res.json(data);
-          });
+        return credentialSrv
+          .insertCredential(consumer.id, req.body.type, req.body.credential)
+          .then((data) => res.json(data));
       })
-      .catch(err => {
-        if (!res.headersSent) { // no need to send error if 422 sent
-          next(err);
-        }
-      });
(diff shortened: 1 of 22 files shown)


XVincentX:
When tests are finally passing after 15 hours of work:

image

XVincentX:
@altsang Yes. That's why I wrote that it **might** be a breaking change.

Master branch: `-c userId` and `-c username` were treated as two different entities, therefore you would get one credential for the `userid` and one for the `username` (which is incorrect — they're the same entity).

With this PR, if you call the CLI with `-c username`, the system will lookup for the stored user, retrive its ID and use that one for all the successive operations. 

Given this change, all the previous (and wrong) credentials stored with username won't be accessible anymore (because of this translation).

We could: 
- Communicate the change and advice everybody to migrate the old credentials manually
- Provide a migration script
- Provide some code-layer that would detect the legacy condition and keep going
- Anything else that's not coming in my mind right now

I'm open to other possibilities of course — I just wanted to understand what's your intention first.

XVincentX:
We agreed we'll create a `migration` command. I guess we can review/merge this PR in meantime.

lib/rest/routes/credentials.js
```diff
@@ -83,10 +78,16 @@ module.exports = function () {
...
 
   router.get('/:consumerId', function (req, res, next) {
-    credentialSrv
-      .getCredentials(req.params.consumerId)
-      .then(credentials => res.json({ credentials }))
-      .catch(next);
+    findConsumer(req.params.consumerId)
+      .then((consumer) => {
+        if (!consumer) {
+          return res.status(422).json(new Error('Consumer Not Found: id:' + req.body.consumerId));
```
DrMegavolt: 422 is kind of form validation. it should be 404, since no consumer exists

XVincentX: That's absolutely correct. 422 means "not processable" entity.

XVincentX:
I'll candidly admit I have no idea why the coverage is decreasing and how to put it back and make it pass.

PR #491, opened 2017-10-27
Configurable proxy options

srcnix:
Added:
- Ability to specify proxy options for proxy policy actions
- Ability to specify proxy options for serviceEndpoint
- Options provided in proxy policy take precedence over serviceEndpoint
- Can specify target certFile and keyFile to have certs loaded on proxy
- Proxy options also passed to strategies

No tests have been written, but I intend to write them shortly.

It would be good to add some documentation also. Where do you recommend I add this information?


DrMegavolt:
connect #335

lib/policies/proxy/proxy.js
```diff
@@ -51,15 +54,49 @@ module.exports = function (params, config) {
...
+  let parsedProxyOptions = {};
+
+  if (Object.keys(endpoint).includes('proxyOptions')) {
+    parsedProxyOptions = Object.assign(endpoint.proxyOptions, proxyOptions);
+  }
+
+  if (!Object.keys(parsedProxyOptions).includes('target')) parsedProxyOptions.target = {};
+
+  if (parsedProxyOptions.target.keyFile) {
+    parsedProxyOptions.target.key = fs.readFileSync(parsedProxyOptions.target.keyFile);
```
kevinswiber: @srcnix Why the difference in option names (e.g., `keyFile` vs `key`)?

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
+const path = require('path');
+const fs = require('fs');
+const request = require('supertest');
+const assert = require('chai').assert;
+const https = require('https');
+const express = require('express');
+const expressApp = express();
```
XVincentX: Nip: this can be created in the tests setup rather than at the top of the file.

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
...
+      });
+
+      it('passes options to proxy', (done) => {
+        request(app).get('/endpoint')
+          .expect(200)
+          .expect('Content-Type', /json/)
+          .end(function (error, res) {
+            if (error) return done(error);
+
+            assert.ok(res.body.port === port);
```
XVincentX: Nip: You can use `assert.strictEqual(res.body.port, port)`

lib/policies/proxy/round-robin.js
```diff
@@ -1,21 +1,24 @@
...
   proxy (req, res) {
-    const options = {
-      target: this.nextTarget()
-    };
+    const nextTarget = this.nextTarget();
+    const parsedEndpointUrl = url.parse(nextTarget);
 
-    logger.debug(`proxying to ${options.target}, ${req.method} ${req.url}`);
-    return this._proxy.web(req, res, options);
+    this.proxyOptions.target = Object.assign(this.proxyOptions.target, parsedEndpointUrl);
```
XVincentX: Why not simply use the equal operator?
`this.proxyOptions.target = parsedEndpointUrl`

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
...
+const clientKeyFile = path.join(__dirname, '../../fixtures/certs/client', 'client.key');
+const clientCertFile = path.join(__dirname, '../../fixtures/certs/client', 'client.crt');
+const chainFile = path.join(__dirname, '../../fixtures/certs/chain', 'chain.pem');
+
+describe('proxy policy', () => {
+  let app, service;
+
+  before('start service', () => {
+    expressApp.all('*', function (req, res) {
+      res.status(200).json({ port: port }).end();
```
XVincentX: Nip: `.end` is not required

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
...
+    });
+
+    service = https.createServer({
```
XVincentX: `server`, `httpServer` are probably better names

test/policies/proxy/proxy.test.js
XVincentX: I can see this piece of code is replicated multiple times. Could you define it somewhere above and reuse where possible?

lib/policies/proxy/proxy.js
```diff
@@ -51,15 +54,49 @@ module.exports = function (params, config) {
...
     return balancer.proxy(req, res);
   };
 };
+
+// Parse endpoint URL if single URL is provided.
+// Extend proxy options by allowing and parsing keyFile and certFile.
+function parseProxyOptions (proxyOptions = {}, endpoint = {}) {
+  let parsedProxyOptions = {};
+
+  if (Object.keys(endpoint).includes('proxyOptions')) {
```
XVincentX: We don't need to enumerate the whole `endpoint` keys. We can just check for property existence:

`if (endpoint.proxyOptions) {…}`

lib/policies/proxy/proxy.js
```diff
@@ -51,15 +54,49 @@ module.exports = function (params, config) {
...
+    parsedProxyOptions = Object.assign(endpoint.proxyOptions, proxyOptions);
+  }
+
+  if (!Object.keys(parsedProxyOptions).includes('target')) parsedProxyOptions.target = {};
```
XVincentX: dtto, key enumeration is not required.

lib/policies/proxy/proxy.js
XVincentX: @kevinswiber `key` is the option that the proxy module wants — `keyFile` is a custom property we use to load the certificate from file before feeding the module. Does that make sense?

srcnix: @kevinswiber, exactly what @XVincentX said. Proxy accepts cert, key and ca as a certificate string. 

In my case, and I'm certain many others, the file location is a more security means of providing the certificates. An example case would be where certificates change value, such as when a PKI with short term certificates are involved.

lib/policies/proxy/proxy.js
srcnix: I didn't realise this enumerates all keys. Assumed (bad!) it works similarly to .includes? in ruby. Fixing.

lib/policies/proxy/proxy.js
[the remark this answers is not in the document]
srcnix: The point of checking those properties is know whether we need to to perform a read from the corresponding file path. If we don't, we are always assuming keyFile will be set.

lib/policies/proxy/proxy.js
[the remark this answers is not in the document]
srcnix: That was my thought also, being the init phase we can wait. However a refactor is suitable also. Do you allow async/await?

lib/policies/proxy/proxy.js
```diff
@@ -51,15 +54,49 @@ module.exports = function (params, config) {
...
+    parsedProxyOptions.target.ca = fs.readFileSync(parsedProxyOptions.target.caFile);
+    delete parsedProxyOptions.target.caFile;
+  }
+
+  if (endpoint.url) {
+    const parsedEndpointUrl = endpoint.url ? url.parse(endpoint.url) : {};
+    parsedProxyOptions.target = Object.assign(parsedProxyOptions.target, parsedEndpointUrl);
+  }
+
+  return parsedProxyOptions;
```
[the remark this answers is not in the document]
srcnix: Using ``` delete ``` is not required, but is keeping the object being manipulated clean. Hence doing it.

What would you prefer?

lib/policies/proxy/round-robin.js
srcnix: The key, cert and ca certificate properties exist in the passed proxyOptions, which is mapped to this.proxyOptions. Your suggestion would remove the SSL configuration passed in.

srcnix:
@XVincentX, RE the tests:

To start, I have created a HTTPS server which will deny all connections that do not have a certificate, or ones that have an invalid certificate passed during the handshake.

The setupGateway function accepts two params. The first is for setting proxy options on the service, the second is for setting the options on the proxy policy action.

The first test checks the response code is 200, which wouldn't be the case if no certificate was configured. A 500 is thrown if the certificates are not sent. 

The first test sets the proxy options on the serviceEndpoint and has nothing set for the policy action. This tests that proxy options set on the serviceEndpoint are used.

The second test performs the same checks as the first test, except the proxy options are set on the policy action.  This tests that proxy options set on the proxy action are used.

The third test also performs the same checks as the first test. Proxy options are set on both the endpointService and policy action, but are different. This checks that if both the serviceEndpoint and the proxy action have the proxy options set, the policy proxy options take precedence.

lib/policies/proxy/proxy.js
kevinswiber: :+1:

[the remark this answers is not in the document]
DrMegavolt: @XVincentX let's keep sync version.  it is indeed initialization\hot-reload phase.  

@srcnix async\await will break in node-6, I think that is the only feature except ES6 modules that we cannot use for now

lib/policies/proxy/proxy.js
```diff
@@ -51,15 +54,49 @@ module.exports = function (params, config) {
...
+
+  if (!Object.keys(parsedProxyOptions).includes('target')) parsedProxyOptions.target = {};
+
+  if (parsedProxyOptions.target.keyFile) {
+    parsedProxyOptions.target.key = fs.readFileSync(parsedProxyOptions.target.keyFile);
+    delete parsedProxyOptions.target.keyFile;
+  }
+
+  if (parsedProxyOptions.target.certFile) {
+    parsedProxyOptions.target.cert = fs.readFileSync(parsedProxyOptions.target.certFile);
```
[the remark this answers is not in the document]
DrMegavolt: @XVincentX technically, if it throws exception then config is not valid and should not be loaded. And EG  should log the error if it fails to load, so no additional try\catch is needed unless exceptions is really not obvious and we need some human readable message here

lib/policies/proxy/proxy.js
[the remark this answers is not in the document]
DrMegavolt: @srcnix @XVincentX 
xxxxFile is just some string. it is not a lot of data. why we cannot just remove the delete part and allow passing everything? 

@XVincentX  creating new object will require either listing all properties or using some `_.omit` it is kind of performance impact and code readabliity  issue

lib/policies/proxy/proxy.js
XVincentX: I would assume that `.includes` does the same thing as Javascript.

My point here is that instead of enumerating all the object keys and then check a property presence from an array of strings, we can check the property existence directly.

```diff
- if (Object.keys(endpoint).includes('proxyOptions')) {
+ if (endpoint.proxyOptions) {
```

https://runkit.com/59ef79493b34ed00123dd9b0/59faf9e0a81fc60012e4f649

lib/policies/proxy/proxy.js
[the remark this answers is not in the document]
XVincentX: Yes absolutely — we can't go async now. It was just a shout on the empty space 😄

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
...
+      key: fs.readFileSync(serverKeyFile),
+      cert: fs.readFileSync(serverCertFile),
+      ca: fs.readFileSync(chainFile),
+      requestCert: true,
+      rejectUnauthorized: true
+    }, expressApp);
+
+    service.listen(port);
```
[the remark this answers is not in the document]
DrMegavolt: and having specific port can (and will) provide issues 
I'm ok leaving this as is for now, just a comment that we should refactor it later to use for example  'startGatewayInstance' from `gateway.helper`

lib/policies/proxy/proxy.js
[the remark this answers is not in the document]
XVincentX: Ah ok — if we have a global error handler somewhere, that's good.
I just wanted to make sure the failed-to-load filename was logged somewhere


```javascript
fs.readFileSync('IDoNotExist.txt');
```

throws:

```
Error: ENOENT: no such file or directory, open 'asd.txt'
```

So we have enough informations to show to the user. Ok!

lib/policies/proxy/proxy.js
srcnix: I get your point, and have made the changes. Just not pushed yet. Waiting on responses to my other comments. 

Also, I meant .key? not includes?! key? does not iterate.

test/policies/proxy/proxy.test.js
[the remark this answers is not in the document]
srcnix: I'll refactor the code to use the gateway helper, it's best we stick to a more DRY principle and you're right, hard coding port is not a good idea - however, I was just following the existing tests already merged into master ;)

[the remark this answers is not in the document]
XVincentX: For now I'm ok leaving it as well — but If you'd like to refactor it we'll be more than happy.
The important part is to handle the async flows correctly or these tests will be flaky

lib/policies/proxy/proxy.js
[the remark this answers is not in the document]
XVincentX: I agree with that — I probably commented on the wrong line.

What I'm trying to say is that if this condition is true:

` if (!Object.keys(parsedProxyOptions).includes('target')) parsedProxyOptions.target = {};`

Then `target` is an empty object for sure. Therefore it does not make sense to make the checks you're doing in the following line such as:

`if (parsedProxyOptions.target.keyFile) {…}`
because target is a empty object for sure.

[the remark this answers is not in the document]
srcnix: Ahah! Yes. Of course! I'll amend.

lib/policies/proxy/proxy.js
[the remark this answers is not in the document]
XVincentX: The thing here is that, in the current state, `parseProxyOptions` function is not pure, because it's modifying the input argument `endpoints`.

This side effect makes the function impossible to be called twice with the same arguments (because properties are being removed) from the input object.

The problematic line is `parsedProxyOptions = Object.assign(endpoint.proxyOptions, proxyOptions);`
(as the returned reference is on `endpoint.proxyOptions`)

Either we create a new, parsed object and return it, or let's do something to make sure we do not modify user provided arguments.

XVincentX:
@srcnix Ok I got the tests now, thanks a lot for the clarification. I'd probably add at least a scenario that's not on the happy path.

srcnix:
@XVincentX, agreed. I'll add tests for failure scenarios.

lib/policies/proxy/proxy.js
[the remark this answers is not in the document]
srcnix: Gotcha! Shall update.

test/policies/proxy/proxy.test.js
srcnix: Updating.

test/policies/proxy/proxy.test.js
srcnix: Yup, just being explicit. Shall modify.

test/policies/proxy/proxy.test.js
srcnix: Shall change to server.

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
...
+  });
+
+  after('clean up', () => {
+    config.gatewayConfig = originalGatewayConfig;
+    service.close();
+  });
+
+  describe('proxyOptions', () => {
+    describe('when proxy options are specified on the serviceEndpoint', () => {
+      before((done) => {
```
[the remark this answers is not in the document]
srcnix: Perfect! I didn't realise you could do that.

test/policies/proxy/proxy.test.js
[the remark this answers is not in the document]
srcnix: Shall update.

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
...
+      requestCert: true,
+      rejectUnauthorized: true
+    }, expressApp);
+
+    service.listen(port);
+  });
+
+  after('clean up', () => {
+    config.gatewayConfig = originalGatewayConfig;
+    service.close();
```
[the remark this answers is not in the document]
srcnix: Shall update.

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
...
+            caFile: chainFile
+          }
+        }).then(apps => {
+          app = apps.app;
+          done();
+        });
+      });
+
+      after(() => {
+        app.close();
```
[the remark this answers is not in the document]
srcnix: Shall update.

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
...
+              caFile: chainFile
+            }
+          }).then(apps => {
+            app = apps.app;
+            done();
+          });
+        });
+
+        after(() => {
+          app.close();
```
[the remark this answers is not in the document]
srcnix: Shall update.

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,174 @@
...
+            assert.ok(res.body.port === port);
+            done();
+          });
+      });
+    });
+
+    describe('When proxy options are specified on the policy action', () => {
+      describe('and no proxy options are specified on the serviceEndpoint', () => {
+        before((done) => {
+          setupGateway({}, {
```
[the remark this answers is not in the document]
srcnix: Shall update.

test/policies/proxy/proxy.test.js
srcnix: Going to remove this check, as it's not at all required. The status and content-type is more than enough to validate the test.

srcnix:
@XVincentX and @DrMegavolt, I have refactored code where possible (where suitable) and added additional tests.

DrMegavolt:
Awesome, thanks for using dynamic port.

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,173 @@
...
+  after('clean up', (done) => {
+    config.gatewayConfig = originalGatewayConfig;
+    backendServer.close(done);
+  });
+
+  describe('proxyOptions', () => {
+    it('raises an error when incorrect TLS file paths are provided', (done) => {
+      const serviceOptions = { target: { keyFile: '/non/existent/file.key' } };
+
+      setupGateway(serviceOptions).catch(err => {
```
XVincentX: Nip: this can be replaced with a single line: `should(setupGateway(serviceOptions)).be.rejectedWith('no such file or directory')`

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,173 @@
...
+    describe('when incorrect proxy options are provided', () => {
+      before(() => {
+        const serviceOptions = { target: { certFile: invalidClientCertFile } };
+
+        return setupGateway(serviceOptions).then(apps => {
+          app = apps.app;
+        });
+      });
+
+      after((done) => {
```
XVincentX: Nit: `after(app.close)`

test/policies/proxy/proxy.test.js
```diff
@@ -0,0 +1,173 @@
...
+        it('uses both configurations, with policy proxy options taking precedence', () => {
+          return expectedResponse(app, 200, /json/);
+        });
+      });
+    });
+  });
+});
+
+function setupGateway (serviceOptions = {}, policyOptions = {}) {
+  return findOpenPortNumbers(1).then((ports) => {
```
XVincentX: Nit: `return findOpenPortNumbers(1).then(([port]) => {……`

PR #494, opened 2017-10-31
Basic Migrations command

### LINKED ISSUE #492: Baisc migrations support
Bad things happens. Good as well. When some of those happen, we need to migrate data so our users can keep using the Gateway without having to destroy the data they accumulated so far.

We'll offer a CLI command to create/store/run migrations. [node-migrate](https://github.com/tj/node-migrate/tree/1.x) might bee a good choice.

We **won't** be offering an API for that. As these might be harmful, I think it's in the best interest to make these things offline.

XVincentX:
This pull request will introduce a migration mechanism for the gateway leveraging [node-migrate@next](https://npmjs.org/packages/migrate) module by Tj.

Connects #490
Connects #492 closes #492

Standing points:

- This module can track the ran migrations using a `.migrate` file or using a custom storage mechanism. Question is whether we want to use the file or store the things in Redis, for example.
If we keep them on Redis, you move the DB, you are ok. If you keep them on the file system you might re-run the migrations if you change the server. ✌️ 

- How do we make sure the users will run the migrations? I've checked Kong and they do not care. They rely on the fact that you read the change log and notice you've got a migration to run; therefore you run them. I'd stick with the same mechanism ✌️ 

- [x] Figure out a way to test the effects by this migration
~Integrate the migration command with the `eg` cmd line~ (we're not doing that)

### DIFF
--- a/test/migrations/credential-username-userid-transform.js
+++ b/test/migrations/credential-username-userid-transform.js
@@ -1,0 +1,55 @@
+const migrate = require('migrate');
+const tmp = require('tmp');
+const db = require('../../lib/db')();
+const { assert } = require('chai');
+const userService = require('../../lib/services/consumers/user.service');
+const credentialService = require('../../lib/services/credentials/credential.service');
+
+describe('Migrations', () => {
+  describe('Username -> UserID credential transform', () => {
+    let tmpFile;
+    let userId;
+    let oldCredential;
+    before('I insert some credentials by user ID', () => {
+      return userService.insert({
+        username: 'vncz',
+        firstname: 'Vincenzo',
+        lastname: 'Chianese'
+      }).then((user) => {
+        userId = user.id;
+        return credentialService
+          .insertCredential('vncz', 'basic-auth', {})
+          .then(() => credentialService.getCredentials('vncz'))
+          .then((creds) => { oldCredential = creds[0]; });
+      });
+    });
+
+    before('I then run the migration script', (done) => {
+      tmpFile = tmp.fileSync();
+      migrate.load({ stateStore: tmpFile.name }, (err, set) => {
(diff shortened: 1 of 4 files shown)


XVincentX:
// @kevinswiber @DrMegavolt What's your idea around the standing points?

DrMegavolt:
what is wrong with just running `migrate` ? why do we need `eg` cli integration ?

XVincentX:
I thought it was a requirement in order to give to the user an uniform interface to use the gateway.
Anyway, `npx migrate` or execute the bin file directly will do the job perfectly and it'll also save me some work. I'm fine removing this 👍

XVincentX:
I removed the CLI integration and we'll stick with the default storing mechanism! 👍

XVincentX:
@DrMegavolt Please review this one ore time. I've modified the code to use the `dao` instead of the `service` — this is to skip the password hashing and have a simple password copy. In this way our users won't have to communicate any new password.

Also, I've added a test that should check the whole flow.

test/migrations/credential-username-userid-transform.js
```diff
@@ -0,0 +1,55 @@
...
+const { assert } = require('chai');
+const userService = require('../../lib/services/consumers/user.service');
+const credentialService = require('../../lib/services/credentials/credential.service');
+
+describe('Migrations', () => {
+  describe('Username -> UserID credential transform', () => {
+    let tmpFile;
+    let userId;
+    let oldCredential;
+    before('I insert some credentials by user ID', () => {
```
DrMegavolt: I guess the name should be 
`I insert some credentials by username`

XVincentX: Yeah!

PR #496, opened 2017-11-01
Automate docker and npm publish part

### LINKED ISSUE #495: Automate release based on tags
Express Gateway is still released manually. We can automate the process using CircleCI triggers based on tags.

It should:

- Publish on Npm registry the new version
- Build and publish new docker image as well update the `latest` tag

XVincentX:
This PR should hopefully allevieate the amount of work we have to do when we want to
release a new express-gateway version. 

It will listen on the master branch and if a new tag with vx.y.z pattern comes in, it will build and publish both npm package and docker images related, and update the 'latest' tag as well.

We need to set both docker and npm credentials through env variables (//cc @kevinswiber )

Unfortunately I can't test this properly without deploying something — so once merged we could give it a try on eg 1.4 or any bug fix that'll come along us.

Connects #495 closes #495

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,9 +20,36 @@ jobs:
       - run: npm install
       - run: npm test
 
+  deploy:
+    docker:
+      - image: circleci/node:8
+    working_directory: ~/repo
+
+    steps:
+      - setup_remote_docker
+      - run: npm publish
+      - run: "docker build -t expressgateway/express-gateway:${CIRCLE_TAG} --build-arg VERSION=latest - < Dockerfile"
+      - run: |
+              "docker build -t expressgateway/express-gateway:${CIRCLE_TAG}-getting-started
+               --build-arg VERSION=latest --build-arg TYPE=getting-started - < Dockerfile"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -20,9 +20,36 @@ jobs:
       - run: npm install
       - run: npm test
 
+  deploy:
+    docker:
+      - image: circleci/node:8
```
DrMegavolt: does it really has to be circleci node ?

XVincentX: Not really.
Technically it could be any docker image that has got `npm`; however I decided to reuse the same because

1. As it has been already downloaded and built, it should start almost immediately.
2. The `deploy` job has a transitive dependency on node8 build — therefore it's kind of a natural dependency.

If you have another idea around that feel free to shout them!

DrMegavolt: not only npm, but also tools to build `bcrypt` lib. As I remember `node:alpine` will work for that

and alpine itself is extremely small,  so push\pull time are reduced significantly   
if it works I would actually use it

Issue #499, opened 2017-11-02
Support for Redis-Sentinel


XVincentX:
Express Gateway can only work with a single Redis instance at the moment. 

*Sentinel* provides high scalability for Redis and it's a mandatory step to go in production when you're storing important data (such as keys and so on and so forth).

We need to implement support for it.

### Battle plan

- Evaluate a new Redis client. The current one does not support Sentinel natively and, although there's a library that [can monkey patch the methods](https://github.com/ortoo/node-redis-sentinel). It's not maintained as well as its usage is discouraged. We should use [ioredis](https://github.com/luin/ioredis) as it is the official solution.

- We're using FakeRedis when the `emulate` flag is true (for testing purposes as well in memory instances). Is this going to work even when the client will be switched? [There's an alternative](https://github.com/nileshmali/ioredis-in-memory)

XVincentX:
## ioredis

- [It's throwing exception when there's no data](https://github.com/stipsan/ioredis-mock/issues/328)
- [It's missing implementation for `scan` command](https://github.com/stipsan/ioredis-mock/blob/master/compat.md), which we're using.

DrMegavolt:
ioredis seems to be the way to go. it is recommended client on the redis page

we should not remove in memory storage. it is crucial for getting started experience. 
and we should not run unit tests against real redis

bugs seems to be non-critical. should be possible to fix

DrMegavolt:
and `*Async` is just a hack to modernize old school library, obviously all `xxxAsync` methods can be renamed to normal looking methods

XVincentX:
This just happened. #537 for followups

PR #501, opened 2017-11-03
Support for Redis Sentinel

XVincentX:
Connects #499 

### Mock library issues

- [x]  [The mock it's throwing exception when there's no data](https://github.com/stipsan/ioredis-mock/issues/328)
- [x] [The mock is missing implementation for `scan` command](https://github.com/stipsan/ioredis-mock/blob/master/compat.md), which we're using.
- [x] [The mock is missing support for reply transformers](https://github.com/stipsan/ioredis-mock#roadmap)

#### Solution

- Get rid of the `emulate` option and use Redis for everything, even for testing on CircleCI (we should do that anyway) https://github.com/ExpressGateway/express-gateway/pull/501/commits/4deacf977589ce74a95eba8e83813b0c18825eef

### Compatibility problems

- [x] All methods already return a Promise, there's no need for `promisify/Async` version https://github.com/ExpressGateway/express-gateway/pull/501/commits/5efe86733a1ce201ba8cec8a9785e1006f2cb0e6

- [x] `hgetall` reply must be modified to guarantee compatibility https://github.com/ExpressGateway/express-gateway/pull/501/commits/6c12cfdfc0324671a5f3752fedf71d67c0e7bfb0

- [x] `exec` function is using a Pipelining feature and has a different response type [fix](https://github.com/ExpressGateway/express-gateway/pull/501/commits/a58bca482f2edc8737b1d57c8d58f06e9b8f1eef#diff-1fe7eeca20bc9f4125db3c2909e5de55R186)

- [x] `bluebird` is overriding the default Promise implementation and this is preventing the `migrate` framework to work detect a Promise and react accordingly [issue](https://github.com/tj/node-migrate/issues/84) https://github.com/ExpressGateway/express-gateway/pull/501/commits/4ad3cfc4bdc23b68fca70fb45f3121b93249babf


XVincentX:
## Interesting followups:

1. Take advantage of [transparent key prefixing](https://github.com/luin/ioredis#transparent-key-prefixing) so we can remove some boilerplate code
2. Scan commands can return a stream — it is a great idea in order to avoid killing the server (both redis and the gateway)
3. [Cluster support](https://github.com/luin/ioredis#cluster). This is a different connection object with a different set of connection options as well.

test/db.test.js
```diff
@@ -1,79 +0,0 @@
-const path = require('path');
```
DrMegavolt: this file was added to test ssl connection just recently why it is removed ?

DrMegavolt:
If the emulate option is working I'm thinking that we can land this massive PR, and break in memory a bit. so it will be another bug to fix it

test/db.test.js
XVincentX: Good question.
I deleted it temporarily because I'll have to figure out how to test that part.

The problem here is — `ioredis` gets created using a constructor and not using a function such as `Redis.createClient()`

`sinon`, the stub library we're using, does not support spies on `constructors`.

I deleted this test just to verify it was working correctly. I'll now figure out another way to test this part.

XVincentX:
The emulate option is working correctly, yes.

I'm not really sure we should land this thing into master as it is.

I'm imagining somebody that wants to contribute or even simply trying the source code: they would clone the repository, install and run it: they would have a broke experience by default (as in-memory is not working and they would need a Redis instance), which is not very pleasant. I'd rather keep this branch here and resurrect it once `ioredis-mock` will be ready.

test/db.test.js
DrMegavolt: 👍

agonbina:
Just for the record, I am using(developing against) this branch to run EG and I haven't run into any issues when running with a single redis host or the sentinel mode.

DrMegavolt:
@agonbina thanks for active testing. the only reason we cannot merge it to master is some issues with InMemory mode. Which is implemented as mock of redis client. once fixed it will land into master

XVincentX:
I've updated the `ioredis-mock` package and rebased with latest master.

Two points still have to be solved — see [here](https://github.com/ExpressGateway/express-gateway/pull/501#issue-270940027)

XVincentX:
Rebase is going to be fun.

DrMegavolt:
squash manually, it will be way easier.

DrMegavolt:
LGTM

agonbina:
Nice to see this was merged! I will switch to using the master branch now 👍 I'm curious as to when is the next release planned?

XVincentX:
Hard work, but we made it!

These two images will summarize last couple of hours here:

image
firecomputer

DrMegavolt:
@agonbina waiting for multi-auth merge. 
Expected Thursday Nov 16. If something goes wrong then next week

PR #504, opened 2017-11-05
Fix/remove event listeners on server close

srcnix:
Remove event listeners from eventBus on server close - much tidier and allows reuse of gateway in tests without the need for forking (Theory, not practiced).

Note: Based on "Redis TLS from files" PR and will rebase to master once ready.

### DIFF
--- a/lib/gateway/server.js
+++ b/lib/gateway/server.js
@@ -6,10 +6,15 @@ const fs = require('fs');
 const path = require('path');
 const logger = require('../logger').gateway;
 const config = require('../config');
+const eventBus = require('../eventBus');
 
 module.exports.bootstrap = function (app) {
   const httpServer = config.gatewayConfig.http ? http.createServer(app) : null;
   const httpsServer = config.gatewayConfig.https && config.gatewayConfig.https.tls ? createTlsServer(config.gatewayConfig.https, app) : null;
+
+  addOnCloseEventHandlingToServer(httpServer);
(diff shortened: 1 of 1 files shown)


DrMegavolt:
There are tests for example E2E, that are using fork approach
All that magic is inside 
https://github.com/ExpressGateway/express-gateway/blob/master/test/common/gateway.helper.js
This is to ensure that cli, npm installation works correctly etc. 

and some tests use (the majority)
https://github.com/ExpressGateway/express-gateway/blob/c0535fea428826e37599378382d7d4a1fa0aeb8a/test/common/routing.helper.js#L36

and there are a lot of tests using later approach, and they don't fail 
Example: https://github.com/ExpressGateway/express-gateway/blob/869c6448c347fed64d8574952501160bb318fd76/test/policies/rate-limiter/rate-limit.by.ip.test.js

In my understanding since EventEmiter is just a js object, all of it listeners will be removed automatically on process shutdown. so why managing them manually ?

srcnix:
@DrMegavolt, the point is, you have to fork the process in order to remove all event listeners. This shouldn't be a requirement - it' nice to tidy up after ourselves.

The tests don't fail, but if you don't fork the gateway instance into its own process, after 11 tests you will start to get warnings about exceeding event listeners - removing them on close of the server ensures this don't happen.

srcnix:
Rebased!

PR #506, opened 2017-11-06
Remove branch requirement for deploy script

XVincentX:
@kevinswiber I might need `express-gateway` user data in order to make this happen.

Connects #495

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,6 +50,8 @@ workflows:
           filters:
             tags:
               only: /^v\d\.\d\.\d/
+            branches:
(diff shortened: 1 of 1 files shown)


DrMegavolt:
if the intention is to fix deploy on non tag, why it is trying to deploy ?

XVincentX:
I haven't figured it out yet. I guess I didn't read the documentation precisely enough.

XVincentX:
@DrMegavolt Fixed. I read the documentation more _carefully_ and I figured out that all branches must be excluded, and only tags included using the regexp.

PR #512, opened 2017-11-07
Use CLI with App name

### LINKED ISSUE #260: CLI: add ability to do operations against Apps with friendly names instead of AppID GUID

### LINKED ISSUE #515: Unify scripts pattern
Our scripts in package.json are a bit inconsistent.
Sometimes we've got `command:subcommand`, some other times `command-subcommand`.

Let's keep one convention (I would suggest `command:subcommand`) and let's stick to that.

XVincentX:
Connects #260 
closes #260
closes #515

Review without whitespaces here: https://github.com/ExpressGateway/express-gateway/pull/512/files?w=1 (I really should solve this problem forever and ever.)

This pull request will update the admin API in order to use all the endpoints using applications name instead of applications id. This was done by providing a [findApplication](https://github.com/ExpressGateway/express-gateway/pull/512/files?w=1#diff-e88ea0d93b455fd6f03ede8ede786698R87) that's looking for applications by name, and then by id.

The index is kept by saving another [key in redis](https://github.com/ExpressGateway/express-gateway/pull/512/files?w=1#diff-a84828d7e5fb2b16bc57843ffd196ee3R26) that's then used when looking the application by name.

There's now a check as well to make sure the tuple [userid appname is unique](https://github.com/ExpressGateway/express-gateway/pull/512/files?w=1#diff-a84828d7e5fb2b16bc57843ffd196ee3R33) and an error message if [more than one app is coming out as result from a search operation](https://github.com/ExpressGateway/express-gateway/pull/512/files?w=1#diff-a84828d7e5fb2b16bc57843ffd196ee3R80).

As always, I've added some tests to make sure the behaviors are respected as well cleaning up other tests I found on the way (in particular I got rid of all `no console.log` warnings)

- [ ] Add tests


lib/services/consumers/application.dao.js
```diff
@@ -51,59 +70,73 @@ dao.findAll = function (query) {
...
-    return Promise.all(deactivateAppPromises);
-  });
+    .then(appIds => {
+      const deactivateAppPromises = appIds.map(appId => this.deactivate(appId));
+      return Promise.all(deactivateAppPromises);
+    });
 };
 
-dao.remove = function (id, userId) {
+dao.remove = function ({ name, id }, userId) {
```
DrMegavolt: I saw in the app service you are passing entire `app` here
so `{name, id, userId}` should work 

as result in dao, `remove` will accept single argument - app

test/rest-api/applications.js
```diff
@@ -0,0 +1,54 @@
...
+      firstname: 'La',
+      lastname: 'Deeda'
+    }).then(() => {
+      return adminHelper.admin.apps.create(username, {
+        name: 'appy1',
+        redirectUri: 'http://localhost:3000/cb'
+      });
+    }));
+
+    it('should return an error on the second attemp', () => {
```
DrMegavolt: 👍

lib/services/consumers/application.dao.js
XVincentX: That's true. I'll see if I can rearrange this part.

PR #520, opened 2017-11-07
fill req.params inside policy by using mergeParams option. fixes #519

### LINKED ISSUE #519: req.params inside policy are not filled with API Endpoint path parts
```
apiEndpoints:
  api:
    host: localhost
    paths: '/:id'
```
Expected req.params.id to be filled for requests like `/45`
It doesn't happened

Notes: 
So this is the part where pipeline is linked with API Endpoint https://github.com/ExpressGateway/express-gateway/blob/master/lib/gateway/pipelines.js#L100
and this is how a pipeline is created from configuration
https://github.com/ExpressGateway/express-gateway/blob/master/lib/gateway/pipelines.js#L142
so pipeline is yet another router that is used as a handler
It looks like fix will be just add mergeParams options into line 142

dotBits:


dotBits:
@DrMegavolt I would also write a test to cover req.params matching with path. Is path-general.test.js suitable for this?

DrMegavolt:
path-general.test.js  will work 
Please extend this helper, so it will be capable of testing params 
https://github.com/ExpressGateway/express-gateway/blob/master/test/common/routing.helper.js#L98
something like
```
if (testCase.test.params){
 assert.equal(req.para,s, testCase.test.params);
```

so it will be possible to test params in any test

PR #521, opened 2017-11-08
Run tests on a real Redis Instance as well

XVincentX:
Review with: https://github.com/ExpressGateway/express-gateway/pull/521/files?w=1

Connects #499 

The following PR/dragoon/monster will enable CircleCI to execute tests on a real redis instance as well, making sure both memory and real database instance will offer the same experience.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - checkout
       - run: npm install
-      - run: npm test
+      - run: npm run test-all
 
   "node-8":
     docker:
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
     "exclude": [
       "logo",
       "node_modules",
-      "coverage"
+      "coverage",
+      "test"
     ]
   },
   "types": "./index.d.ts",


lib/db.js
```diff
@@ -14,7 +14,11 @@ module.exports = function () {
 
   // special mode, will emulate all redis commands.
   // designed for demo and test scenarious to avoid having real Redis instance
-  const emulate = process.argv[2] === 'emulate' || redisOptions.emulate;
+  let emulate = process.argv[2] === 'emulate' || redisOptions.emulate;
+
+  if (process.env.EG_DB_EMULATE) {
+    emulate = !!parseInt(process.env.EG_DB_EMULATE);
```
DrMegavolt: I like `process.env.EG_DB_EMULATE`
can we make this resistant to string or  !!parseInt(true) !!parseInt("true") !!parseInt("emulate") is false

XVincentX: Hopefully we will be able to remove this hack when configuration files will support env variables. When this will happen, that value will be injected directly from the CI environment.

XVincentX: @DrMegavolt I would not spend time on this as this is not something we will let people use. I would neither document it.

It's just a temporary thing to make this PR happen. Once config files will support env variable, this piece is going to vanish.

XVincentX:
image

I'll investigate here, although I basically didn't change any code. Sigh!

XVincentX:
@DrMegavolt It just landed in #501

PR #522, opened 2017-11-08
Publish only the required files on npm

XVincentX:
This PR will modify the package.json files section — these are the only one taken in
account when npm is publishing the package. By skipping these useless files, we'll be
able to reduce the footprint size and make our eslint/node plugin (even though is not
enabled yet) happy again.

And — add myself as contributor 😈


XVincentX:
Sure thing — I am not opinionated on that at all. I'll create a `.npmignore` file.

Just a note: `LICENSE, README, package.json` always gets published, no matter if you include them or not (that's why I didn't insert them)

PR #523, opened 2017-11-09
allow keyId and KeySecret to be set when creating key-Auth connect #517 

### LINKED ISSUE #517: Key-Auth always generate keyId\KeySecret
https://github.com/ExpressGateway/express-gateway/blob/master/lib/services/credentials/credential.service.js#L67

Key-auth credential do not respect provided keyId keySecret 
it will always autogenerate 

Useful for migration \ testing \data transfer scenarios 
Originally by @agonbina

DrMegavolt:
connect #517 
closes #517


XVincentX:
closes #517

Issue #493, opened 2017-10-30
Standardize formatting rules


XVincentX:
We do have multiple code style in the current codebase — and different editor are treating it in a different way with integrated autoformatters.

Therefore, each pull request contains tons of autoformatting changes that make the code review a pain.

We should give prettier a try and eventually integrate it in our eslint installation.

1. [x] Stop dealing with whitespaces
2. [ ] Introduce `prettier`

XVincentX:
I will hold on with this until we have almost 0 PR to review/merge, or rebasing will be a mess.

DrMegavolt:
after new PR with git pre commit hook. just add `-n` to skip check for this commit

dotBits:
since is available pre-commit hook with "lint", what about to create a list of requirements in order to hook that specific eslint configuration?

XVincentX:
Prettier is not happening. I've been using it for a while in Stoplight and I haven't seen any real value out of it (if not increasing the time to commit something). Definitely not worth it.

PR #541, opened 2017-11-17
Fix the formatting problem forever and ever.

XVincentX:
Connect #493 

Upgrading to the latest version of eslint let me run a format fix on the whole codebase in one single hit.

If we merge this, we can finally stop dealing with whitespaces and review real code.

Next step is prettier, but I need time to figure out what's the best way to introduce it without having to switch mental model.


package.json
```diff
@@ -56,7 +56,8 @@
       "node_modules",
       "coverage",
       "migrations",
-      "test"
+      "test",
+      "bin"
```
DrMegavolt: it is location of CLI tool, it should be checked

XVincentX: Ok, but Codeconv will not pass, as the formatting process has touched parts of the code that aren't tested (such as `bin`)

XVincentX:
@DrMegavolt Done. We'll never see the big green here unfortunately, but it should be ready to go.

XVincentX:
@DrMegavolt Rebased with current master. How do you feel about this monster?

PR #543, opened 2017-11-17
add RequestID for each request

DrMegavolt:


altsang:
I believe this has become `requestId` or something alone those lines - can we update the issue description

XVincentX:
@altsang 
Just a clarification — this pull request will add this ID for each single request that's hitting the gateway, even if you're not interested in any tracking feature at all.

Do we want go in this way? Wouldn't it be better to be a core policy — so you pay only for what you use?

altsang:
yes , i believe that’s fine and also reasonable 

i just didn’t want it to be called “trace” anything because that’s not what it’s doing

altsang:
core policy like a system wide setting that can be turned on and off? that is fine too

XVincentX:
@altsang No, we do not have such feature right now. In the current PR the `requestID` will always be added, whether you want it or not.

If we make it as a policy, the user would be responsible to put the ID in the request or not.

altsang:
i rather not make it a policy on its own, it can be part of core and it can
be turned off and on through system.config.yml

way too granular to be a policy on its own, when we introduce a correlation
policy we can add parameters for real tracing use cases including
instrumentation and audit/compliance

this is just a request identifier

altsang:
so I looked at this outside of just the Waffle.io comments, it's fine - it adds a variable as part of egContext object, if later folks don't like this we can put in a flag to turn it off in `system.config.yml` for now it's fine

lib/gateway/context.js
```diff
@@ -1,7 +1,13 @@
 const vm = require('vm');
-
-function EgContextBase () {}
-
+const uuid62 = require('uuid-base62');
+function EgContextBase () {
+  this.requestId = uuid62.v4();
```
kevinswiber: Any linter advice on `requestId` vs `requestID`.  I tend to follow the latter pattern as that's what built-in JavaScript functions adhere to (e.g., `encodeURI`).

kevinswiber:
Looks good to me.  Just one question on the case of the property name.

lib/gateway/context.js
DrMegavolt: ok, ObjectID, MySQL, GraphQL kind of thing.

XVincentX:
That makes sense to me as well.

DrMegavolt:
Changed to requestID, will update 546 as well

Issue #502, opened 2017-11-04
canonical objects: standardize across Express middleware objects


altsang:

altsang:
_As a user and developer of Express Gateway, I'd like to be able to understand what Express Gateway utilizes canonically so that I can map Express middleware that utilizes the same object under a different name_

example:
the user's ID:

in middelware 1: req.user.id
in middleware 2: req.userid
in EG: ??? to map middleware 1 and 2

PR #546, opened 2017-11-17
Feature/set headers policy connect #502

DrMegavolt:


lib/policies/proxy/proxy.js
```diff
@@ -50,23 +43,37 @@ module.exports = function (params, config) {
...
 
+function prepareHeaders (params, proxyOptions, egContext) {
+  proxyOptions.headers = proxyOptions.headers || {};
+  // Default headers always sent to downstream
+  // TODO: allow configuration
+  if (egContext.requestId) {
+    proxyOptions.headers['eg-request-id'] = egContext.requestId;
+  }
+  proxyOptions.headers['eg-consumer-id'] = egContext.consumer && egContext.consumer.id
+  ? egContext.consumer.id : 'anonymous';
```
XVincentX: Or — we could not send the header at all if the consumer is not specified.

lib/policies/proxy/proxy.js
```diff
@@ -21,13 +14,13 @@ module.exports = function (params, config) {
   const proxyOptions = parseProxyOptions(params.proxyOptions, endpoint);
 
   if (!endpoint) {
-    throw new ConfigurationError(
+    throw new Error(
```
XVincentX: 👍

DrMegavolt:
@XVincentX , I've updated the schema, except for making property required. 
required is something that will make policy fail (like no serviceEndpoint in proxy) 

on the other hand `headers` will work fine

lib/policies/proxy/proxy.js
DrMegavolt: equivalent approaches. 
let's send it for now. we can make username configurable or provide option to disable it in future. 

I like sending the header to explicitly show that user has not passed auth

lib/policies/headers/headers.js
```diff
@@ -0,0 +1,14 @@
+module.exports = function (params) {
+  return (req, res, next) => {
+    if (params.forwardHeaders) {
+      const prefix = params.headersPrefix || params.headerPrefix || '';
```
XVincentX: @kevinswiber Did you discuss around this one as well?

lib/policies/headers/index.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  policy: require('./headers'),
+  schema: {
+    type: 'object',
+    properties: {
+      headersPrefix: {
+        type: 'string'
+      },
+      forwardHeaders: {
```
XVincentX: I saw a comment from you where you explained why this can't be mandatory, but I lost it during the rebase. Could you re-explain that?

DrMegavolt: From my understanding 
if the policy will break if not provided - mandatory 
this one will not break.

XVincentX: Your supposition is correct but — what's the point of having a forward policy if you do not provide the list of headers you want to forward? According to the code, it would result in a no operation. It's like silently swallowing the error.

On the other hand, fail fast can help people understand that the policy is misconfigured.

For instance, let's say I misconfigure the policy by writing `forwardHeader` (missing `s`). 

With the current code, I would neither notice the error and I might get hard times understanding I wrote the word incorrectly.

With a fail fast approach, I would see `Missing forwardHeaders` parameters, and probably notice that I made a typo.

Does that make sense?

DrMegavolt: how required here will help ? Schemas are just helpers for API

To answer the usecase you described I've added the log waning that there is no `forwardHeaders`

think of another use case 
you have configured headers, but want to disable 
it is convenient to just comment  the properties of object and get {} as parameter. 

And another 
ExpressJS allows you to write empty (req,res,next)=>next()

There is no problem in tolerating empty input

XVincentX: > How required here will help ? Schemas are just helpers for API

Hmm, that shouldn't be correct. Schemas are evaluated before the [condition](https://github.com/ExpressGateway/express-gateway/blob/master/lib/conditions/index.js#L11),[policy](https://github.com/ExpressGateway/express-gateway/blob/master/lib/policies/index.js#L31),or plugin gets installed.

It has nothing to do with the Admin API itself. The only thing the Admin API provides is the possibility to query the schema (for the UI, I guess).

DrMegavolt: >Hmm, that shouldn't be correct.
just try it out, you'll be surprised, EG will start as usual  

requests indeed will fail with something like `policyMiddleware is not a function` (very helpful)

But let's imagine things would work as you described
how it will help for this policy? and what you are trying to achieve ? 

I can still write `forwardHeaders: {}` there is no way schema can validate keys. It will still be empty and policy will run in `(r,res,next)=>next()` mode. 

@kevinswiber Please check this out and tell if required should be part of the scheme. I'll add it if you thing it is needed in the morning

P.S. This schema validation thing requires another iteration and good vortex on how it suppose to work. for now the only working part is admin API validation

XVincentX: > requests indeed will fail with something like policyMiddleware is not a function (very helpful)

A warning should be issued — and you're absolutely right, it's not failing fast because it's returning `null` instead of throwing an error or doing something else, and the error message is not clear.

> I can still write forwardHeaders: {} there is no way schema can validate keys. 

I think — but I need to verify — there should be the possibility to verify that a json object is not empty — but in this case I do not care, because it has been set empty explicitly from the user, and not considered empty by the gateway itself.

> This schema validation thing requires another iteration and good vortex on how it suppose to work. for now the only working part is admin API validation

Totally true. If **this** is the reason why you want to hold on then I'm fine with reiterating later. I've added a point in the relative issue. Let's merge and 🚀 this.

PR #548, opened 2017-11-19
Remove uuid-base62

XVincentX:
This PR will remove `uuid-base62` package (that's internally using multiple deprecated packages). We can use `uuid/v4` directly or, when really needed, `Buffer.from(uuid()).toString('base64')->NjY1Mzk2MTQtZGU5OC00ZGM1LTg0ZDgtOWRkNjVkYTQ2MzIz`

### DIFF
--- a/lib/gateway/context.js
+++ b/lib/gateway/context.js
@@ -1,8 +1,8 @@
 const vm = require('vm');
+const uuid = require('uuid/v4');
 
-const uuid62 = require('uuid-base62');
 function EgContextBase () {
-  this.requestID = uuid62.v4();
+  this.requestID = uuid();
 }
 Object.defineProperty(EgContextBase.prototype, 'egContext', {
   get () {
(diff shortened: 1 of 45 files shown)


DrMegavolt:
I didn't realize that you plan to remove it. 
Actually base62 is kind of a standard for key-auth generation. 
In tests it is used more as a convenience. It is easier to see diff of base 62 rather than long uuid 

And the main point behind base62 is that it is url safe. 

What is the problem with base62 ?

XVincentX:
I should stop to open wild PRs around and keep my tests s branches 😺 

Anyway I was just playing around with the idea — my rationale behind that is

1. That library is doing two things in one: generate uuid *and* encoding those as base62.
2. Its dependencies are deprecated (`node-uuid`)
3. [base62.js](https://github.com/andrew/base62.js/) could be a good replacement
4. 99.9% of the times this library is used in tests

DrMegavolt:
https://www.npmjs.com/package/uuid62 seems to be drop in replacement.

PR #551, opened 2017-11-21
Improve auto deploy tag regexp

XVincentX:


.circleci/config.yml
```diff
@@ -64,7 +64,7 @@ workflows:
       - deploy:
           filters:
             tags:
-              only: /^v\d\.\d\.\d/
+              only: /v[0-9]+(\.[0-9]+)*/
```
DrMegavolt: How about
v\d+(\.\d+){2} 

the problem I see is v1.4 is also valid as well v1

XVincentX: Good idea, this will prevent to make mistakes with tags.

.circleci/config.yml
```diff
@@ -64,7 +64,7 @@ workflows:
       - deploy:
           filters:
             tags:
-              only: /^v\d\.\d\.\d/
+              only: /v\d+(.\d+){2}/
```
DrMegavolt: it should be escaped dot `\.\d+` 
or it can be as it was before just with stars 
/^v\d*\.\d*\.\d*/

DrMegavolt: regex101 to test

XVincentX: Gosh! I'll fix it now.

XVincentX:
@DrMegavolt Is it ok now?

PR #555, opened 2017-11-22
Introduce JWT token type

### LINKED ISSUE #419: JWT Plugin: verification

XVincentX:
Connects #419 
Closes #419 

The following PR will introduce a JWT verification policy to the gateway.
People will be able to protect their API and check for a token verification strategy with a secret (HS256) or with a certificate (RS256).

### Notes

- I figured out while implementing the thing that we have a lot of model copies spread around the repository. https://github.com/ExpressGateway/express-gateway/pull/555/commits/ff658ab78fe6bfb2b9d801a5bbd4094e99f58415. As all of these are basically equal, if you guys are ok I'd delete those and make sure they all refer to the same file. I created #556 to track this eventually. @DrMegavolt 

- I tried to write a comprensive test suite. It's testing the whole part with secrets and certificates, checking that the issuer is correct, expiration and not before are checked correctly and so on and so forth. You'll decide if it's enough or not.

- Given that `key-auth` and `jwt` are working almost in the same way (regarding the credential creation/modification and deletion) I've added another condition to the `if` branches where appropriate. We all agree that's not optimal, but that'll do for now, I guess.

- What should be the name of the policy? `jwt`? `jwt-auth`?`jwt-verify`? @altsang @kevinswiber

### DIFF
--- a/lib/policies/jwt/extractors.js
+++ b/lib/policies/jwt/extractors.js
@@ -1,0 +1,9 @@
+const passportJWT = require('passport-jwt');
+
+module.exports = {
+  'header': passportJWT.ExtractJwt.fromHeader,
+  'body': passportJWT.ExtractJwt.fromBodyField,
+  'query': passportJWT.ExtractJwt.fromUrlQueryParameter,
+  'authScheme': passportJWT.ExtractJwt.fromAuthHeaderWithScheme,
+  'authBearer': passportJWT.ExtractJwt.fromAuthHeaderAsBearerToken
+};
--- a/lib/policies/jwt/extractors.js
+++ b/lib/policies/jwt/extractors.js
@@ -2,7 +2,6 @@ const passportJWT = require('passport-jwt');
 
 module.exports = {
   'header': passportJWT.ExtractJwt.fromHeader,
-  'body': passportJWT.ExtractJwt.fromBodyField,
   'query': passportJWT.ExtractJwt.fromUrlQueryParameter,
   'authScheme': passportJWT.ExtractJwt.fromAuthHeaderWithScheme,
   'authBearer': passportJWT.ExtractJwt.fromAuthHeaderAsBearerToken
--- a/lib/policies/jwt/index.js
+++ b/lib/policies/jwt/index.js
@@ -19,9 +19,6 @@ module.exports = {
       jwtExtractorField: {
         type: 'string'
       },
-      issuer: {
-        type: 'string'
-      },
       audience: {
         type: 'string'
       }
--- a/lib/policies/jwt/index.js
+++ b/lib/policies/jwt/index.js
@@ -5,10 +5,10 @@ module.exports = {
   schema: {
     type: 'object',
     properties: {
-      secret: {
+      secretOrPubKey: {
         type: 'string'
       },
-      secretFile: {
+      secretOrPubKeyFile: {
         type: 'string'
       },
       jwtExtractor: {
@@ -27,6 +27,6 @@ module.exports = {
       }
     },
     required: ['jwtExtractor'],
-    oneOf: [{ required: ['secret'] }, { required: ['secretFile'] }]
+    oneOf: [{ required: ['secretOrPubKey'] }, { required: ['secretOrPubKeyFile'] }]
   }
 };
--- a/lib/policies/jwt/index.js
+++ b/lib/policies/jwt/index.js
@@ -1,0 +1,32 @@
+const extractors = require('./extractors');
+
+module.exports = {
+  policy: require('./jwt'),
+  schema: {
+    type: 'object',
+    properties: {
+      secret: {
+        type: 'string'
+      },
+      secretFile: {
+        type: 'string'
+      },
+      jwtExtractor: {
+        type: 'string',
+        enum: Object.keys(extractors),
+        default: 'authBearer'
+      },
+      jwtExtractorField: {
+        type: 'string'
+      },
+      issuer: {
+        type: 'string'
+      },
+      audience: {
+        type: 'string'
+      }
+    },
+    required: ['jwtExtractor'],
+    oneOf: [{ required: ['secret'] }, { required: ['secretFile'] }]
+  }
+};
--- a/lib/policies/jwt/index.js
+++ b/lib/policies/jwt/index.js
@@ -21,6 +21,9 @@ module.exports = {
       },
       audience: {
         type: 'string'
+      },
+      issuer: {
+        type: 'string'
       }
     },
     required: ['jwtExtractor'],
--- a/lib/policies/jwt/index.js
+++ b/lib/policies/jwt/index.js
@@ -24,9 +24,13 @@ module.exports = {
       },
       issuer: {
         type: 'string'
+      },
+      checkCredentialExistence: {
+        type: 'boolean',
+        default: true
       }
     },
-    required: ['jwtExtractor'],
+    required: ['jwtExtractor', 'checkCredentialExistence'],
     oneOf: [{ required: ['secretOrPubKey'] }, { required: ['secretOrPubKeyFile'] }]
   }
 };
--- a/lib/policies/jwt/jwt.js
+++ b/lib/policies/jwt/jwt.js
@@ -1,0 +1,1 @@
+module.exports = {};
--- a/lib/policies/jwt/jwt.js
+++ b/lib/policies/jwt/jwt.js
@@ -1,1 +1,26 @@
-module.exports = {};
+const fs = require('fs');
+const passport = require('passport');
+const JWTStrategy = require('passport-jwt').Strategy;
+const extractors = require('./extractors');
+
+module.exports = function (params) {
+  const secret = params.secretFile ? fs.readFileSync(params.secretFile) : params.secret;
+
+  passport.use(new JWTStrategy({
+    secretOrKey: secret,
+    jwtFromRequest: extractors[params.jwtExtractor](params.jwtExtractorField),
+    audience: params.audience,
+    issuer: params.issuer
+  }, (jwtPayload, done) => {
+    if (!jwtPayload) {
+      return done(null, false);
+    }
+
+    return done(null, jwtPayload);
+  }));
+
+  return function (req, res, next) {
+    params.session = false;
+    passport.authenticate('jwt', params, params.getCommonAuthCallback(req, res, next))(req, res, next);
+  };
+};
--- a/lib/policies/jwt/jwt.js
+++ b/lib/policies/jwt/jwt.js
@@ -11,13 +11,14 @@ module.exports = function (params) {
   passport.use(new JWTStrategy({
     secretOrKey: secret,
     jwtFromRequest: extractor,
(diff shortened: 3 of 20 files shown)


lib/policies/jwt/extractors.js
```diff
@@ -0,0 +1,8 @@
+const passportJWT = require('passport-jwt');
+
+module.exports = {
+  'header': passportJWT.ExtractJwt.fromHeader,
+  'query': passportJWT.ExtractJwt.fromUrlQueryParameter,
+  'authScheme': passportJWT.ExtractJwt.fromAuthHeaderWithScheme,
```
DrMegavolt: is it not in use ? I believe we can do same logic as in key-auth - allow schema override and header name

MyHeader: MySchema token

lib/policies/proxy/proxy.js
```diff
@@ -56,12 +56,13 @@ module.exports = function (params, config) {
     const targetUrl = balancer.nextTarget();
     proxyOptions.target = Object.assign(proxyOptions.target, targetUrl);
 
-    logger.debug(`proxying to ${targetUrl}, ${req.method} ${req.url}`);
+    logger.debug(`proxying to ${targetUrl.href}, ${req.method} ${req.url}`);
```
DrMegavolt: 👍

test/policies/jwt/jwt.test.js
```diff
@@ -0,0 +1,171 @@
...
+              }
+              ]
+            }
+          }
+        };
+
+        return db.flushdb()
+          .then(() => userService.insert({
+            username: idGen.v4(),
+            firstname: 'Vincenzo',
```
DrMegavolt: In general, Can we use some superhero theme or some not real names for tests?

XVincentX: Yes, it makes sense. `Clark Kent` will do the job.

We could also emply [faker](https://github.com/marak/Faker.js/) but I do not think we really need this right now.

DrMegavolt: took a look at faker, it is awesome. Eventually we can use it.

XVincentX: Well check also this then: it's mind blowind. [json-schema-faker](https://github.com/json-schema-faker/json-schema-faker)

lib/policies/jwt/extractors.js
XVincentX: If by "not in use" you mean not tested — it is simply because it's the same as `header` extractor but with a precise scheme. I can write a test for that, though.

The reason to include it is that with a single additional parameter you can handle all the use cases such as:

| extractor  	| parameter      	|
|------------	|----------------	|
| header     	| headerName     	|
| query      	| queryParameter 	|
| authScheme | schemaName     	|
| bearer     	| null           	        |

I'm not sure we should make `MyHeader: MySchema token` a viable choice.
I'll summon @altsang and @kevinswiber to decide.
My point is that `Schema Token` is only valid in an `Authorization` header. In all the other cases it does not have any special meaning.

I'd force the people to follow the standards — but **tha bosses** 😎  will decide on that.

DrMegavolt: authScheme - the default value is `JWT`
`Authorization: JWT JSON_WEB_TOKEN_STRING.....`
and the default header name is `Authorization`

this is what the doc says https://www.npmjs.com/package/passport-jwt
and it allows schema-less variant 

and `bearer` is just one of the options. 

So if you just expose original params of the lib we are all good.

XVincentX: If I understand correctly your question, then this thing is happening already.

Extractor param can be specified can be specified using the [jwtExtractorField parameter](https://github.com/ExpressGateway/express-gateway/pull/555/files#diff-ed83d52e12a6ff25c43a778db10e8cd4R19).
This parameter get passed the extractor [here](https://github.com/ExpressGateway/express-gateway/pull/555/files#diff-454b54463ecac8b096621c36034aab08R9).

So for example you can do:
```yml
jwtExtractor: bearer
```

```yml
jwtExtractor: header
jwtExtractorField: customAuthorizationHeader
```

```yml
jwtExtractor: schema
jwtExtractorField: customSchema
```

```yml
jwtExtractor: query
jwtExtractorField: jwtKey
```

Is that what you're talking about?

DrMegavolt: yes, it seems to be fully configurable with good defaults. 
so no problems here, users can configure `MyHeader: MySchema token` if they want to

lib/policies/jwt/jwt.js
```diff
@@ -0,0 +1,41 @@
...
+  passport.use(new JWTStrategy({
+    secretOrKey: secret,
+    jwtFromRequest: extractor,
+    audience: params.audience
+  }, (jwtPayload, done) => {
+    if (!jwtPayload || !jwtPayload.iss) {
+      return done(null, false);
+    }
+
+    services.credential.getCredential(jwtPayload.iss, 'jwt')
```
kevinswiber: Per our discussion, we need a better way to uniquely identify a JWT credential, perhaps some combination of `iss`, `sub`, and `aud`.

XVincentX: Done in 36205fdbabe6a3a846e908ef76f170d9419d9d46

lib/policies/jwt/extractors.js
XVincentX: I've verified and unfortunately the scenario you posted won't be possible.

`passport-jwt` assumes that if you want to use a schema, then the header is fixed and is `Authorization`.

If you want to use a custom header, then you won't have a schema.
The assumption they're making is reasonable to me.

We can implement the behavior you were mentioning, if required, writing a custom jwt extractor. @kevinswiber @DrMegavolt 

Let me know what it's your idea around.

DrMegavolt: you are not limited to extrators provided by the lib 
https://github.com/themikenicholson/passport-jwt/blob/d5af311dce342bd8038a383b86fd8735d4375ac9/lib/extract_jwt.js#L53

XVincentX: > Less is More.
_John Maeda_

The default value is indeed `authBearer`. I think the plugin as it is is providing enough customizability to have the 80% of use cases. We could wait for feedback and go from there. Does that make sense?

kevinswiber:
@XVincentX With this new approach, the policy optionally includes static `iss` and `aud` claims and then matches the `sub` claim to the credential key.  Is that right?

lib/policies/jwt/index.js
```diff
@@ -0,0 +1,32 @@
+const extractors = require('./extractors');
+
+module.exports = {
+  policy: require('./jwt'),
+  schema: {
+    type: 'object',
+    properties: {
+      secret: {
```
kevinswiber: I think `secret` and `secretFile` should be `secretOrPublicKey` and `secretOrPublicKeyFile`, right?  It can be a public key depending on the algorithm (HMAC, RSA, ECDSA).

lib/policies/jwt/jwt.js
```diff
@@ -0,0 +1,42 @@
...
+  const extractor = extractors[params.jwtExtractor](params.jwtExtractorField);
+
+  passport.use(new JWTStrategy({
+    secretOrKey: secret,
+    jwtFromRequest: extractor,
+    audience: params.audience,
+    issuer: params.issuer
+  }, (jwtPayload, done) => {
+    if (!jwtPayload || !jwtPayload.sub) {
+      return done(null, false);
```
kevinswiber: I think we need an option here where we _don't_ validate the user in Express Gateway.  There are plenty of use cases where I just want to verify the integrity of the JWT token and pass it along.  If using an external Identity Provider (like Auth0), it will be a bit of a pain to have to import and synchronize those user records.

kevinswiber: I think we need an option here where we _don't_ validate the user in Express Gateway.  There are plenty of use cases where I just want to verify the integrity of the JWT token and pass it along.  If using an external Identity Provider (like Auth0), it will be a bit of a pain to have to import and synchronize those user records.

XVincentX:
@kevinswiber Correct. `issuer` and `audience` can be statically configured in the policy section and they will be verified by the middleware automagically.

lib/policies/jwt/jwt.js
XVincentX: Hold on sir — we're not validating the user here, as we do not have such information (or probably we have, but we don't know where it's stored). `sub` has the credential ID here, not the user ID.

lib/policies/jwt/index.js
XVincentX: That's correct. You can technically put the public key there as a string.
Can we try to find a more generic name that would encompass both cases?

`secretOrPublicKey` isn't very appealing to me. I'd rather try to find a generic name that would make users understand this parameter means this or that, such as `signature`, something like that.

XVincentX: Tyk is calling this field `jwt_source`.
Kong is using two different fields: `secret` and `rsa_public_key`

XVincentX: Done in 6c77e3c

lib/policies/jwt/jwt.js
kevinswiber: Pardon my slip.  Yes, it's the credential ID.  The point is still the same.  There are plenty of projects that won't use our identity management system at all and only need to verify the token before proxying to a backend.  We should support that use case.

XVincentX: The credential ID is not really part of the identity system, it's nothing more than a list of keys the gateway knows that might come.

Creating a new key in the gateway should not really be a big deal. Let's keep in mind that a JWT key is per consumer, not per user: 1 key can handle millions users. I don't see any giant amounts of data to import, in the eventuality of a migration to EG.

If we renounce on this, the gateway would trust any JWT with any content inside, as long the signature is valid, which I believe it's harmful.

Keep in mind, for instance, that JWT tokens cannot be revoked. Is a system gets compromised, the tokens will continue to flow and to be validated correctly as long you do not replace your secret or private key. This could break all the clients.

Having the keyId stored in the gateway as a mandatory parameters will let the user have a way to revoke a consumer from doing requests, if necessary. I believe it's a valuable security feature to have.

I can live with such a flag, but I am not completely sure it's the way to go.

Maybe @altsang has his opinion on that as well?

XVincentX: @kevinswiber 
I've implemented that flag with tests as well:
Feature: bd43b46
Tests: b49682e

XVincentX:
Ok! @DrMegavolt do you want to have another look?

DrMegavolt:
I have last comment -  hardcoded ports `http://localhost:6057`
For example using the helper below or `findOpenPorts` helper
https://github.com/ExpressGateway/express-gateway/blob/master/test/common/gateway.helper.js#L23

XVincentX:
Good point! I'll fix and merge this 👍

PR #559, opened 2017-11-25
DRY Config class and disable initial discovery event

### LINKED ISSUE #547: Plugins policies seem to be loaded twice or even three times
I noticed, while developing plugins, that exported policies as well conditions functions get called twice or even three times, although I would expect a single call, during the loading or eventual other loadings if there's an hot reloading event.

Try to understand why — whether it's a bug or a normal behavior and document it eventually so people willing to develop plugins do not get scared.

XVincentX:
Connect #547 
Closes #547 

This pull request will refactor a bit the `Config` class. 

1. It will extract a generalized method for loading config files, as the code is the same between the two functions.

2. It will tell `chokidar` to avoid the initial discovery event for `add`. This was the cause behind #547 

3. It will use a single watcher for both files instead of instantiating two of them

There's definitely still room for improvement, but this should be hopefully a good start.

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -38,9 +38,17 @@ class Config {
       }
     }
     log.debug(`${baseProp}Path: `, configPath);
-  }
+  };
+
+  reloadConfigFile (name, watchEvent) {
+    const type = name === this.systemConfig ? 'system' : 'gateway';
+    log.info(`${watchEvent} event on ${name} file. Reloading config file.`);
+    type === 'system' ? this.loadSystemConfig() : this.loadGatewayConfig();
+    eventBus.emit('hot-reload', { type, config: this });
+  };
 
   loadSystemConfig () { this.loadConfig({ baseName: 'system.config', baseProp: 'systemConfig' }); }
+
   loadGatewayConfig () { this.loadConfig({ baseName: 'gateway.config', baseProp: 'gatewayConfig' }); }
 
   loadModels () {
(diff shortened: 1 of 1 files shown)


XVincentX:
Yes @DrMegavolt — this PR will solve #547 

Basically when the gateway was loaded 2 hot reload events were generated (although no file have been changed), therefore all the pipelines were reloaded twice for no reason.

lib/config/config.js
```diff
@@ -70,41 +59,9 @@ class Config {
   }
 
   watch () {
-    const watchEvents = ['add', 'change'];
-
-    const watchOptions = {
```
DrMegavolt: just change it to
    const watchOptions = {
      awaitWriteFinish: true,
      ignoreInitial: true
    };

XVincentX:
@DrMegavolt 
Do you have some ideas on how to avoid that ugly `baseProp` thing? I’ll try to think about it later.

I understand you concern, but on the other hand I still think the refactor there makes sense. 
We have 2 pairs of functions and 2 watchers that are doing the same things, except for input differences. What if we change the logic? We would need to copy and paste the implementations again and change the object names :(

DrMegavolt:
I don't see any real benefit of refactoring this part. This code is not complex and does not change often. We definitely have more important tasks to do. 

But there is a feature that is coming: env vars support. and later redis support and later SQL db support. And think of implementing the feature instead of just renaming variables and extracting some functions. 

https://docs.google.com/document/d/1HrC75oHWeQG-50JjglR1UkkuzV4JUaUyrpDiMtE7Wgk/edit
I think that env vars support can be a decorator (processor) *after* config provider load initial value

so if everybody agree to that the task can be to make current code as `file` provider + implement a post processor for env vars support.

closing since you fixed bug in different PR.

XVincentX:
I'd be willing to make another attempt if you guys agree. I can think a better solution for that part — although that method `loadConfig` was meant to be private. Ugly, but hidden.

DrMegavolt:
I agree that config providers are big and need to be broken but I don't like the idea of wild refactoring. 

If there is a clear vision of public interface of config provider I have no problem if we refactor current code to implement that interface

I propose to review my spec or create a new one and finalize interface first, before doing something in code 

Original proposal of `ConfigProviderInterface:`
* Load - gets entire configuration (system + gateway) from db etc.
* Set(‘key’, object)
* Call eventBus.emit(‘hot-reload’, new_config) if capable

PR #562, opened 2017-11-25
Console logger prints warns by default

XVincentX:
This pull request will show on console warns and errors by default.
Warns are important messages (such as Json Schema validation error) that won't prevent
the gateway from loading, but it's definitely something the user should look at.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -3,55 +3,55 @@ const winston = require('winston');
 // system config will provide env vars support
 winston.loggers.add('EG:gateway', {
   console: {
-    level: process.env.LOG_LEVEL_GATEWAY || process.env.LOG_LEVEL || 'error',
+    level: process.env.LOG_LEVEL_GATEWAY || process.env.LOG_LEVEL || 'warn',
     colorize: true,
     label: 'EG:gateway'
   }
 });
 winston.loggers.add('EG:policy', {
   console: {
-    level: process.env.LOG_LEVEL_POLICY || process.env.LOG_LEVEL || 'error',
+    level: process.env.LOG_LEVEL_POLICY || process.env.LOG_LEVEL || 'warn',
     colorize: true,
     label: 'EG:policy'
   }
 });
 winston.loggers.add('EG:admin', {
   console: {
-    level: process.env.LOG_LEVEL_ADMIN || process.env.LOG_LEVEL || 'error',
+    level: process.env.LOG_LEVEL_ADMIN || process.env.LOG_LEVEL || 'warn',
     colorize: true,
     label: 'EG:admin'
   }
 });
 winston.loggers.add('EG:test', {
   console: {
-    level: process.env.LOG_LEVEL_TEST || process.env.LOG_LEVEL || 'error',
+    level: process.env.LOG_LEVEL_TEST || process.env.LOG_LEVEL || 'warn',
     colorize: true,
     label: 'EG:test'
   }
 });
 winston.loggers.add('EG:config', {
   console: {
-    level: process.env.LOG_LEVEL_CONFIG || process.env.LOG_LEVEL || 'error',
+    level: process.env.LOG_LEVEL_CONFIG || process.env.LOG_LEVEL || 'warn',
     colorize: true,
     label: 'EG:config'
   }
 });
 winston.loggers.add('EG:db', {
   console: {
-    level: process.env.LOG_LEVEL_DB || process.env.LOG_LEVEL || 'error',
+    level: process.env.LOG_LEVEL_DB || process.env.LOG_LEVEL || 'warn',
     colorize: true,
     label: 'EG:db'
   }
 });
 winston.loggers.add('EG:plugins', {
   console: {
-    level: process.env.LOG_LEVEL_PLUGINS || process.env.LOG_LEVEL || 'error',
+    level: process.env.LOG_LEVEL_PLUGINS || process.env.LOG_LEVEL || 'warn',
     colorize: true,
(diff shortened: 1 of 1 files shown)


DrMegavolt:
It is so debatable. 
For example, Common node module pattern is silence unless DEBUG=* is specified. And EG is kind of a module 

Also there are tools where common practice is to use WARN as default. 

It may be that INFO should be the default, it is typical to output that kind of level to file, however, it seems to be too much for console. (Some routing messages are INFO some are DEBUG, helpful for API Endpoint configuration)


There is a long story of postponing normal log solution, but I don't believe changing to WARN will help. 
We need to have configurable log levels, log format for each log target and configurable targets like ELK, syslog etc.  
As well as walking through EG to add new logs, adjust levels of messages, standardize error codes like NODE.js is doing it and a lot of that 

@altsang @kevinswiber alert(`Vortex`) ^^

XVincentX:
> It is so debatable. 

It is! This is probably one of the topics where no matter the love you put into, there's going to be always somebody unhappy 😱 

I'll give you my point of view so we can discuss it later with **tha bosses** 😎 

> For example, Common node module pattern is silence unless DEBUG=* is specified.

True, but I think that's another story here. 
The `DEBUG` module is usually used for _libraries_ that are included into your project, such as `express`, `ioredis`, which aren't supposed to log anything (libraries can't effectively have a story for logging). Therefore they use DEBUG because:

1. It's completely disabled by default
2. It logs in the console and only in the console — no dependencies.

> And EG is kind of a module.

I think Express-Gateway is not a library you should include in your project. Express-Gateway is a product and it should be treated as a product. It should have a complete logging story (and we're probably using the right tool for it, `winston`)

> Also there are tools where common practice is to use WARN as default. 

`npm` is one of those. They are silent except for what you should know during the installation process that is — `warning`, `errors` and `notice` messages.

> It may be that INFO should be the default, it is typical to output that kind of level to file, however, it seems to be too much for console. (Some routing messages are INFO some are DEBUG, helpful for API Endpoint configuration)

I agree. `INFO` is too verbose and it does not help.

> There is a long story of postponing normal log solution, but I don't believe changing to WARN will help. 
We need to have configurable log levels, log format for each log target and configurable targets like ELK, syslog etc.  
As well as walking through EG to add new logs, adjust levels of messages, standardize error codes like NODE.js is doing it and a lot of that 

Cool, I'm all in on iterating and improving the log story. 
The reason behind this pull request is — I think that `warn` are one of those messages the user should know as soon as possible.

While I was writing code for the gateway and then testing it out, I found myself multiple times in troubles and I ended up putting multiple `console.log` around and then ultimately attaching a debugger on the gateway itself to see what was going on.

At the end of the day, if `warn` messages would have been put on the console by default, I would have immediately noticed that a plugin wasn't being loaded because of a wrong version specified or for a non-matching JSON schema. Warn messages are messages like "Sir you probably made a mess, but I'll probably try to make it work anyway"

kevinswiber:
Lots of other common logging frameworks have a level named `fatal`, though this seems to be less common in Node.

Typically, `fatal` would mean an error has occurred that is killing the application.  This is critical.  The `error` level is used to communicate errors that are _recoverable_ from the application's perspective.  The `warn` level is used to communicate issues that _could_ be problematic.  `info` is info and `debug` is debug; choosing which to use is a different conversation.

Winston has no concept of `fatal` with its out-of-the-box labels.

Express Gateway can be run as a library or an application.  We have the option of changing log strategies depending on how the code is being executed.  However, I'd prefer to keep this as an external concern with EG logging to stdout/stderr.  Docker, for example, has the ability to use different log drivers.  Logstash has the ability to subscribe to different inputs, as well, such as journald.  There's plenty of ops tooling to handle this.

So, two things:

1. How should Express Gateway differentiate between `error` and `warn` logs?
2. I suggest we stick to stdout/stderr as the only log target for now.

XVincentX:
My view on `error` and `warn`

`error`: You made a mess. I'll try to not crash, but it might happen. Nevertheless, the current operation will probably be aborted/skipped.

`warning`: Something is not correct. You probably forgot a parameter or something is not in the way it should be. But I have a plan B, a fallback, or something, therefore I'll go ahead but you might want to check it out anyway.

Both cases — the product is almost certainly not behaving in the way it has been specified in the configuration. Therefore the user should know unless you tell it to not do so.

I agree with the point 2. Console is fine, people will figure it out how they want to log it.

DrMegavolt:
we can add fatal 
from here we use only 0,1,2,4
{ 
  error: 0, 
  warn: 1, 
  info: 2, 
  verbose: 3, 
  debug: 4, 
  silly: 5 
}

kevinswiber:
I think we should skip `fatal`.  I've never seen it used in Node.  It's basically an `error` that kills the process.  I'm good with @XVincentX's definitions above.  Just need to make sure all logger usages are reporting properly after this change.

Re-stating the definitions just to make sure I'm on the same page:
* `error`: _Express Gateway_ is throwing an error (either system or user initiated).  If the error was recoverable, the process will still be running.  Otherwise, it will shut down.  We should strive to reduce these.
* `warn`: User made a mistake in one of their inputs (e.g., gateway config), but it's recoverable.  It's not critical, but it should draw some attention from ops, as the gateway may not be functioning as they should.

Sound right?

XVincentX:
Yes, I'm fine with that.

> Just need to make sure all logger usages are reporting properly after this change.

Right, but it's something we can do iteratively. Scanning the whole codebase now can be kind of long.

DrMegavolt:
ok so we are lowering defaults to `warn` ?

PR #563, opened 2017-11-27
Add ignoreInitial flag on configuration file watchers

### LINKED ISSUE #547: Plugins policies seem to be loaded twice or even three times

XVincentX:
Follow up on #559 

Connect #547 
Closes #547

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -73,7 +73,8 @@ class Config {
     const watchEvents = ['add', 'change'];
 
     const watchOptions = {
-      awaitWriteFinish: true
+      awaitWriteFinish: true,
+      ignoreInitial: true
     };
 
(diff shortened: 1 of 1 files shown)


XVincentX:
I had the same feeling. I think there's a missing ESLint rule on graph parenthesis that are resolved when you're saving the file with VSCode (that's also doing some formatting).

They shouldn't be that much and can be prevented/solved when the second part of #493 will be completed.

Issue #537, opened 2017-11-15
Follow up ioredis client switch


XVincentX:
1. Re—enable the `db.test.js` test that was checking certificate files. That should not be that hard.

2. Continuously track `ioredis-mock` for argument/reply transformer support here: https://github.com/stipsan/ioredis-mock/issues/322. We cut a corner to make it happen and we should repay the debt once the main library will be fixed.

stipsan:
New release with transformers support just got pushed https://github.com/stipsan/ioredis-mock/releases/tag/v3.4.0

PR #565, opened 2017-11-27
Update ioredis-mock to latest version

### LINKED ISSUE #564: Use transformers implementation in ioredis-mock and remove workarounds in code
Due to original version of ioredis-mock not containing ioredis argument\reply transformers some workarounds were taken to make the code work. 

The problem is now solved and those workarounds can be safely removed

XVincentX:
Connect #564 
Connect #537 
Closes #564

Pending new `ioredis-mock` release

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -28,20 +28,6 @@ if (redisOptions.tls) {
 const Redis = require(emulate ? 'ioredis-mock' : 'ioredis');
 const db = new Redis(redisOptions);
 
-// TODO: ALERT: this is temporary fix before transformers will land into ioredis-mock
-db.prepareHMSET = function (hashKey, obj) {
-  const result = [];
-  let pos = 0;
-  for (const key in obj) {
-    if (obj.hasOwnProperty(key)) {
-      result[pos] = key;
-      result[pos + 1] = obj[key];
-    }
-    pos += 2;
-  }
-  return [hashKey].concat(result);
-};
-
 db.on('ready', function () {
   logger.debug('Redis is ready');
 });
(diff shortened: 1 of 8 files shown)


DrMegavolt:
check out  `return db.hmset(appHashKey(id), 'isActive', 'true', 'updatedAt', String(new Date()));`
style usage, can it be reverted to `{isActive:"true",  updatedAt : String(new Date())) } ` style ?

XVincentX:
Whoa! Good catch! Although it seems like it was an array, not an object: 
https://github.com/ExpressGateway/express-gateway/pull/501/files#diff-a84828d7e5fb2b16bc57843ffd196ee3L110

I'll try once the library is up

XVincentX:
@DrMegavolt It seems to be ok now

DrMegavolt:
`  return db.hset(buildIdKey(type, id), 'isActive', 'false', 'updatedAt', String(new Date()));`
It seems to be a bug. don't know how it was introduced. 

HSET will set only first field
can you change that to hmset and update arguments to object format ?

XVincentX:
@DrMegavolt 
I was trying to figure out why the tests were passing correctly and…

> As of now, Oct 2017, HSET in the current redis version (4.0.2) can also set multiple key-value pairs.

It seems like these commands now behave in the same way.

DrMegavolt:
hm, ok
* HSET is now variadic, and HMSET is considered deprecated (but will be supported for years to come). Please use HSET in new code.
https://raw.githubusercontent.com/antirez/redis/4.0/00-RELEASENOTES

so we can skip it for now.

XVincentX:
I'll create a new issue with veeeery low priority.

DrMegavolt:
I'm not sure we should fix it at all. 
I don't have any credible info about version in flights. 
Consider this https://duo.com/blog/over-18000-redis-instances-targeted-by-fake-ransomware
so it was a lot of redis servers below current at that time 3.2 

and since 4.0 has some breaking changes I guess 3.x is still in wide usage

XVincentX:
That makes sense. Let's talk again about it in…3 years :trollface:

PR #567, opened 2017-11-28
Modify oAuth2 policy to verify JWT tokens as well issue them

### LINKED ISSUE #558: JWT Plugin: Integration with Express-Gateway identity features
### EG as a token issuer

Express Gateway could can also issue a JWT token using the informations we have in the gateway (such as `consumerId`, `consumerName`, the `scopes` array and so on and so forth). Question is, how do we want to issue such token? One way would be through the `oAuth2` plugin. If so, how should this token look like (in terms of payload)?

XVincentX:
Connect #558 
Closes #558 

Requires #569 merged in first.

The following PR will modify the `oAuth2` policy in order to issue a JWT when configured in `system.config` file and if so — the token verification will leverage the JWT verifier policy (#555).

`system.config` file has been modified. In particular the `accessToken` section now contains:

* `tokenType`: when set to `jwt`, it will issue/verify JSON Web Tokens instead of opaque ones
* `issuer`: The issuer for JWT tokens. This could be `Express-Gateway` by default, eventually
* `audience`: Audience for the token
* `subject`: Subject for the token
* `secretOrPrivateKey`: Secret or privateKey to sign the tokens
* `secretOrPrivateKeyFile`: Secret or privateKey file to sign the token

All these parameters are global, as the Gateway is exposing a single `oauth2/*` set of endpoints.
Therefore they're valid for any pipeline, policy and so on.

Based on the `tokenType` parameter, the oauth2 policy will try to verify an opaque token or a Json Web one — this is happening [here](https://github.com/ExpressGateway/express-gateway/pull/567/files#diff-2bd6cfe6a725b80aa7379d1cbeb03644R158).

As I didn't get any indication about the JWT payload shape, I went with:

```javascript
{
    consumerId: 'something'
    scopes: ['read','write']
}
```

I found this set of headers that are being proxied 

| Header                              | Description                                                                    |
| ----------------------------------- | ------------------------------------------------------------------------------ |
| `eg-token-scopes`                   | authorized scopes for the token                                                |
| `eg-token-redirect-uri`             | redirect URI associated with the token                                         |
| `eg-token-id`                       | token ID                                                                       |
| `eg-token-expires-at`               | metadata about when the token expires                                          |
| `eg-token-created-at`               | metadata about when the token was created                                      |
| `eg-token-consumer-id`              | ID of the consumer to whom the token belongs                                   |
| `eg-token-authenticated-user-id`    | The id of the authenticated user that the client is acting on behalf of        |
| `eg-token-auth-type`                | the type of authentication (oauth, basic-auth, key-auth, etc.)                 |
| `eg-consumer-user-id`               | ID if the user that is tied to the application (if consumer is an application) |
| `eg-consumer-updated-at`            | metadata about when the consumer was last updated in Express Gateway           |
| `eg-consumer-redirect-uri`          | redirect URI for the credential                                                |
| `eg-consumer-user-defined-property` | any user defined property associated with the consumer                         |
| `eg-consumer-is-active`             | a boolean indicating whether the consumer is active                            |
| `eg-consumer-id`                    | ID of the consumer                                                             |
| `eg-consumer-created-at`            | metadata about when the consumer was                                           |
| `eg-consumer-type`                  | type of consumer - application or user                                         |

I'd like to know whether I should be sending all these informations.
JWT should small, and this seems to be too much data to me.

Configuration for the verifier part of policy itself is using the same `jwt` configuration part through a [Json Schema reference](https://github.com/ExpressGateway/express-gateway/pull/567/files#diff-95eed83c27121494d8440fc57de9eb5fR9). It's basically using exactly that policy.

Unfortunately I had to make some copy paste around when issuing JWT — but that does not worry me that much given the bigger reorganization of the credential service we'll hopefully approach soon.

---

Please review this carefully, as the context is big and I hope I didn't miss anything here. Thanks!


XVincentX:
Node 6 problem: https://github.com/istanbuljs/nyc/issues/657
Workarounded in 9c84d0f

lib/services/tokens/token.service.js
```diff
@@ -144,7 +145,24 @@ s.revoke = function (accessToken) {
   });
 };
 
-function createInternalToken (criteria, id, token, type) {
+s.createJWT = function (payload) {
+  const accessTokenConfig = config.systemConfig.accessTokens;
+  const secret = accessTokenConfig.secretOrPriKeyFile ? fs.readFileSync(accessTokenConfig.secretOrPriKeyFile) : accessTokenConfig.secretOrPriKey;
```
DrMegavolt: 1) Pri = Private ? 
2) why not resolve it in policy\plugin and make service accept it as parameter? 
current implementation will reread the file on each JWT generation

XVincentX: 1. Yes, it's private. It's probably a very bad combination, I'm open to better ideas.
2. Definitely good point — I recall I was having some problems with that this morning (c2d30d1) but I forgot to go back on it. 👍

DrMegavolt: Private seems to be good word ? why not call it that ?

lib/policies/oauth2/oauth2.js
```diff
@@ -153,6 +155,11 @@ function authenticateLocal (req, clientId, clientSecret, done) {
 }
 
 module.exports = function (actionParams) {
+  if (config.systemConfig.accessTokens.tokenType === 'jwt') {
+    const params = Object.assign({ getCommonAuthCallback: actionParams.getCommonAuthCallback }, actionParams.jwt);
+    return jwtPolicy(params);
```
DrMegavolt: `jwtPolicy(params);` is init of entire policy, it will reregister passport strategy and whatever parsing of options is going there. 

ideally it would be just 
`passport.authenticate(['jwt', 'opaque'], params, params.getCommonAuthCallback(req, res, next))(req, res, next);`

DrMegavolt: and If I understand correctly you can register multiple JWT strategies (or at least parametrize the strategy)
so it will not interfere with other JWT\oAuth2 policies

lib/policies/oauth2/oauth2-server.js
```diff
@@ -119,12 +118,19 @@ server.exchange(oauth2orize.exchange.code((consumer, code, redirectUri, done) =>
 
       if (codeObj.scopes) tokenCriteria.scopes = codeObj.scopes;
 
+      if (config.systemConfig.accessTokens.tokenType === 'jwt') {
+        return tokenService
+          .createJWT({ consumerId: consumer.id, scopes: codeObj.scope })
```
DrMegavolt: should it be `codeObj.scopeS` ?

lib/services/tokens/token.service.js
XVincentX: Ah good question — it's because in the JWT policy the equivalent parameter is called `secretOrPubKeyFile`, and I wanted to kind to stick to the convention.

lib/policies/oauth2/oauth2-server.js
XVincentX: Gosh! Great catch! Fixed in 0992d5d

lib/services/tokens/token.service.js
DrMegavolt: since it is not released, what if we change `pub` to `public`  ? 
I believe it will be easier to maintain code

lib/policies/oauth2/oauth2.js
XVincentX: Actually your comment here made me think about couple of points that we should discuss/figure out (thank you for that!)

Passport is storing all the strategies by name in a [simple javascript object](https://github.com/jaredhanson/passport/blob/master/lib/authenticator.js#L58).

This can be harmful because, in this situation:

```yml
http:
  port: 8080
apiEndpoints:
  api:
    path: /hello
  api2:
    path: /hello2
serviceEndpoints:
  backend:
    url: 'http://localhost:9876'
policies:
  - proxy
  - jwt
pipelines:
  pipeline1:
    apiEndpoints:
      - api
    policies:
      - jwt:
        - action:
          secretOrPubKey: superSecretString
      - proxy:
          - action:
              serviceEndpoint: backend
  pipeline2:
    apiEndpoints:
      - api2
    policies:
      - jwt:
        - action:
          secretOrPubKey: anotherSuperSecretString
      - proxy:
          - action:
              serviceEndpoint: backend
```

The second JWT strategy instance would overwrite the first one (because they have the same name). As they're using different signing keys, the first policy would reject everybody.

This can be prevented using unique strategies name (POC: 6c6489e8c2d94f9e2e8694ee77ae6bc66b0bea9e)

---

Regarding your question, if I understand correctly: would you like to move the `passport.use(JWTStrategy)` during the init phase of the oauth2 strategy, so we do not have to call the whole policy? I am ok with that! Please confirm that I understood what you meant and I'll do the change.

lib/services/tokens/token.service.js
XVincentX: If @kevinswiber agrees (as he was the one suggesting the name) I am ok with the name change.

lib/policies/oauth2/oauth2.js
DrMegavolt: 1) I would like to keep them separate. even they may reuse some code eventually those policies can be split out in different plugins. 

2)there is no need to register using passport.use with all the params. 
you can add the parameters later. at `passport.authenticate` 

see key-auth as example. 
One of the reasons to implement custom strategy was that default one is not that configurable through `passport.authenticate` params

DrMegavolt:
general comment to task description. 
oauth2 no longer proxies headers. it was extracted as headers policy

DrMegavolt:
once all is finished, please write e2e tests that will check scopes with JWT token

XVincentX:
1. Should we update the documentation and say "Hey if you want the headers you should use this policy?"
2. Sure. An E2E test will definitely be a good thing.

lib/policies/oauth2/oauth2.js
XVincentX: > I would like to keep them separate. even they may reuse some code, eventually, those policies can be split out in different plugins.

Ok! If then we'll split these into two plugins, the JWT one can always become a dependency of the first one and reuse its exported features.

> There is no need to register using passport.use with all the params. you can add the parameters later. at passport.authenticate

I've checked this and it seems that `passport-jwt` is completely [ignoring the passed options object](https://github.com/themikenicholson/passport-jwt/blob/master/lib/strategy.js#L90). We can still make it, but we need to make another custom strategy (like what you guys did for `key-auth`), forking the library, right?

DrMegavolt: Strategy is not a magic code that only super experts can write. 
we can always copy it and modify as needed

If it is possible to make a PR to original lib to make it support params it is awesome, or we can embed it in our codebase. For key-auth it was some functional issue (cannot remember) as well as params. So I used a copy approach. 

@kevinswiber ^^

XVincentX: Sure thing — I'm fine forking the library if required and issue a PR.

XVincentX:
### Outstanding things:

- [ ] Write a test to verify scopes are propagated correctly
- [ ] Decide on `Pub-Public/Pri-Private` naming for secret property
- [ ] Decide whether use 6c6489e or fork the jwt middleware to accomodate per-request-options instead of immutable ones

lib/services/tokens/token.service.js
kevinswiber: I would prefer to be explicit and obvious with longer parameter names.  `Pub` => `Public`, `Pri` => `Private`

XVincentX: Done in 93b0f1d

XVincentX: @DrMegavolt N.2 resolved in 58e3ce81aea8206134ce1c8b4ff79b64dd401f23

test/policies/jwt/jwt.test.js
```diff
@@ -194,4 +197,59 @@ describe('JWT policy', () => {
       backend.close(() => gateway.close(done));
     });
   });
+
+  describe('Two JWT policies in two different pipelines', () => {
```
XVincentX: This test fails without 6c6489e

DrMegavolt:
I didn't test but looks good in general. I would approve after demo.

XVincentX:
@DrMegavolt 
I can definitely prepare a demo for that. In meantime you could have a look to the [test file](https://github.com/ExpressGateway/express-gateway/pull/567/files#diff-62e57ae2be09350579623ff3e0e52a0a) where the feature is used. That might help understand how's that working from an user perspective.

One clarification I'd like to ask:
You said it looks good, but if I recall correctly we decided that I would have forked/embedded the JWT library and make it support options per request instead of creating a new middleware instance for each policy. Is this still valid? If so, then there's still some work to be done here.

If not — then ok, I probably missed the discussion as I was out for conference.

DrMegavolt:
@XVincentX as I understand you are no longer using the same strategy with shared config for oauth2\jwt. you are creating unique uuid named strategies for each polices. 

If I understood correctly it allows individual config per policy, which looks good for me. 
Did I get it wrong? 

will take a look at the test

XVincentX:
You get it perfectly! That was the intent!

I'll craft an example for you now.

XVincentX:
@DrMegavolt [I've deployed a sample gateway on Glitch](https://glitch.com/edit/#!/express-gateway-test?path=server.js:30:13).

Just configure your preferred oauth2 client with the credentials and you're good to go.
You should receive back a JWT token.

PR #569, opened 2017-12-01
Password exchange should check for basic-auth credentials

XVincentX:
I discovered what I think it might be a bug in our password flow while I was writing tests for #567.

According to the specs and to multiple source I checked out around, the `password` flow is meant to be used when you totally trust your client and use it to forward **user's** id and password directly to the auth server, without having to make the whole flow and login/consent screen (that is usually used for untrusted clients).

[Source](https://www.slideshare.net/gauravroy/stateless-auth-using-oauth2-jwt)

On the other hand, according to the code, when the Gateway has to validate an userId/password pair is looking for an `oauth2` credential; that is not where the current user's credentials are stored.

They're a `basic-auth` credential type. In the current state, the gateway is instead looking where we store the client ID and secret that is used in to recognize the client executing oAuth2 flow.

This will therefore make any attempt to login fail.

I'm not sure if am correct — could you guys have a look here and confirm, eventually?

Connect #558

### DIFF
--- a/lib/policies/oauth2/oauth2-server.js
+++ b/lib/policies/oauth2/oauth2-server.js
@@ -138,7 +138,7 @@ server.exchange(oauth2orize.exchange.password((consumer, username, password, sco
     .then(consumer => {
       if (!consumer) return done(null, false);
 
-      return authService.authenticateCredential(username, password, 'oauth2')
+      return authService.authenticateCredential(username, password, 'basic-auth')
         .then(user => {
           let scopeAuthorizationPromise;
 
--- a/test/cli/tokens/revoke.test.js
+++ b/test/cli/tokens/revoke.test.js
@@ -24,19 +24,22 @@ describe('eg tokens revoke', () => {
     })
       .then(createdUser => {
         user = createdUser;
-        return adminHelper.admin.credentials.create(user.username, 'oauth2', {secret: 'test'});
+        return Promise.all([
+          adminHelper.admin.credentials.create(user.username, 'oauth2'),
+          adminHelper.admin.credentials.create(user.username, 'basic-auth')
+        ]);
       })
-      .then(createdCred => {
-      // cred = createdCred;
+      .then(([oauthCredentials, basicCredentials]) => {
+        const base64Credentials = Buffer.from(`${oauthCredentials.id}:${oauthCredentials.secret}`).toString('base64');
         return request(app)
           .post('/oauth2/token')
-          .set('Authorization', 'basic ' + (Buffer.from(user.username + ':test').toString('base64')))
+          .set('Authorization', `basic ${base64Credentials}`)
           .set('content-type', 'application/x-www-form-urlencoded')
           .type('form')
           .send({
             grant_type: 'password',
-            username: user.username,
-            password: 'test'
+            username: basicCredentials.id,
+            password: basicCredentials.password
           })
           .expect(200);
       })
--- a/test/oauth/password.test.js
+++ b/test/oauth/password.test.js
@@ -70,8 +70,10 @@ describe('Functional Test Client Password grant', function () {
 
                 credentialService.insertScopes('someScope')
                   .then(() => {
-                    Promise.all([credentialService.insertCredential(fromDbUser1.id, 'oauth2', { secret: 'user-secret' }),
-                      credentialService.insertCredential(fromDbApp.id, 'oauth2', { secret: 'app-secret', scopes: ['someScope'] })])
+                    Promise.all([
+                      credentialService.insertCredential(fromDbUser1.id, 'basic-auth', { password: 'user-secret' }),
+                      credentialService.insertCredential(fromDbApp.id, 'oauth2', { secret: 'app-secret', scopes: ['someScope'] })
+                    ])
                       .then(([userRes, appRes]) => {
                         should.exist(userRes);
                         should.exist(appRes);
@@ -83,7 +85,7 @@ describe('Functional Test Client Password grant', function () {
       })
       .catch(function (err) {
         should.not.exist(err);
-        done();
+        done(err);
       });
   });
 


DrMegavolt:
I don't see any form during that flow. 
In current implementation you need 
```
request
      .post('/oauth2/token')
      .set('Authorization', 'basic ' + credentials)
      .set('content-type', 'application/x-www-form-urlencoded')
      .type('form')
      .send({
        grant_type: 'password',
        username: 'irfanbaqui',
        password: 'user-secret'
      })
```
And spec says 
https://tools.ietf.org/html/rfc6749#section-4.3.1
```
 POST /token HTTP/1.1
     Host: server.example.com
     Authorization: Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW
     Content-Type: application/x-www-form-urlencoded

     grant_type=password&username=johndoe&password=A3ddj3w
```

so it seems to be equivalent. 

What is wrong here ?

XVincentX:
The problem is that currently you need two different pairs of userID-Password depending on the flow you're using, even though you should have only one (if I read the spec correctly).

Here's an example

I set up a policy to use oAuth2, and I want to use it with the **Code Grant Flow** and **Password Flow**

Therefore:

* I create an user called `user`
* I create an application called `appy`
* I configure the oauth2 policy in a pipeline, and I get the things protected.

Now I want to access this resource using the `user` I just created — therefore:

* I create an `oauth2` credential for my `appy` application, and I receive a clientID and a secret
* I create a `basic-auth` credential for `user`, and I get an username and a password
* I try to login following the **Code Grant Flow**; therefore
  - I go to `http://express-gateway/oauth2/authorize` filling the `clientID` and `secret` query parameters
  - I am not logged in, therefore I get redirected to `oauth2/login`
  - I type my username and password (obtained from `basic-auth` credential)
  - I go to the authorize endpoint and, after the consent, I receive a token I can exchange.

Now let's say I want to do the exact same thing with `Password flow`

* I create a web request sending 
  * ClientID
  * Client secret
  * Username (`basic-auth` one — the same as above)
  * Password (`basic-out` one — the same as above)

My request will be rejected now, because the `password` grant is checking for an [oauth2 credential type](https://github.com/ExpressGateway/express-gateway/blob/master/lib/policies/oauth2/oauth2-server.js#L141) and not for a `basic-auth` one.

This means that if I want to login using the password flow, I need to create an `oauth2` credential type for my user and having another userId and password set, which (If I understood the whole flow correctly) is wrong. 

1. The `oauth2` credential type is meant to be listing clients willing to autenticate the users somehow, not to list users credentials
2. Regardless of the flow, the resource owner has the same user id and password, and it should not have different one according to the flow he wants to use. This is correctly handled for all the other flows. Just the `password` one seems to be behaving differently.

I hope this clarifies. If not, I can craft and example and jump into a call!

@kevinswiber @DrMegavolt

DrMegavolt:
ok, it seems you found something. I need more clarifications:
`eg credentials create -c test -t oauth2` 
it gives me `secret`. and `basic-auth` variant will give me password 

What I don't understand why we involve `basic-auth` policy at all?
 should I be able to send 
`username=test&password=my_secret` as body and be authenticated ?

DrMegavolt:
assuming clientid\clientSecret are in the Authorization header as basic encoded value

kevinswiber:
@DrMegavolt It's weird.  This has always been weird.  You have to create a `basic-auth` credential for any user, including one that's only going to be logging in for an Authorization Code grant type flow.  @XVincentX's solution looks correct to me.  But the bigger problem is that this is weird and confusing.

DrMegavolt:
wait, you are asking user to login with guid instead of username ?

DrMegavolt:
https://github.com/ExpressGateway/express-gateway/pull/569/files#diff-7a11becb767d516a27a6c41facd1b7fcR41

I believe this part is not correct. it should be username, and it should not be basic-auth verification.

kevinswiber:
> wait, you are asking user to login with guid instead of username ?

@DrMegavolt 

I believe `authenticateCredential` tries to authenticate by username, as well.  This whole chunk of code could be refactored, really.  I think it actually tries user ID... and if that fails, it tries username... but in this case, it should likely _never_ work for user ID.

DrMegavolt:
1) at least test still should use username. because nobody is using id in username\password way
2) now why I need `credentialService.insertCredential(fromDbUser1.id, 'basic-auth', { password: 'user-secret' }),` ?
why `basic-auth` ? just because I have clientID\Secret passed as encoded `basic` format it doesn't mean I need completely different auth type for user. It is if key-auth was required in that flow.

XVincentX:
> What I don't understand why we involve basic-auth policy at all?
should I be able to send
username=test&password=my_secret as body and be authenticated ?

That won't be enough. In an oAuth2 flow there are two things that should get recognised: the user (that is the resource owner) and the client (that the user is using to access the resource server).

The client is **always** recognised by putting a client id and a client secret as a part of the http request. The flexibility in the oauth flow is that the **user** can be recognised in different ways.

I *think* (but I might be wrong, I'm just reading the code and trying to resemble the original idea) the intention was: 

1. When you want to create a **client credential**, you create an `oauth2` credential type.
2. When you want to create an user able to access a resource, you create a `basic-auth` credential type

> wait, you are asking user to login with guid instead of username ?

It does not have to be. Again — I am going by interpreting the code.

The way I think it was intended to work is

You've got an user (that is a consumer) and you can have multiple representation of it through a set of credentials. Now — you get an `uuid` by default if you do not specify an username, but technically you can always do

`eg credentials create -c user -t basic-auth -p "id=user"`

DrMegavolt:
-p "id=user" should not work 
it is implemented on having credential id as a userId. 
allowing this would open security issue - assigning credential to not my accounts. 

The doc is even saying that you have to have `basic-auth` credential 
but why? 

I want oauth2 flow 
I create user and app
I create `oauth2`credential for them 

now I have username\secret + id\secret 
it seems to be enough for any flow, right ?

Why should I have one more credential type ? 

I'm just thinking it is somehow wrong. Is it possible to do a fix assuming basic-auth does not exists ?

XVincentX:
> The doc is even saying that you have to have basic-auth credential. but why?

No clue guys. I can't help here.

> now I have username\secret + id\secret it seems to be enough for any flow, right ?

Technically yes. It would be a breaking change though (since all the users would have to recreate credentials), but that will work.

> Why should I have one more credential type ?

I would say it's for keeping the things separate and make them reusable. For example, I like the thing that I can use the same set of username/password in an oAuth flow or in a basic-flow. I have an userid and a password, and not a set of them for each one. Your mileage might vary here, anyway.

DrMegavolt:
> I like the thing that I can use the same set of username/password in an oAuth flow or in a basic-flow

but why do you need that? what if key-auth can be composed from clientid + secret generated for oauth2, does it make any good? 

> (since all the users would have to recreate credentials), but that will work.

you found a bug (I believe a big one). I just want that we understand what we are fixing here.  

And the only way I see is to just remove basic-auth from oauth2 and see what is missing.

XVincentX:
> but why do you need that? what if key-auth can be composed from clientid + secret generated for oauth2, does it make any good?

I'm sorry but I do not follow you here anymore. I am missing the relationship with the `key-auth` policy.

> And the only way I see is to just remove basic-auth from oauth2 and see what is missing.

You guys know better than me probably how to proceed here. The reason why I was pushing for `basic-auth` was because I think (reading the source code) it was the original intention by the guy that crafted this.

DrMegavolt:
sorry for confusion. 
I'm just saying that basic-auth can be completely different plugin unrelated to oauth2 
and unlikely should be used in oauth2 policy, same as key-auth, and same reason we are trying to keep JWT and oauth2 separate.  

>The reason why I was pushing for basic-auth was because I think (reading the source code) it was the original intention by the guy that crafted this.

and this is strange right? 

Here is my thinking

We have some part of system  secured using basic-auth, another with oauth2
scopes are not involved, purely authentication, no authorization

There is mobile app that has valid client\clientSecret targeting oauth2 part

I have basic-auth creds for another part. 

Now with this line 
`authService.authenticateCredential(username, password, 'basic-auth')` 
I can use mobile app to login to another part of app I'm not suppose to have access. 

or a person who just knows about mobile app can actually login to the part secured by basic-auth

And If there was a third part, secured by key-auth. You wouldn't expect mobile app to  grant access if you enter keyId\KeySecret, right ? 
so why it is allowed for basic-auth ? 

That is why I believe `authService.authenticateCredential(username, password, 'oauth2')` is actually correct.
And most likely the password flow is working correctly. 


it may be `code flow` have issues.
This part `I type my username and password (obtained from basic-auth credential)`
Maybe it should be 
`I type my username and secret obtained from oauth credential `
Make sense ?

XVincentX:
Aaaaaaah I think I got what you mean!

So your point is — each credential method should have its **own** set of credential and they cannot be shared across multiple policies. Is that what you mean?

In that case, here are my thoughts:
> > The reason why I was pushing for basic-auth was because I think (reading the source code) it was the original intention by the guy that crafted this.

> and this is strange right?

Unfortunately I can't tell you that precisely. It highly depends what was the original intention you guys decided. @kevinswiber or @altsang maybe recall the rationale behind this?

> I'm just saying that basic-auth can be completely different plugin unrelated to oauth2

Correct, and I think it's a good point. I might want to install the `oauth2` plugin but not the `basic-auth` plugin. How would I create credentials then? Good point. We need to discuss this probably.

On the other hand, regarding the example you mentioned and looking at this from an higher lever perspective:

oAuth2 is nothing about authentication, it's again all about authorization. Therefore the idea is to leverage another authentication method you have in your gateway in order to make it, and not to keep another credential database/method because of it.

From this perspective, I would not have any problem in effectively putting a dependency with an authentication method the gateway could offer (`basic-auth` or whatever it's going to be. `face-recognition` could be one in future, who knows).

What I totally agree with you, is that the strong coupling with `basic-auth` is probably wrong. At the moment, our oAuth2 policy must be used in conjunction with `basic-auth` or nothing. I'd rather specify this as a policy parameter. 

What if I have a passwordless authentication method and I want to use it with an oAuth2 flow? I can't (today).

Even if we go with the solution you propose, the policy would probably be independent, but coupled to the internal authentication method bundled with it. And, again, oAuth is not supposed to provide any credential type.

> Now with this line
authService.authenticateCredential(username, password, 'basic-auth')
I can use mobile app to login to another part of app I'm not suppose to have access.

Isn't that we should exactly aim for? If you login on github.com directly or you login using oAuth for a thirty party application, or you use their [desktop](https://desktop.github.com) client you use exactly the same credential set, and not a different one.

DrMegavolt:
oAuth2 is nothing about authentication, it's again all about authorization. 

Oauth2 in current EG implementation has authentication capabilities. 

authenticate means to get user attributes like id, username, email etc. 
and EG can find user by token. 

that is why you don't really need anything before oauth2

> passwordless authentication
client credential flow should work in that way. 

> If you login on github.com directly or you login using oAuth for a thirty party application, or you use their desktop client you use exactly the same credential set, and not a different one.

admin.github.com - basic
repo.github.com - oauth2 

I should not be able to even login to `repo` part with basic creds 
even before I hit authorization part with scopes

Issue #574, opened 2017-12-13
Move custom models (user, applications and credentials) to JSON Schema


XVincentX:
## Preparation

- [x] Understand the role of  `yeoman-generator` in the game. Looking around it seems to be useful only when used with `yo` — that is not employed here at all. 🤔 
- [x] Investigate at what degree we can integrate `yargs` with provided schemas. `check` can do the job with ajv, however it would be great to show detailed informations on the CLI level based on the schema.

## Things to clarify/ask around

- [x] What is the `no-colors` flag for? It's not used anywhere.

- [x] DRY all the models in the application so we do not have to specify them every single time #556 
- [x] Make the switch

Issue #432, opened 2017-09-25
host versus hostname: need consistency


altsang:
the `admin:` section and `apiEndpoints:` use `host:` versus `hostname:` should pick one

- [ ] update generators
- [ ] update docs
- [ ] update CLI
- [ ] update tests

ravikp7:
Hey @XVincentX , what should be picked? `host` or `hostname` ?

XVincentX:
@ravikp7 I believe we should go for `host`. It's important to keep the compatibility with the old configuration file so we should suppor both for a while, emit a warning and remove it in EG 2.0.

ravikp7:
Hmm, makes sense for the compatibility. As per my understanding, tasks are

- Replace `hostname` with `host` in the in the generator files, so that new setups would have updated config:
  - `./bin/generators/gateway/templates/getting-started/config/gateway.config.yml`
  - `./bin/generators/gateway/templates/basic/config/gateway.config.yml`
  - `./bin/eg-generator.js`

- Add `host` support with `hostname` in `./lib/rest/index.js` and emit warnings when `hostname` is found on admin config

- In the schema `./lib/config/schemas/gateway.config.json`, add `host` under admin and leave `hostname` as it is for now

- Replace `hostname` with `host` in `./lib/config/gateway.config.yml`. I also want to know, when this config is used?

- Update `tests` and `docs` according to above changes

And later in `EG 2.0`, remove references to `hostname` completely.

@XVincentX Please let me know if I'm missing something.

XVincentX:
You've got it. The additional task would then be to update our documentation website http://github.com/expressGateway/express-gateway.io — but we usually do that after the changes have been merged in the `master` branch.

Does that make sense to you?

ravikp7:
Yeah, got it.Thanks :)

PR #576, opened 2017-12-13
Deduplicate model instances in the Gateway

### LINKED ISSUE #556: Deduplicate models and configurations
Models and configuration files for the gateway tests, templates and other usages are duplicated multiple times around the repository and from a first look, they're all the same.

Let's delete all the copies and make the tools point to a single file instead of having copies around.

XVincentX:
Connect #574 
Connect #556 
Closes #556 

- [x] This needs #571 before merging this.

This pull request will remove all model instances in the gateway (templates, tests) replacing them with a copy action (given they're all the same) from the templates placed into the `lib/config/models` directory.

Unfortunately I haven't been able to remove the one in `test/config` — because the configuration there is different — but we'll get there hopefully in another PR.

This is a prerequisite to make #581 happen.


bin/eg-generator.js
```diff
@@ -46,15 +46,14 @@ module.exports = class EgGenerator extends Generator {
   // configuration defaults
   _wrapConfig (yargs) {
     return yargs
-      .boolean(['no-color', 'q', 'v'])
```
kevinswiber: Didn't like `--no-color`?  This is mostly for anything using Yeoman output.

XVincentX: Ah — that's where's used! I've been looking around and I wasn't able to find it.

Let me verify one more time then. Hold on on this please 🙏

test/policies/rate-limiter/rate-limit.by.host.test.js
```diff
@@ -25,7 +25,7 @@ describe('rate-limit by host', () => {
                 action: {
                   max: 1,
                   // eslint-disable-next-line no-template-curly-in-string
-                  rateLimitBy: '${req.host}'
+                  rateLimitBy: '${req.hostname}'
```
kevinswiber: What's going on here?  Does Express add a `req.hostname` but not `req.host`?  Seems weird to have to go the header for the latter.  In any case, might want to either change this to use the Host header or change the name of the file/test from `host` to `hostname`.  These mean two different things.

XVincentX: I've checked and they're the same thing.

> express deprecated req.host: Use req.hostname instead

kevinswiber:
Conceivably, each template could use its own set of custom models.  This change prevents that.  Not sure how I feel about that, honestly.  If we merge this and later want to use a custom set of models with a gateway template, we'll have to revert these changes.

test/policies/rate-limiter/rate-limit.by.host.test.js
DrMegavolt: If you are into hostname, please make changes as fix for #432 then
technically, since we are listening on the single port 
hostname is equivalent to host

the only difference would be if you want different rules for https vs http

`host` should allow that separation. but that is highly unlikely scenario

XVincentX:
@kevinswiber That's a valid concern, but on the other hand so far what this so fine grained thing brought was a bunch of copy paste every time we had to update the models.

I'd rather prefer a "copy-on-write" approach rather than having to maintain 30 different things that so far have been perfectly equal.

test/policies/rate-limiter/rate-limit.by.host.test.js
DrMegavolt: but in express 5 it will be back
host is hostname:port

XVincentX: Ah! I didn't know that!
I should check this thing more carefully then. I'll write back once I finish to investigate.

kevinswiber: Maybe they're the same thing in Express's vocabulary, but they're not the same thing to HTTP, so this is confusing.

* hostname: `example.org`
* host: `example.org:80`

DrMegavolt:
@XVincentX @kevinswiber can we then remove hostname from this PR and make any changes as part of #432 ?

XVincentX:
@DrMegavolt Fair enough. I'll remove this commit so we can investigate more on that.

kevinswiber:
> can we then remove hostname from this PR and make any changes as part of #432 ?

Honestly, I don't think this is related to #432 at all.  Isn't #432 about our own config?  Express is going to keep providing `hostname`, and I don't think we should change that.  Might screw with middleware.

test/e2e/hot-reload.test.js
```diff
@@ -13,7 +13,7 @@ const yaml = require('js-yaml');
 const { findOpenPortNumbers } = require('../common/server-helper');
 
 const GATEWAY_STARTUP_WAIT_TIME = 5000;
-const TEST_TIMEOUT = 10000;
+const TEST_TIMEOUT = 11000;
```
DrMegavolt: how it helps ? why 11 sec ?

XVincentX: This is a leftover. I'll remove it.

XVincentX: Done in 00ba93e

PR #581, opened 2017-12-19
Move JS models to JSON Schema models

### LINKED ISSUE #514: Admin API allows to insert an application although if the user is not existing
```bash
eg apps create -u IDoNotExist
```

It will pass correctly, although the user does not exist.
At least a warning should be thrown. At the moment it's confusing as you might inserting things using the user name instead of the user id, and you do not get any warning that it's not going to work.

### LINKED ISSUE #574: Move custom models (user, applications and credentials) to JSON Schema

XVincentX:
Connect #574 
Closes #574
Closes #514 

The following 🐉 /PR will replace the custom JSON Schemaesuqe implementation to use a real JSON Schema for all our models.


test/services/users.test.js
```diff
@@ -23,7 +23,7 @@ describe('User service tests', function () {
         .insert(user)
         .then(function (newUser) {
           const expectedUserProps = ['firstname', 'lastname', 'email', 'isActive', 'username', 'id', 'createdAt', 'updatedAt'];
-          Object.keys(newUser).should.eql(expectedUserProps);
+          should(Object.keys(newUser)).containDeep(expectedUserProps);
```
DrMegavolt: this is kind of `expect(foo).tobe(bar)` syntax
why do you like it? and why it is called should in this case ? 

I expect should style like
`foo.should.be(bar)`

lib/config/models/credentials.json
```diff
@@ -0,0 +1,85 @@
...
+            "array"
+          ],
+          "items": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "jwt": {
+      "type": "object",
```
DrMegavolt: JWT has not scopes verification, right? because there is no common field to store them, so every third party has own way. 

scopes are validated only if JWT is a token type for oauth2, correct?

lib/rest/routes/credentials.js
```diff
@@ -1,8 +1,6 @@
-
 const express = require('express');
 const credentialSrv = require('../../services').credential;
-const usersSrv = require('../../services').user;
-const appsSrv = require('../../services').application;
+const findConsumer = require('../utils/findConsumer');
```
DrMegavolt: why do you need some util if you can put the logic directly into service ?

lib/rest/routes/apps.js
```diff
@@ -8,7 +9,15 @@ module.exports = function () {
   router.post('/', function (req, res, next) {
     const userId = req.body.userId;
     delete req.body.userId;
-    applicationsSrv.insert(req.body, userId)
+
+    findUser(userId)
```
DrMegavolt: this logic should be in the service `userSrv.get(userIdOrUsername)`

lib/services/consumers/application.service.js
```diff
@@ -1,30 +1,30 @@
...
+const SCHEMA = 'http://express-gateway.io/models/applications.json';
 
 const s = {};
 
 s.insert = function (_app, userId) {
-  let app;
   try {
-    app = validateAndCreateApp(_app, userId);
+    return validateAndCreateApp(_app, userId)
+      .then((app) => Promise.all([app, applicationDao.insert(app)]))
```
DrMegavolt: you lost me there. what is `app => Promise.all([app, ...])` ? why app is a promise ?

test/cli/apps/create.test.js
```diff
@@ -247,46 +248,19 @@ describe('eg apps create', () => {
     assert.equal(usage, 'must include --stdin or -u, --user');
   });
 
-  it('prints error on invalid JSON from stdin', done => {
```
DrMegavolt: why it is removed ?

test/services/credential-service/credentials.test.js
```diff
@@ -251,27 +250,6 @@ describe('Credential service tests', () => {
         .be.rejectedWith('one or more scopes don\'t exist');
     });
 
-    it('should use default property if not defined', () => {
```
DrMegavolt: why it is removed ?

altsang:
LOL - nice

XVincentX:
Outstanding parts

- [ ] Admin API payload validation

- [ ] Avoid continuous schema dereferencing

- [ ] Code review comments

- [ ] Self review the whole solution

lib/services/consumers/application.service.js
XVincentX: `app` is not a Promise, it's a simple value being passed to the next function in the chain.

This is the usual way to pass values down to a promise chain. This improves the locality of the task because you do not have to rely on global values spread in the test file.

So basically with this trick you can do this

```javascript
getUser('Vincenzo')
  .then((user)=>Promise.all([user, getCredentials(user)]))
  .then(([user, credentials]) => {/*…stuff…*/})
```

instead of 

```javascript
let theUser = undefined;

getUser('Vincenzo')
  .then((user)=>{ theUser = user; return getCredentials(user)})
  .then((credentials) => {/*…stuff…*/})
```

It is basically the Promise version of [async.waterfall](https://caolan.github.io/async/docs.html#waterfall)

Does that clarify?

test/cli/apps/create.test.js
XVincentX: I can put this back if you want, but basically this is testing something that now is handled by JSON Schema library (ajv) — and it's guaranteed to work. Previously it made sense, because it was handled with custom code, but now I think this can be safely removed.

In particular, an invalid JSON will _never_ pass any schema validation. So you'll get an error, no doubts on that.

test/services/credential-service/credentials.test.js
XVincentX: I can put this back if you want, but basically this is testing something that now is handled by JSON Schema library (ajv) — and it's guaranteed to work. Previously it made sense, because it was handled with custom code, but now I think this can be safely removed.

In particular, default values are automatically filled by ajv thanks to the JSON Schema definition. So this test will pass, always.

XVincentX:
> And do you still need to copy models to different config folders ?
I was thinking if EG can load default models
so the config folder will contain only 2 files.
and only if /models/... is present it will override default model

So your idea would be to bundle the models directly in the gateway and eventually specify those as an external file only if you need to override them, right?

I generally like the idea, although we should decide how to handle the override.

Let's say I want to use custom models. This means I need to create a `models` directory and copy the model I want to modify. Where do I find the original model? I'd need to dig into the gateway source code and manually copy-paste it there, and then start making changes. It's not a big deal, I just want to give you guys an heads up on that.

//cc @kevinswiber

test/services/users.test.js
XVincentX: I am not that opinionated on the style — both are ok. 

However I prefer `should(object).have.properties(…)` over `object.should.have.properties` is because the former is more explicit than the latter.

`should()` is an explicit function reference that has been imported on the top of the file.
`const should = require('should');`

On the other hand, the `should` property on tested objects is implicitly injected into the object. This feels like magic.

lib/rest/routes/apps.js
XVincentX: You're right, it might make sense. I'll have a look to the code and see if I can move it.

lib/services/consumers/application.service.js
DrMegavolt: ok, I get it. dao.insert for some reason returns `true` instead of inserted app

can you check is it really have to return true\not-true or it can just return the app ?

test/cli/apps/create.test.js
DrMegavolt: what if I accidentally disable the code that does validation ? 
previous tests are positive, so I will never know that validation logic is missing, right ?

test/services/credential-service/credentials.test.js
DrMegavolt: same as above, no doubt it will work, but what if I remove ajv itself ? how do I know it was removed

lib/config/models/credentials.json
XVincentX: That's generally correct but incomplete. Here's the whole story:

By default, JWT payload is not predictable so, as you stated, we do not know where the scopes could be put.

On the other other hand, you can define a scope on a credential type (just like you would do on `basic-auth` or `key-auth`), so you can check gateway scopes (internal scopes?) on it. 

Therefore the idea would be:

1. You create a `jwt` credential type and you put some scopes on it
2. You configure your system of record (Auth0 or whatever) so that the credential id is put in the `sub` standard field of the JWT
3. The gateway [will recognize](https://github.com/ExpressGateway/express-gateway/blob/master/lib/policies/jwt/jwt.js#L31) the sub, look for the credential and check the scopes on it. 

Again — these are scoped coupled to the credential itself wrote in the gateway — it has nothing to do with token scopes that, again, we can't detect.

This check can be disabled using `checkCredentialExistence` parameter to false.

Does that answer your question? I could expand the glitch demo to show the whole story if necessary.

lib/rest/routes/apps.js
XVincentX: @DrMegavolt 

I've checked and you're totally right. These things should go in the services.

Would it be ok to create a new issue and work on this as a separate pull request? This one is already big and in order to move these function in the services there are some errands to do (such as function renaming) and I do not want to mess up the tests here.

If you agree, I'll create a new issue.

DrMegavolt: I'll be reviewing it all together. It is hard for me to keep in mind all my comments between PRs. If it is easier for you, np, just make it on top of this branch so it will look as mega massive PR. 


btw, what you will have is:
PR1 
apps.js -1 line +1 line (findUser)
usage of findUser everywhere  +N lines

PR2
rm findUser usage -N lines
revert app.js  -1 +1
copy logic for find user from rest interface to the service -X +X lines 

so for review it will be 
+ (1+N+X+1) -(1+N+X+1) 

If you do it in the same PR 
it will be only +X -X because you just move code

DrMegavolt:
CLI can have a flag to generate models. it can be properly documented. 
It is advanced feature, so it seems acceptable to have to read the docs

eg create gateway --with-models

lib/rest/routes/apps.js
XVincentX: I like the math you did!
All right, I'll just push one more commit to move these parts 👍

lib/config/models/credentials.json
DrMegavolt: thanks for explanation. 

please bring this topic to one of the next standups

I remember some another vortex @altsang @kevinswiber 
as result there is current implementation where scopes are attached to **credential** not to consumer

I personaly like when actual consumer is assigned with permissions
like IAM is doing that. and when you need different scope you create different app with credential 

and when you need some common set of permission\scopes you create role\group

In any case having both ways is quite confusing. So we probably should reenter this vortex

XVincentX:
@DrMegavolt I've moved the functions you mentioned in the respective services in 96783cacf071e76b1dbca47dfc6d6385e3ed0e2a

Hopefully we'll be able to refactor a bit the service soon. The function names are really confusing, I think we can cleanup something there.

lib/config/models/credentials.json
XVincentX: All right. I've setup a reminder on Slack so we recall to discuss the feature.

lib/rest/routes/credentials.js
XVincentX: This should now be fixed in 96783ca

test/services/users.test.js
DrMegavolt: ok, I have no preference, just asking

lib/rest/routes/credentials.js
XVincentX: Uh actually no. Thing is `findConsumer` is to find an `User` or an `Application`.

In which service should I put this? We do not have a Consumer service

lib/services/consumers/application.service.js
XVincentX: I think it should technically be possible.

The only thing I should figure out is why [this check](https://github.com/ExpressGateway/express-gateway/blob/master/lib/services/consumers/application.dao.js#L27) is being performed (and the result returned) on the DAO object once the object has been inserted and more importantly who's relying on it.

lib/rest/routes/credentials.js
DrMegavolt: id or username resolution -> user service 
id or app name -> app service 

user or app? 
isn't it the only place find consumer is needed ? if so I would keep it there

XVincentX: Yes, it' the only place. Let's keep it there for now then.

lib/services/consumers/application.service.js
DrMegavolt: be careful with `.then(res => res.every(val => val));` last time I was trying to remove it everything blew up  

up to you, I'm ok with refactoring it in another task

test/cli/apps/create.test.js
XVincentX: > what if I accidentally disable the code that does validation ?
In that case, I suppose, a lot of other tests will break.

I've followed the code paths done by these two tests and they're covered by  a lot of other tests.

You can also see that although I removed two tests, the code coverage stayed the same (actually I think it increased).

I will readd these tests, but I feel they're not adding any new value.

XVincentX:
@DrMegavolt I've restored the test I deleted in the previous commits.

Feel free to review it one more time. We're almost there!

lib/rest/routes/users.js
```diff
@@ -1,12 +1,14 @@
...
 const usersSrv = require('../../services').user;
 const logger = require('../../logger').admin;
 
+const userSchema = 'http://express-gateway.io/models/users.json';
+
 module.exports = function () {
   const router = express.Router();
 
-  router.post('/', function (req, res, next) {
+  router.post('/', validateSchema(userSchema), function (req, res, next) {
```
DrMegavolt: what if I say that this validation should be as part of service ? 

let say you want to define another admin API endpoint\CLI through extension 
that somehow will create a user. (Let say some import endpoint) 

you will have to replicate the validation logic 
if it is part of the service it will be centralized

DrMegavolt:
I have one comment on placement of validation logic 

and what about having no models in config folder by default ? (I'm ok to have it in another PR, but I believe it will simplify experience)

other than that everything looks great

XVincentX:
> and what about having no models in config folder by default ? (I'm ok to have it in another PR, but I believe it will simplify experience)

Let's make it as a separate task.

lib/rest/routes/users.js
XVincentX: Validation is happening as well as part of the service. This validation actually is redundant; I put it here as well so if the payload is invalid we can skip the whole lookup in the database and all that stuff. Does that make sense?

DrMegavolt: but this means you have duplicated logic. and you are creating user, so the service should like 
-- validate schema or fail
-- check for dupes or fail
...

so it should not be a db call at all, right ?

XVincentX: That's correct. I think we can remove this layer. Let me do that right now.

XVincentX: Removed in a8686d2

Thanks for noticing this, now that I think about it it really does not make any sense to have it there as well 👍

XVincentX:
Note to self: a validation error should not be returned as a `500` status code, but as 415 (not processable entity)

DrMegavolt:
did you mean 422 ?

XVincentX:
Yeah probably. It's another issue though, so I'll handle it in a separate PR.

PR #587, opened 2017-12-20
DRY Config file and use a single watcher

### LINKED ISSUE #560: Hanging and mandatory watchers in the Config class
According to the `Config` class source code, the instantiated watchers are never closed.
This can be harmful for tests (as it will keep more and more handlers as we run tests).

Furthermore, the `watch` function will fail if the paths aren't provided (inline instantiated gateway). This forces us to run the whole test suite with the disabled watch flag [like here](https://github.com/ExpressGateway/express-gateway/blob/master/package.json#L39).

This means that some tests will fail locally if I forget to put that env variable (basically, always).

This is also related with #422

XVincentX:
Connect #560 
Closes #560

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -1,5 +1,5 @@
-const fs = require('fs');
 require('util.promisify/shim')(); // NOTE: shim for native node 8.0 uril.promisify
+const fs = require('fs');
 const util = require('util');
 const readFile = util.promisify(fs.readFile);
 const writeFile = util.promisify(fs.writeFile);
@@ -10,6 +10,7 @@ const chokidar = require('chokidar');
 const glob = require('glob');
 const yamlOrJson = require('js-yaml');
 const eventBus = require('../eventBus');
+
 class Config {
   constructor () {
     this.gatewayConfig = null;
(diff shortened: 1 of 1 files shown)


DrMegavolt:
I see that code became more complicated (`loadConfig`) and cannot understand how it helps ? 
having 2 watcher or one assuming system.config is not hot reloadable, what is the difference ? 

we can even remove watcher for system.config and it will be ok

XVincentX:
@DrMegavolt so — if I understand correctly — are you saying that `system.config.yml` does not support hot-reload?

DrMegavolt:
system config does not support hot reload 

in my original implementation it was only gateway config, then someone refactored it and add system.config for symmetry, but it never worked as hot reload 

technically it is possible, but a lot of things need to be considered. 
1) there are db settings (no need for hot reload)
2) some oauth2 settings that will migrate to plugins, so no need for hot reload either

3) and then the `plugins` section
It is possible to force re require on plugin package, but it may still leave subscriptions to event bus and http servers. 
and what if plugin adds new services? will it work ? 

Ideally I would approach this code only if there is an understanding of how new version should work
In current version watcher just sitting always idle, I don't see any problem with that

XVincentX:
Yeah ok it makes sense. Let's re-visit it when the time will come.

PR #592, opened 2018-01-02
Provide JSON Schema for configuration files

### LINKED ISSUE #415: Express-Gateway should fail to boot if a policy is missing an action
The following (wrong) configuration file for the gateway is making eg boot correctly, although the defined policy is missing an action (which should be mandatory)

```yaml
http:
  port: 8080
admin:
  port: 9876
  hostname: localhost
apiEndpoints:
  api:
    host: '*'
serviceEndpoints:
  httpbin:
    url: 'https://httpbin.org'
policies:
  - proxy
  - rewrite
pipelines:
  - name: basic
    apiEndpoints:
      - api
    policies:
      - rewrite:
          - condition:
              name: 'match'
              match: '/calendar/:test/hello'
          - action:
              rewrite: 'hello/:test/calendar'

      - proxy:
          - action:
              serviceEndpoint: httpbin
              changeOrigin: true
```

### LINKED ISSUE #575: Provide a JSON schema for configuration files
`system.config.yml` *and* `gateway.config.yml` could have a validation phase using JSON Schema. This would allow to skip a lot of unvalid configurations we would have to check manually such as `validateGatewayConfig` function in the gateway.

XVincentX:
Connect #575 
Closes #575
Closes #415

This pull request will introduce JSON Schemas for configuration files. This should prevent a lot of troubles when trying to start the gateway.

Unfortunately, it is not complete and it is not possible to make it so complete, as we have some hermaphrodite parts that will have to be solved first. It's a good first step, though.


DrMegavolt:
I saw the comment that `validateGatewayConfig` cannot be done . then why we need this logic that does the same with reduced functionality ? 

validateGatewayConfig has mixed functionality. some code is validating. some setting defaults
some allows different format
validating and defaults can be moved to the schema

XVincentX:
> validateGatewayConfig has mixed functionality. some code is validating. some setting defaults
some allows different format validating and defaults can be moved to the schema

I've moved everything I could from that two functions. There's no more validation or default values in the functions now.

What is left is simply [these single object to array transformations fallbacks](https://github.com/ExpressGateway/express-gateway/blob/feat/config-json-schema/lib/gateway/pipelines.js#L164-L166) and some [runtime checks with regards of other parts of the gateway](https://github.com/ExpressGateway/express-gateway/blob/feat/config-json-schema/lib/gateway/pipelines.js#L208-L210) that a JSON Schema cannot do.

I'd love to get rid of these but that would be a breaking change — so we can't make it now I guess.

>  then why we need this logic that does the same with reduced functionality?

JSON Schema is doing way more than the current logic we have in place. It's checking required parameters, their types, eventually dependencies and so on. We gain a lot more than what we have.

DrMegavolt:
do you need https://github.com/ExpressGateway/express-gateway/blob/feat/config-json-schema/lib/gateway/pipelines.js#L158-L160 ? 

and after that validateGatewayConfig is actually a `normalizeGatewayConfig`

DrMegavolt:
and for policiesConfig.indexOf(policyNames[0])

isn't it just a enum validation ?  
so name can be one of registered policy.

lib/gateway/pipelines.js
```diff
@@ -237,10 +205,6 @@ function validatePipelinePolicies (policies, config) {
   policies.forEach(policyObj => {
     const policyNames = Object.keys(policyObj);
 
-    if (policyNames.length !== 1) {
-      throw new Error('there should be one and only one policy per policy-object in pipeline configuration');
-    }
-
     if (policiesConfig.indexOf(policyNames[0]) === -1) {
```
DrMegavolt: and pls replace this part with schema

XVincentX: I am not completely sure whether this is a possible scenario.

Basically you're asking to validate that field based on an enum whose possible fields aren't part of the schema, but part of the payload itself. (Not sure if clear).

I'll try to research to see if it's possible, but at first glance I do not think it is.

DrMegavolt: ok, we can skip this part
$data seems to be not ready yet https://github.com/json-schema-org/json-schema-spec/issues/331

lib/config/schemas/gateway.config.json
```diff
@@ -0,0 +1,108 @@
...
+        "object",
+        "null"
+      ],
+      "additionalProperties": {
+        "type": "object",
+        "properties": {
+          "host": {
+            "type": "string"
+          },
+          "paths": {
```
DrMegavolt: pathRegex is also valid

lib/config/schemas/gateway.config.json
```diff
@@ -0,0 +1,108 @@
...
+        "properties": {
+          "apiEndpoints": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          },
+          "policies": {
+            "type": "array",
+            "items": {}
```
DrMegavolt: is there a way to describe wildcard policy name and then array of condition\action pairs ?

lib/config/schemas/system.config.json
```diff
@@ -0,0 +1,107 @@
...
+          "type": "string",
+          "default": "aes256"
+        },
+        "saltRounds": {
+          "type": "number",
+          "default": 10
+        }
+      },
+      "required": [
+        "cipherKey",
```
DrMegavolt: if you have defaults, are they required ?

DrMegavolt: And does it mean that we can comment out those values in system.config ?

XVincentX:
> do you need https://github.com/ExpressGateway/express-gateway/blob/feat/config-json schema/lib/gateway/pipelines.js#L158-L160 ?

Right, I've removed it.

> and after that validateGatewayConfig is actually a normalizeGatewayConfig without true\false magic in code

Regarding the name, you're perfectly right. I've renamed the function in 5b3baec.

Regarding return `true/false`, that value is not indicating whether the validation passed or not (and so starting or not the gateway). 

That value is indicating whether continuing the processing the configuration file or not. The gateway will start anyway.

Does that clarify?

lib/gateway/pipelines.js
```diff
@@ -186,7 +154,7 @@ function configurePipeline (pipelinePoliciesConfig, config) {
   return router;
 }
 
-function validateGatewayConfig (config) {
+function normalizeGatewayConfig (config) {
```
DrMegavolt: can schema require gatewayConfig.pipelines or  gatewayConfig.pipeline ?
 
I just don't understand the reason to leave true\false login

DrMegavolt:
@XVincentX for true/false 
I can leave with it for now.

please check comments in schema and we should be good to go

lib/config/config.js
```diff
@@ -89,14 +111,14 @@ class Config {
       systemWatcher.on(watchEvent, name => {
         log.info(`${watchEvent} event on ${name} file. Reloading system config ` +
           `file from ${this.systemConfigPath}`);
-        this.loadSystemConfig();
+        try { this.loadSystemConfig(); } catch (e) { }
```
DrMegavolt: can we add some log.debug here ?

lib/config/schemas/gateway.config.json
XVincentX: Yes. I'll modify the schema so policies gets typed. 👍

lib/config/schemas/system.config.json
XVincentX: > if you have defaults, are they required?

Well I'd say yes. They're required fields, but we'll provide good defaults in case you do not provide a value. This is useful when somebody is giving an invalid payload such as 

```javascript
const object = { field: undefined }
``` 

In this case, defaults won't happen (because the field has been provided) but the `required` constraint will mark it as invalid.

> And does it mean that we can comment out those values in system.config ?

Yes. If you have all the default values in place you can technically run the gateway without any config file at all.

lib/config/config.js
XVincentX: Sure!

lib/config/schemas/gateway.config.json
XVincentX: Actually I recall why I did not define this: as condition/action is in an array:
```
{[condition: {}, action: {}]}
```

this can be a bit tricky to accomplish. I'll see what I can do here.

XVincentX:
@DrMegavolt Done!

lib/config/schemas/gateway.config.json
DrMegavolt: cool, I see pathRegex, 
I also remember `verbs` (or should we call it methods ) so we can skip it for now

lib/config/schemas/gateway.config.json
```diff
@@ -0,0 +1,137 @@
...
+          "type": "string"
+        }
+      }
+    },
+    "apiEndpoints": {
+      "type": [
+        "object",
+        "null"
+      ],
+      "additionalProperties": {
```
DrMegavolt: what about scopes ?

lib/config/schemas/system.config.json
DrMegavolt: aha, now I understand about undefined. 

So can we update system config (in `lib` and in CLI templates) to comment out all values that have defaults ?
So it will look like standard apache\nginx\squid or whatever config files. A lot of settings are there with brief explanation,  but they all are disabled

DrMegavolt:
Almost there 
see comment Api Endpoint scopes 
and I would comment out all settings that have defaults in config files

I understand that it cannot be default for http/admin port. because no value means disable. for system config it seems to be all defaults

and especially for db section. the config file can contain redis example, so users will not have to go the doc site

lib/config/schemas/system.config.json
XVincentX: You mean like provide the config file with exact the default values but everything commented out, right?

That might sound a good idea. @kevinswiber ?

1.6? Separate release?

lib/config/schemas/gateway.config.json
XVincentX: Added in 7fdb0b3

XVincentX:
> and I would comment out all settings that have defaults in config files

I think it's a good idea. My only concern is whether it's a good idea to put it as part of 1.6 release. This might be a bit hard to swallow for the guys.

My opinion would be to keep the things a little bit more explicit now, and then in 2.0, when we can finally break some things, we can revisit multiple things, such as default config.

@kevinswiber What's your idea?

lib/config/schemas/gateway.config.json
XVincentX: It's `methods` and you said it's undocumented. Do you want me to add it?

lib/config/schemas/system.config.json
kevinswiber: Let's not block this particular PR waiting for the detailed comments.  Make an issue for it, though, and we can try to get it done for 1.6.

lib/config/schemas/gateway.config.json
```diff
@@ -0,0 +1,152 @@
...
+    },
+    "https": {
+      "type": "object",
+      "properties": {
+        "port": {
+          "type": "number"
+        },
+        "tls": {
+          "type": "object",
+          "additioanlProperties": {
```
kevinswiber: `additioanlProperties` => `additionalProperties`

lib/config/schemas/system.config.json
```diff
@@ -0,0 +1,107 @@
...
+        "algorithm",
+        "saltRounds"
+      ]
+    },
+    "session": {
+      "type": "object"
```
kevinswiber: Can we get more specific here?  We know the set of properties in here, yeah?

DrMegavolt: as part of implementing session providers it can be that the only property we know is name of the provider

DrMegavolt: I suggest to leave it as is

lib/config/schemas/gateway.config.json
kevinswiber: Why is `methods` undocumented?

XVincentX: I don't know guys, I wasn't part of the discussion.

lib/config/schemas/gateway.config.json
XVincentX: 😱 
Fixed in 8617d37

lib/config/schemas/system.config.json
XVincentX: @kevinswiber List of options is [here](https://github.com/expressjs/session#options)

I didn't enumerate them because our config file is listing only some of these and I wanted to understand more.

I'm fine enumerating all of them eventually.

Issue #586, opened 2017-12-20
Achieve compatibility with Node 9 by fixing the tests


kevinswiber:

XVincentX:
The problem seems to be related to the way we're executing the test, but the gateway itself seems to be working correctly.

XVincentX:
Closed with #596

PR #596, opened 2018-01-03
Make the gateway tests run on Node 9 correctly

DrMegavolt:


XVincentX:
Connect #586 
Closes #586

XVincentX:
The reason why tests are failing is because of the way we spin up gateway instances for e2e tests.
I do not properly remember the thing, but I'll investigate again and fix that hopefully.

XVincentX:
This PR will ensure Node 9 compatibility with the following things:

* Node is slowly moving to [static error codes](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V9.md#9.0.0), so the string errors will slowly disappear: a4a85cc

* The E2E tests were initiating a whole new gateway instance from scratch, reinstalling all the dependencies, *BUT* at the end the new installed gateway is not used at all. It was probably just a way to initialize a config directory with the required data. This was cause of flakyness and errors in node 9 ce5da59 and 197a0fa

* Removed a stale Docker file and modified the current one so that's relying on `node8-alpine` instead to the latest `alpine`

XVincentX:
@DrMegavolt I can't assign you as a reviewer since you opened the PR.

DrMegavolt:
@XVincentX LGTM

PR #598, opened 2018-01-04
Provide migration scrips for old models

### LINKED ISSUE #594: Provide a migration script for old models
Provide a migration script to move old models to the new ones.

XVincentX:
Connect #594 
Closes #594 

This PR will provide a migration script for old models, looping through them and converting them to a JSON Schema equivalent one.

I found one issue on the way: https://github.com/tj/node-migrate/issues/96; I had to make a small trick to make it work properly.

P.S: Legacy is hard.

### DIFF
--- a/lib/config/schemas/system.config.json
+++ b/lib/config/schemas/system.config.json
@@ -10,7 +10,7 @@
           "properties": {
             "emulate": {
               "type": "boolean",
-              "default": "true"
+              "default": false
             }
           },
           "required": [
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -7,7 +7,6 @@ const YAWN = require('yawn-yaml/cjs');
 const path = require('path');
 const log = require('../logger').config;
 const chokidar = require('chokidar');
-const glob = require('glob');
 const yamlOrJson = require('js-yaml');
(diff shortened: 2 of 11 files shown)


lib/config/schemas/system.config.json
```diff
@@ -10,7 +10,7 @@
           "properties": {
             "emulate": {
               "type": "boolean",
-              "default": "true"
+              "default": false
```
DrMegavolt: why ?

XVincentX: `"true"` is a string, not a boolean. `ajv` is not configured to coerce values.

Also, I had to change the default value from `true` to `false` because if not, the following scenario would not work:


```yml
db:
  redis:
    host: redis
    port: 6379
    namespace: EG
```

Because `emulate` would default to `true`, thous skipping my connection parameters.

DrMegavolt: ah, ok. As long as default templates have emulate:true set explicitly it is ok

XVincentX: Yes, they have it. I've checked them.

XVincentX:
I'll merge this and based on the discussion around the document I posted, will move from there.

Issue #573, opened 2017-12-12
session is created for auth policies even if it is set as session:false


DrMegavolt:
issue by @pbhadauria2000

line 157 of lib/policies/oauth2/oauth2.js calls out session as false but that is ignored as line 20 of lib/gateway/actionParams.js is not passing the options to logIn function and passport ends up calling sessionmanager to create session.

this is because 
https://github.com/jaredhanson/passport/blob/master/lib/http/request.js#L42

is creating session by default

PR #600, opened 2018-01-08
fix for options.session was not passed in req.login function connect #573

DrMegavolt:


XVincentX:
@DrMegavolt Is this PR ready for review or are you still working on this?

DrMegavolt:
@XVincentX this and #601 are ready.

XVincentX:
Connect #573 
Closes #573

lib/gateway/actionParams.js
```diff
@@ -17,7 +17,7 @@ ActionParams.prototype.getCommonAuthCallback = function (req, res, next) {
       return next(err);
     }
     if (user) {
-      req.logIn(user, (err) => next(err));
+      req.logIn(user, this, (err) => next(err));
```
XVincentX: Can we just pass `{session: this.session}` so we do not pass the whole object?

Now it is just that parameter, but just in case...

DrMegavolt: passport has it as object. so I assume they can at some point add other params. in this case passing only session will require code modification. in current version any new params is passed.

PR #601, opened 2018-01-08
Feature/session providers connect #597 closes #597 

### LINKED ISSUE #597: Allow plug-in another session storage for oauth2 module
oauth2 auth code grant requires session. 
currently it is implemented with default in memory express-session storage. 
that is configured in system.config

EG should allow customization of storage type it is using through config
```
session:
  type: connect-redis 
  other-options-expected-by-connect-redis: ...
```

DrMegavolt:
based on #600 

docs are in eg_io branch

### DIFF
--- a/lib/policies/oauth2/oauth2-routes.js
+++ b/lib/policies/oauth2/oauth2-routes.js
@@ -13,6 +13,8 @@ module.exports = function (app, config) {
     try {
       const ProviderStore = require(config.systemConfig.session.storeProvider)(session);
       config.systemConfig.session.store = new ProviderStore(config.systemConfig.session.storeOptions);
+      delete config.systemConfig.session.storeProvider;
+      delete config.systemConfig.session.storeOptions;
     } catch (error) {
       logger.error(`Failed to initialize custom express-session store, please ensure you have ${config.systemConfig.session.storeProvider} npm package installed`);
       throw error;
--- a/lib/policies/oauth2/oauth2-routes.js
+++ b/lib/policies/oauth2/oauth2-routes.js
@@ -5,9 +5,19 @@ const express = require('express');
 const passport = require('passport');
 const site = require('./site');
 const oauth2Server = require('./oauth2-server');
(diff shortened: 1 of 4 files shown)


lib/policies/oauth2/oauth2-routes.js
```diff
@@ -5,9 +5,19 @@ const express = require('express');
...
 const site = require('./site');
 const oauth2Server = require('./oauth2-server');
+const logger = require('../../logger').policy;
 
 module.exports = function (app, config) {
   app.set('view engine', 'ejs');
+  if (config.systemConfig.session.storeProvider) {
+    try {
+      const ProviderStore = require(config.systemConfig.session.storeProvider)(session);
+      config.systemConfig.session.store = new ProviderStore(config.systemConfig.session.storeOptions);
```
XVincentX: If possible, I would then delete `storeProvider` and `storeOptions` once the custom provider has been created, do we do not pass this stuff down in the `express-session` middleware.

test/common/session-provider.js
```diff
@@ -0,0 +1,19 @@
+let data;
```
XVincentX: Can we use [Sinon mocks](http://sinonjs.org/releases/v2.1.0/mocks/) instead of writing our own code?
It's way easier to inspect the calls and we can even be more precise if required.

DrMegavolt: the trick is that it needs to be required. So I cannot just mock. 

so for verifying data I would have to export one of sinon stubs. 
Doesn't really save lines of code. and will increase complexity

`module.exports.getOptions`
will have code like 

stub.getCalls(0).args[0]

do you insist on sinon approach ?

XVincentX: Ah I see, you're right. In that case it does not make sense effectively.

We could use `proxyquire` eventually, but it's fine for now.

DrMegavolt: there was `mock-require` lib, you removed it ))

PR #602, opened 2018-01-09
Update all possible packages

### LINKED ISSUE #582: Try to update all the possible dependencies
You know, newer is better.

XVincentX:
Connect #582 
Closes #582


package.json
```diff
@@ -37,9 +37,9 @@
     "pretest": "eslint .",
     "precommit": "lint-staged",
     "test": "npm run mocha:istanbul",
-    "test:all": "cross-env EG_CONFIG_DIR=test/config EG_DISABLE_CONFIG_WATCH=true mocha --recursive test --timeout 60000",
-    "test:unit": "cross-env EG_CONFIG_DIR=test/config EG_DISABLE_CONFIG_WATCH=true mocha --recursive \"./test/{,!(e2e)/**/}*.test.js\" --timeout 5000",
-    "test:e2e": "mocha --recursive test/e2e --timeout 60000",
+    "test:all": "cross-env EG_CONFIG_DIR=test/config EG_DISABLE_CONFIG_WATCH=true mocha --recursive test --timeout 60000 --exit",
```
DrMegavolt: and do we have something that prevents stoping ?

XVincentX: Unfortunately yes. Every time we do not clean up the resources properly (leaving file handles or sockets) it will hang.

> To avoid false positives and encourage better testing practices, Mocha will no longer automatically kill itself via process.exit() when it thinks it should be done running.

> If the mocha process is still alive after your tests seem "done", then your tests have scheduled something to happen (asynchronously) and haven't cleaned up after themselves properly. Did you leave a socket open?

That's why I was so insisting in calling `done` and all that stuff around.

> Supply the --exit flag to use pre-v4 behavior.

If this gets merged, I'll open an issue and see if I can clean up all the resources. [This package](https://www.npmjs.com/package/why-is-node-running) should help me during the process.

Issue #579, opened 2017-12-14
Tests aren't cleaning up the resorces


XVincentX:
Our test suite has multiple parts where async cleanup resources aren't waited for its completion. This introduces flakyness as well as hardening the things when you want to DRY the things.

[Example](https://github.com/ExpressGateway/express-gateway/blob/master/test/common/routing.helper.js#L50): `close` method of express applications are async, and they accept a `done` parameter.

XVincentX:
With mocha 4.0 this is now kind of a priority. Mocha now does not exit anymore in case there are pending handles (such as sockets, file descriptors and shit like that).

Therefore, we might really need to clean this up, one per time.

PR #604, opened 2018-01-10
Remove chai in favor of should

### LINKED ISSUE #535: Unify the assertion library
It turns out we're using a mix of `chai`, `should.js` and the built-in `assert` to make Assertions in our tests.

Decide which one to use and go through the whole codebase to make it happen.

XVincentX:
Connect #535 
Closes #535 

# WIP

### DIFF
--- a/test/common/routing.helper.js
+++ b/test/common/routing.helper.js
@@ -1,6 +1,5 @@
-// const path = require('path');
 const request = require('supertest');
-const assert = require('chai').assert;
+const should = require('should');
 const logger = require('../../lib/logger').test;
 const gateway = require('../../lib/gateway');
 const config = require('../../lib/config');
@@ -27,12 +26,12 @@ module.exports = function () {
   }
   return {
     addPolicy: (name, handler) => { // TODO: make it plugin
-      policies.register({policy: handler, name});
+      policies.register({ policy: handler, name });
     },
-    setup: ({config, plugins} = {}) => {
+    setup: ({ config, plugins } = {}) => {
       originalPolicies = policies;
 
-      return gateway({config, plugins})
+      return gateway({ config, plugins })
         .then(apps => {
           app = apps.app;
           httpsApp = apps.httpsApp;
@@ -96,14 +95,14 @@ module.exports = function () {
           .expect('Content-Type', /json/)
           .expect((res) => {
             if (testCase.test.result) {
-              assert.equal(res.body.result, testCase.test.result);
+              should(res.body.result).be.eql(testCase.test.result);
             }
-            assert.equal(res.body.url, testCase.test.url);
+            should(res.body.url).be.eql(testCase.test.url);
             if (testCase.test.host) {
-              assert.equal(res.body.hostname, testCase.test.host);
+              should(res.body.hostname).be.eql(testCase.test.host);
             }
             if (testCase.test.scopes) {
-              assert.deepEqual(res.body.apiEndpoint.scopes, testCase.test.scopes);
+              should(res.body.apiEndpoint.scopes).be.deepEqual(testCase.test.scopes);
             }
           })
           .end((err, res) => {
@@ -117,7 +116,7 @@ module.exports = function () {
         const testScenario = prepareScenario(testCase);
         testScenario
           .expect((res) => {
-            assert.deepEqual(res.body.params, testCase.test.params);
+            should(res.body.params).be.deepEqual(testCase.test.params);
           })
           .end((err, res) => {
             if (err) { logger.error(res.body); }
--- a/package-lock.json
+++ b/package-lock.json
@@ -279,12 +279,6 @@
       "resolved": "https://registry.npmjs.org/assert-plus/-/assert-plus-1.0.0.tgz",
       "integrity": "sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU="
     },
-    "assertion-error": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/assertion-error/-/assertion-error-1.1.0.tgz",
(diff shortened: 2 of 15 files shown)


test/common/routing.helper.js
```diff
@@ -1,6 +1,5 @@
-// const path = require('path');
 const request = require('supertest');
-const assert = require('chai').assert;
```
DrMegavolt: why not replace with require('assert') ? it will be very little modification for source code in this case 

 and in general 
`should(found.schema).not.be.undefined();`
how this is better from `assert(found.schema)`  ?

XVincentX: Good question. So the reason why I am trying to go in this way is:

1. `should`, at the moment, is way more (in our codebase) used than `assert` and `chai.assert`; so using should everywhere requires way less changes than pursuing the opposite way.

2. `require('assert')` has way less assertions than chai and should have. For example, it lacks any Promise support, Throws assertions and a lot of utilities that we're using (and I think they're useful as they improve the readability).

DrMegavolt: https://nodejs.org/api/assert.html
assert.throws is there. and promise support is not needed. typically you just return promise
deepEqual etc.

`it('should bla' , ()=> myPromise)`

For all kind of requests, `supertest` typically have you covered  

For example express.js is happy with assert 
https://github.com/expressjs/express/search?utf8=✓&q=assert&type=

they have some should usage in cases like   `express.Router.should.be.a.Function()`
and this specific usage is  better than `assert(typeof express.Router === "function")` 

From contributor perspective you need to learn yet another lib to write a test

I'm ok removing `chai`. I would leave `should` as an option to write some tests, it sometimes indeed have nice helpers like equivalent, type testing etc. But I don't see any reason rewriting assert tests to should. 

@kevinswiber we may need some voting here

XVincentX: I'm not that opinionated on testing frameworks, so either way is ok to me. However my rationale behind using one test library everywhere (and `should` instead of `assert`) is:

* You have one library, and you have one thing to learn. It think it would be a bit confusing to see some parts with `assert`, some with `should`, some with `supertest`. Whether we want to go with one or another, I'd at least make sure that it'd be just one. From this perspective, I'd also try to get rid of `supertest` and replace it with an HttpClient + the assertion library we convey on.

* While it is true that with mocha you can just return a Promise, the problem is that some time you want to make sure the Promise has been rejected. While with `should` you can just do `should(p).be.rejected()`, with mocha you would have to kind of "inverse" the flow and return a fulfilled promise when the selected one has been rejected, and viceversa.

* Error messages and tests readability. `should` contains a lot more utilities that make the tests easier to write, but more importantly to read and parse. Let's consider this example:

```javascript
const x = 2;

should(x).be.above(2);
// Output: AssertionError: expected 2 to be above 2

assert(x > 2);
// Output: AssertionError [ERR_ASSERTION]: false == true
```

You can see here that while Assert is telling `false==true` (that is effectively what happened). Now go figure out what's going on in the source code.

It's *way* easier to understand what's going on when you have specific assertions such as `above` and so on.

With promises also (rejected, fulfilled, that stuff) I think the value of such messages that is increasing.

P.S: [jest](https://facebook.github.io/jest) would be the solution to all of that, but we can defer the discussion on that when tests will clean up the resources properly. #579 

@kevinswiber Please join the party 😄, and tell us your thoughts.

kevinswiber: I'm not very opinionated here.  I typically go minimal and use Node's core `assert` module. +1 to removing Chai.   I think the answer here is very simple.  

Does the test file use anything special from `should`?  Okay, then... use `should`.

Is the test file only using `assert` compatible assertions?  Okay, then... use `assert`.

If we need more features in the future, we can just move the test file over to using `should`.

XVincentX: @kevinswiber 

> Is the test file only using `assert` compatible assertions? Okay, then... use `assert`.

The problem here is that ultimately every assertion can be expressed using `assert`.

```javascript
should(object).have.properties(['a','b'])

assert(object.a);
assert(object.b);
```

…and the list goes on and can go to "the infinite and beyond"

With such "flexibility" (and I am mean it, exclusively in this context, in a bad way unfortunately) every usage of `should` or `assert` becomes debatable. 

I am already imagining code reviews where you have

```javascript
assert(a > 0);
```

where somebody (probably me, 😈 ) would say:

> Can you replace that assertion with `should(a).be.above(0)`?

And the answer could probably be

> assert is good enough here.

Depending on your mood, how is your day, or even the laziness level.

Do not get me wrong — the answer is effectively true, but modern testing should make sure tests have enough expressivity and a semantic meaning — thing that built in `assert` can't assure. We need decent stack traces and meaningful error messages if something goes wrong. I've already paid this price while fixing things for the identity part.

Testing in Javascript is a pain, especially when compared to [other languages](http://honzajavorek.cz/blog/testing-in-javascript-and-in-python), but the things are advancing and we should use these. Jest or Ava might be the answer to that, and I feel we should have the codebase prepared for such change, when that'll happen. Assert won't help us towards that direction.

kevinswiber: > The problem here is that ultimately every assertion can be expressed using assert.

Doesn't sound like a problem to me.  Sounds like a solution.  :)

Look, I'm not sure we benefit from being heavy-handed on this.  If an assertion is better expressed through `should`, that's just great.  If a PR comes in with an assertion that uses `assert`, and we can't clearly identify what it's testing, then we should offer a better suggestion, perhaps using `should`.

Our first goal with tests should be correctness.  Our second goal should be clarity.  

I'm totally fine merging these changes.  Merging these changes does not necessarily draw a line in the sand that we'll only accept `should` assertions from now on.  Happy to be consistent where it makes sense and as much as possible.  Let's move on and stop beating this horse.

XVincentX: Please approve then. Thanks!

PR #606, opened 2018-01-11
restore docker file that can be used to build images from source code

DrMegavolt:
Have no idea why the file was removed from source code


XVincentX:
I think I deleted it by mistake.

PR #607, opened 2018-01-11
Autorun the migration script for JSON Schema

XVincentX:
Connect #594 

# WIP

I'm still looking for other edge cases, in meantime you can start to review it.

### DIFF
--- a/Dockerfile-master
+++ b/Dockerfile-master
@@ -1,11 +1,11 @@
-FROM node:alpine
+FROM node:8-alpine
 
 ARG VERSION
 ARG TYPE=basic
 
 ENV NODE_ENV production
 # this will enable polling, hot-reload will work on docker or network volumes
-ENV CHOKIDAR_USEPOLLING true 
+ENV CHOKIDAR_USEPOLLING true
 
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
@@ -17,8 +17,8 @@ COPY . /usr/src/app
 EXPOSE 8080
 
 # HTTPS
-# EXPOSE 8443 
+# EXPOSE 8443
 
 # Admin API
-EXPOSE 9876 
+EXPOSE 9876
 CMD [ "npm", "run", "start" ]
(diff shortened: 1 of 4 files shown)


Dockerfile-master
```diff
@@ -17,8 +17,8 @@ COPY . /usr/src/app
 EXPOSE 8080
 
 # HTTPS
-# EXPOSE 8443 
+# EXPOSE 8443
```
DrMegavolt: may be this as well ?

XVincentX: I'll copy paste my Gitter comment here:

> Regarding the Docker image — it is hard to make everybody happy, as the topic is so opinionated…

> So my standpoint has always been to not expose the Admin port by default. You can always do that while running the docker image -p or using docker-compose.yml

> On the other hand, if you expose the port then you have it opened, regardless if you plan to use it or not.

> So — while with the first approach you have a solution, with the second one you can't really do anything about it.

XVincentX:
@DrMegavolt @kevinswiber If you could try this branch on a current gateway installation you have and try it out, it'd be great.

Dockerfile-master
DrMegavolt: just thinking about opposite direction. 

what if no expose at all ? 

in any case you need to do `-p 9999:8080` to map container to host port to allow traffic 

EXPOSE is only for accessing it from another container inside docker 

the docs needs to be written

XVincentX: > what if no expose at all ?

I've been checking multiple popular images (nginx, redis, mysql, mongodb) and they're all exposing the main port; and I think it's fair enough: the purpose of these kind of containers (eg included) is to communicate with the other containers in the network.

The general philosophy I think should be: expose the **main** port, let the others expose the **side** ports.

> in any case you need to do -p 9999:8080 to map container to host port to allow traffic

This is not completely correct. I can also do `docker run -P` and it will automatically expose all the ports listed through `EXPOSE` on my host.

XVincentX: Also — note that this change is not introducing anything, it's just removing the white space 😱

DrMegavolt: ok, I'm happy with admin API exposed so some internal operations can happened https most likely will be public so -p will most likely will be used

PR #610, opened 2018-01-15
Properly shutdown some resources during tests

XVincentX:
This PR will make sure to return Promises on multiple tests when using `helper` function so all resources are properly cleaned up once done.

This won't solve the problem, but it will help the investigation. 

Connect #579

### DIFF
--- a/test/policies/basic-auth-policy.test.js
+++ b/test/policies/basic-auth-policy.test.js
@@ -95,11 +95,10 @@ describe('Functional Tests basic auth Policy', () => {
       });
   });
 
-  after('cleanup', (done) => {
+  after('cleanup', () => {
     app.close();
     config.gatewayConfig = originalGatewayConfig;
-    helper.cleanup();
-    done();
+    return helper.cleanup();
   });
 
   it('should not authenticate token for requests without token header', function () {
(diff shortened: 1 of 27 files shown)


test/policies/basic-auth-policy.test.js
```diff
@@ -72,34 +72,24 @@ describe('Functional Tests basic auth Policy', () => {
             should.exist(_fromDbUser1);
             user = _fromDbUser1;
 
-            credentialService.insertScopes('authorizedScope', 'unauthorizedScope')
+            return credentialService.insertScopes('authorizedScope', 'unauthorizedScope')
               .then(() => {
```
DrMegavolt: this should be extracted to the higher level. 
it should be 
f
.then
.then

not f.then(x=>x.then)

current syntax works for callback style, but once you start do `return` it is super easy to miss correct place to return

test/policies/basic-auth-policy.test.js
```diff
@@ -72,34 +72,24 @@ describe('Functional Tests basic auth Policy', () => {
...
                   .then((userRes) => {
                     should.exist(userRes);
                     return serverHelper.generateBackendServer(6067)
-                      .then(() => {
-                        helper.setup()
-                          .then(apps => {
-                            app = apps.app;
-                            done();
-                          });
+                      .then(helper.setup).then(apps => {
```
DrMegavolt: 👍 like here

test/policies/basic-auth-policy.test.js
XVincentX: Fixed in 2205b78

PR #612, opened 2018-01-16
Return 404 when no policy can handle a request

### LINKED ISSUE #611: Express-Gateway is crashing when a request can't be handled by any pipeline
Express Gateway is starting when is fed with the current configuration file

```yml
http:
  port: 8080
apiEndpoints:
  verticals:
    path: '/verticals'
  countries:
    path: '/countries'
  msm_standard_proxy:
    host: '*'
serviceEndpoints:
  msm_origin:
    url: 'https://httpbin.org/'
policies:
  - proxy
  - terminate
pipelines:
  common:
    apiEndpoints:
      - verticals
      - countries
    policies:
      - terminate:
          - action:
              statusCode: 200
              message: 'should stop here'
  verticals_pipeline:
    apiEndpoints:
      - verticals
    policies:
      - terminate:
          - action:
              statusCode: 500
              message: 'should NOT stop here'
```

and you try to hit an any route that is not in any pipeline.

[Test instance](https://glitch.com/edit/#!/express-gateway-test)

EG should probably return a 404 when no suitable no pipeline is matching the current request.

XVincentX:
Connect #611 
Closes #611 

This PR will make sure the gateway return a 404 in case there's no policy that can handle a request.

At the moment, the gateway is crashing.

### DIFF
--- a/lib/gateway/pipelines.js
+++ b/lib/gateway/pipelines.js
@@ -41,7 +41,13 @@ module.exports.bootstrap = function ({ app, config }) {
       }
       for (const path of mountPaths) {
         log.debug('mounting routes for apiEndpointName %s, mount %s', route.apiEndpointName, path);
+
         const handler = (req, res, next) => {
+          if (!apiEndpointToPipelineMap[route.apiEndpointName]) {
+            log.debug(`No suitable pipeline found for ${route.apiEndpointName}`);
+            return res.sendStatus(404);
+          }
+
           log.debug('executing pipeline for api %s, mounted at %s', route.apiEndpointName, path);
 
           req.egContext = Object.create(new EgContextBase());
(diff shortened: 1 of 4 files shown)


lib/gateway/pipelines.js
```diff
@@ -199,14 +192,35 @@ function normalizeGatewayConfig (config) {
...
     const policyNames = Object.keys(policyObj);
 
-    if (policiesConfig.indexOf(policyNames[0]) === -1) {
+    if (avaiablePolicies.indexOf(policyNames[0]) === -1) {
       throw new Error(`${policyNames[0]} policy not declared`);
     }
   });
 }
+
+const generatePipelineHandler = (path, pipeline, route) => {
```
DrMegavolt: may be object  `{path, pipeline, route}`  ?

test/pipelines/empty.test.js
```diff
@@ -0,0 +1,53 @@
+const supertest = require('supertest');
+const testHelper = require('../common/routing.helper');
+const config = require('../../lib/config');
+
+describe('Pipelines', () => {
+  describe('send a request that can\' be handled by any pipeline', () => {
```
DrMegavolt: 1) is it `can't` ? 
2) can we use "send a request that can't be handled by any pipeline"
so if you do a string search it will actually find a test ?

lib/gateway/pipelines.js
XVincentX: Done in 31b8497

test/pipelines/empty.test.js
XVincentX: Fixed.

XVincentX:
Do you think this could go in 1.6? It's scheduled for tomorrow.

lib/gateway/pipelines.js
```diff
@@ -199,14 +192,35 @@ function normalizeGatewayConfig (config) {
...
 
-    if (policiesConfig.indexOf(policyNames[0]) === -1) {
+    if (avaiablePolicies.indexOf(policyNames[0]) === -1) {
       throw new Error(`${policyNames[0]} policy not declared`);
     }
   });
 }
+
+const generatePipelineHandler = ({ path, pipeline, route }) => {
+  if (!pipeline) {
```
DrMegavolt: ok, so this is pretty much the 3 lines that do the work. I get it

PR #615, opened 2018-01-21
Fix paths schema

### LINKED ISSUE #614: BUG: Getting started template config error
[ BUG ] Getting started template has an error in gateway.config.yml

data.apiEndpoints['api'].paths should be array and currently isn't.

### LINKED ISSUE #616: Oauth2 related configs are required in system.config even if it is not in use
props like `crypto` `session` etc. are required after JSON schema validation was introduced. 
previously it was possible to use EG  with just `db` section set to emulate

XVincentX:
Connect #614 
Closes #614 

Connect #616 
Closes #616

### DIFF
--- a/lib/config/schemas/gateway.config.json
+++ b/lib/config/schemas/gateway.config.json
@@ -55,7 +55,10 @@
             "type": "string"
           },
           "paths": {
-            "type": "array",
+            "type": [
+              "string",
+              "array"
+            ],
             "items": {
               "type": "string"
             }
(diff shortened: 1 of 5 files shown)


XVincentX:
@DrMegavolt I've now provided empty default for such objects — they'll be then filled automagically with the default properties.

Please review it one more time — if ok I'll release.

DrMegavolt:
ok, assuming at some point one of the templates will have system config with just db section (or maybe even without it) then there is no need for special test for empty system config. 
LGTM, go 1.6.1

PR #620, opened 2018-01-25
Support for env variables in config file

### LINKED ISSUE #510: Env var support in config files: implementation
_As an EG user/developer, I'd like to be able to templatize the YAML files to be able to accept env variables driving config dynamically so that I can stand up multiple instances with different configuration._

### LINKED ISSUE #560: Hanging and mandatory watchers in the Config class

XVincentX:
#### This is still a wip…

This pull request will implement env variable support for configuration files (both system and gateway). The syntax and the usage is the same as the one provided from `docker-compose.yml` files.

### Syntax

`value: ${ENV_VAR:-DEFAULT}`

Eg will look for an environment variable called `ENV_VAR`; if provided, the replacement will happen. If not, it will be replaced with the default value.

#### Example

```yml
http:
  port: ${EG_HTTP_PORT:-#8080}
```

This change is happening in 3b6f743add625cb4708b5bf9b48a967b83f6d535 and 0b562e76038300f0be924bb888ccd114933df3c0 and we can just merge this commit if we want to make it happen quickly but I invite you to guys to have a look here as well:

## Config changes

While implementing this change I went again into the same problem of our configuration class: duplicated code. Yet another time I had to copy paste the code because we have two identical functions.

Given that's the third time, I wanted to stop this forever and ever.

* Removed `loadSystemConfig` and `loadGatewayConfig` in favor of `loadConfig(type)` function
* Removed the two watcher in favor of a single one
* Emit the `hot-reload` event only if the file has been correctly reloaded

## Tests changes

The env variable support is also what we needed to remove another copy of gateway and config files for testing purposes. We can now use `8080` for development and a random port for tests so they do not collide. This is happening in 81a89a1

## No tests

While I was writing the tests I realized we do not really need them, because the implemented behavior is tested by the way we're running the tests themselves (it's using `EG_HTTP_PORT` env variable). I can add an explicit test if you want but I think we're already covered here.

Connect #510 
Closes #510 

Connect #560 
Closes #560


XVincentX:
`config.loadConfig('gateway')` is not necessarily better than `config.loadGatewayConfig()`, but is definitely more generic, right?

`config.loadConfig(type)` is way easier to extend rather than adding yet another function in the future such as `config.loadAnotherConfig — config.loadYetAnotherConfigType` and the list could grow exponentially.

It does not have to happen necessarily, but it's good to have a code ready to accommodate such changes, if that'll happen.

I can restore the original functions and use the new one as underlying implementation; however this is not a public interface so I kind of miss the point of keep these.

DrMegavolt:
there is no other config
config.loadConfig('gateway') is error prone. count how many time you've used it 
I don't mind `config.loadConfig` but want to keep it internal 

and for update part it should use this.configTypes

XVincentX:
@DrMegavolt aa370311a164365789c335d569cc4a74fb12c76e
This should be what was missing from yesterday, right?

PR #627, opened 2018-02-02
Put better schema defaults to policies

### LINKED ISSUE #624: Review schemas to make sure "default" parameters are there

XVincentX:
The following PR will provide default values for some policies we neglected.

The way I figured out the parameters was to simply check the relative middleware documentation or eventually dig up in the source code to check what values were set if not provided.

Connect #624 
Closes #624

@kolbinski Can you see if these were the policies you were talking about of if there were other more that were concerning you?

### DIFF
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -50,17 +50,18 @@ module.exports = function (params, config) {
     const strategy = params.strategy || 'round-robin';
     balancer = createStrategy(strategy, proxyOptions, endpoint.urls);
   }
+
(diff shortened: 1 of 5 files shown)


lib/policies/proxy/proxy.js
```diff
@@ -50,18 +50,19 @@ module.exports = function (params, config) {
     const strategy = params.strategy || 'round-robin';
     balancer = createStrategy(strategy, proxyOptions, endpoint.urls);
   }
+
+  const intermediateProxySettings = process.env.http_proxy || params.proxyUrl;
```
DrMegavolt: 👍

Issue #589, opened 2017-12-21
Improve EG on Docker


DrMegavolt:
https://hub.docker.com/r/expressgateway/express-gateway/

has generic readme, but it should contain info how to mount config like here
https://www.express-gateway.io/docs/runtime/#configuration-loading-and-override

XVincentX:
Also, let's try to be part of the "Official images" program

DrMegavolt:
and for some reason it is not searchable with dash 
express-gateway. 
I need expressgateway

altsang:
- investigate being part of official docker image program

XVincentX:
1. Investigate moving building the images from CircleCI scripts (that fails everytime something small changes) to Docker cloud directly
2. Investigate and do the necessary changes so we can be part of the Docker Official Image Program (update image and readme)

XVincentX:
Create a different repository — https://github.com/mongo-express/mongo-express-docker might be a good source of inspiration.

Also — how to automate such thing when we release EG?

PR #628, opened 2018-02-04
Dockerfile Improvements

XVincentX:
Connect #589 

Reference: https://github.com/docker-library/official-images/pull/3980

This pull request will modify our docker files so it will conform to the guidelines to get into the Docker Official Image Program.

In particular:

* [x] Images should have a maintainer. That name is just the maintainer page and not the project maintainer. I put my name but — I'm more than happy to put another one :trollface: .

* [x] Keep the build context at the bare minimum (that is — make sure your `.dockerignore` file is removing anything that is not needed)

* [x] All ports should be exposed. This was cumbersome so I started to dig into that and it turns out that the `EXPOSE` command is not doing anything for real, it's just for documentation purposes. All container ports can always communicate when they're in the same network and you need to "map" them anyway using `-p` when using `docker run`. I checked other popular images and they're indeed doing it. (Outstanding point: what about the https port for traffic and admin api?)

* [x] You can't provide different images (that is we are doing right now, such as `-getting-started` and `-basic`. So I have modified the docker file to simply provide a bare installation ready to be used, once you have provided the `EG_CONFIG_DIR` environment variable. This is now mandatory.

Probably there might be other changes to do, so please consider this as a _WIP_, but start to have a look.


Dockerfile
```diff
@@ -1,19 +1,16 @@
...
 ENV CHOKIDAR_USEPOLLING true
 
-RUN mkdir -p /usr/src/app
-WORKDIR /usr/src/app
+RUN npm install express-gateway@$EG_VERSION
 
-RUN npm install express-gateway@$VERSION && \
-    ./node_modules/.bin/eg gateway create -n gateway -d . -t $TYPE && \
-    npm cache clean --force
+WORKDIR /node_modules/express-gateway
```
kevinswiber: This is pretty bizarre.  Why do this over `eg gateway create`?  Or it might be possible to just copy the `basic` template and use that as a starting point.

XVincentX: I analyzed different docker official images based on npm packages and it seems to be the standard. In particular:

1. You `npm install -g package-name` and then you start the package through the cli (that's something we do not have)

2. You install the package locally and you put yourself in the directory where the relevant files are. See [here for an example](https://github.com/mongo-express/mongo-express-docker/blob/master/Dockerfile#L33). We can put the default template as part of the configuration, though.

Our scaffolded mechanism won't be accepted — you'd have two gateway copies.

kevinswiber: What about doing 

```
COPY ./bin/generators/gateway/templates/basic/* /usr/src/app
WORKDIR /usr/src/app
RUN npm install express-gateway@$EG_VERSION
```

Thoughts?

XVincentX: I think that can work (if they do not force us to have the docker image in a separate file).

The basic template might need to be modified though because:

1. The admin API is listening on `localhost` only — so it won't be accessible outside the container.

2. The template is configuring the gateway to do basically nothing — so if anybody tries to send a request, they'll receive a `404`. I'm totally ok with that, I just wanted to make you guys aware of it.

XVincentX: > It turns out that if you scaffold a gateway, the hardcoded `configPath` takes the precedence [over the environment variable](https://github.com/ExpressGateway/express-gateway/blob/master/lib/index.js#L20-L21). We need to invert this.

@DrMegavolt Maybe you know/recall the reason?

DrMegavolt: I don't know why. +1 to making some generated files to run, not entire EG module. it should be possible to override server.js file as well in this case by mounting volume

Dockerfile-master
```diff
@@ -1,24 +1,17 @@
...
-EXPOSE 8080
-
-# HTTPS
-# EXPOSE 8443
+EXPOSE 8080 9876
 
-# Admin API
-EXPOSE 9876
-CMD [ "npm", "run", "start" ]
+CMD [ "node", "lib", "index.js" ]
```
XVincentX: Before you ask "Why not use `npm start`?" — it's because npm start, in docker environments, is swallowing signals (such as `SIGKILL`, `SIGTERM`) and not sending them to `node`, that should handle them accordingly.

XVincentX:
@kevinswiber Bump. Do you want to have another look into?

Dockerfile
```diff
@@ -1,19 +1,21 @@
...
 ENV CHOKIDAR_USEPOLLING true
 
-RUN mkdir -p /usr/src/app
-WORKDIR /usr/src/app
+VOLUME /var/lib/eg
 
-RUN npm install express-gateway@$VERSION && \
-    ./node_modules/.bin/eg gateway create -n gateway -d . -t $TYPE && \
-    npm cache clean --force
+RUN yarn global add express-gateway@$EG_VERSION
```
kevinswiber: Why use `yarn` here instead of `npm`?

XVincentX: I remember there was a reason but I forgot it and I didn't take a note. I'll retry and let you know.

I am quite sure npm was failing because of the way it store the global packages — and it's not `docker-ish`. yarn specifically solved this issue.

DrMegavolt: why use global at all?

XVincentX: It was local before, but Kevin was against that because of the weird work dir.
See [here](https://github.com/ExpressGateway/express-gateway/pull/628#discussion_r166337649) for the discussion.

DrMegavolt: To make docker page look normal it has to be another repo, right? so why not commit some generated code into the separate repo. so no need for global or magic path

XVincentX: It's not mandatory — but if it makes the thing easier I'm absolutely in for that. What do you have in mind precisely?

DrMegavolt: A new repo containing Docker file, Readme + whatever `eg gateway create` generates 
the last part is refreshed on each version build 

And It will make it easier for customers to build their own images, fork and build your own image

XVincentX: So I made a quick test — and unfortunately this approach will have another problem.

The generated `server.js` file is explicitly setting the dir path for the configuration files and models — and will ignore the environment variable (if provided). This means that if I mount my configuration files somewhere and I set the `EG_CONFIG_DIR` variable to the correct directly — the gateway will not honor the request.

Unless we fix that in the code (we're having the discussion in #632) we cannot do that.

Guys — I'd say we can iterate on this with the passing of the time. I'd rush just a bit more to make sure it gets on Docker official registry first. As the checks on their side will require time, we need to start the process as soon as possible.

XVincentX:
Missing thing: 

- [ ] Merge #644
- [ ] List all environment variables

Dockerfile
kevinswiber: @XVincentX If you think this is good enough to get into the official registry, go for it.  If we make changes, do they also have to go through a complete review process?  If so, I'd say let's wait to get the general approach right.  I think the idea of having a different directory or repo for the Docker image is a good one.  These limitations are superficial and can be overcome easily.

XVincentX: I think it's good enough.

I need to resurrect this PR a bit (it was a bit stale) and try to automate couple of things, but I can get this over the finish line tomorrow and merge it.

I've followed all the guidelines so — if nothing goes wrong — it should get accepted. I can't guarantee that, though.

PR #632, opened 2018-02-09
allow-in-code-config-update

DrMegavolt:
Appeared after discussion with @patrickleet in gitter. 

It is clear it should be a way to use custom code to modify config. For example based on docker/k8s/vault etc.secrets 

Originally I was thinking of separate concept of config post processor 
but actually it is likely that actual way developers will be using secrets is so deferent that it will be hard to build reusable post processors 

https://docs.google.com/document/d/1HrC75oHWeQG-50JjglR1UkkuzV4JUaUyrpDiMtE7Wgk/edit#heading=h.un7ixe95bi7m

so now it will be possible to just replace code in server.js from template declaration 
```
let gw = gateway()
  .load(path.join(__dirname, 'config’))
// do whatever I want by direct access to gw.config
gw.run()
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -13,16 +13,16 @@ if (require.main === module) {
 
     load (configPath) {
       this.configPath = configPath;
+      process.env.EG_CONFIG_DIR =
+      this.configPath || process.env.EG_CONFIG_DIR;
+      this.config = require('./config'); // this is to init config before loading servers and plugins
       return this;
     }
 
     run () {
-      process.env.EG_CONFIG_DIR =
-        this.configPath || process.env.EG_CONFIG_DIR;
-      const config = require('./config'); // this is to init config before loading servers and plugins
-      const plugins = pluginsLoader.load({config});
-      require('./gateway')({plugins, config});
-      require('./rest')({plugins, config});
+      const plugins = pluginsLoader.load({config: this.config});
+      require('./gateway')({plugins, config: this.config});
+      require('./rest')({plugins, config: this.config});
 
       return this;
     }


DrMegavolt:
@XVincentX  the server.js is a customer's property. As a user of EG I can completely replace the code there (to not even run EG). 

And it is ok that config is not matching what is in file. This is the intention with env vars support, for example. You don't want your secrets in yaml. And you have EG debug logs that will show what pipelines are processed, what policies are loaded. 

> this.configPath || process.env.EG_CONFIG_DIR
this is correct from what I see. if you want your own path you can provide it.
if you didn't provide it explicitly EG tries common place

lib/index.js
```diff
@@ -13,16 +13,16 @@ if (require.main === module) {
 
     load (configPath) {
       this.configPath = configPath;
+      process.env.EG_CONFIG_DIR =
+      this.configPath || process.env.EG_CONFIG_DIR;
```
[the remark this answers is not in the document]
patrickleet: Yea I see what you're saying.. config not matching what's happening could be pretty confusing.

This does look like it would work functionally though.

What if we stole a page from Jenkins.. in a Jenkinsfile, if you want to reference something defined in Jenkins Global Environment, it's referenced under the `env` key.

So in your Jenkinsfile you have Strings like "${env.monitorDomain}".

If we provided the config with a `secrets` object, then from the YAML file, those secrets could be referenced, and an error can be thrown if one is referenced but not defined, or fallback's could be provided like env the go syntax/docker-compose var's `${ENV_NAME:-default}`.

That way if you inspected the file, it would match what is running.

```
# Core
db:
  redis:
    host: ${REDIS_HOST:-redis}
    port: ${REDIS_PORT:-6379}
    namespace: EG

crypto:
  cipherKey: sensitiveKey
  algorithm: aes256
  saltRounds: 10

plugins:
  rewrite:
    package: express-gateway-plugin-rewrite

session: 
  storeProvider: connect-redis
  storeOptions:
    host: localhost
    port: 6379
  secret: ${secrets.session:-adefaultsecret} # replace with secure key that will be used to sign session cookie
  resave: false
  saveUninitialized: false
```

lib/index.js
```diff
@@ -13,16 +13,16 @@ if (require.main === module) {
...
+      this.config = require('./config'); // this is to init config before loading servers and plugins
```
kevinswiber: I believe this was in `run()` instead of `load()` due to a timing issue with loading the config.  That may be obsolete by now, but it's worth pointing out.

XVincentX: Timing issue? Do you recall an example in particular?

XVincentX:
Hold on guys. We are discussing two issues here.

--- 

Environment variables do **always** take precedence on any other in-code or configuration-file specified setting. It's the way you modify a system behavior without changing its source code.

If we break this expectation (code takes precedence over env variables, a lot of people will have troubles; me included every single time I instantiate a new gateway).

> the server.js is a customer's property.

I get it, but it's not something they do because they want. It's something they have because we force them to have (I would do the thing a bit differently, but that's another story).

In any case we are providing a file that is doing the wrong thing: deliberately overwriting the environment variable provided from outside. In-code values should be a fallback, not the default.

Consider this use case:

```bash
$ eg gateway create test
// Output things

$ cd test
$ EG_CONFIG_DIR=/var/lib/config npm start
```

The gateway does not work as expected — because it's using the `./test/config` directory instead of the one I specified.

> if you didn't provide it explicitly EG tries common place

This is not what is happening. EG always loads data from `egdir/config` as long I do not modify the `server.js` file.

Look at this image.

image

You can see the configuration files are being loaded from a different path than the one I **explicitly** set. Environment variable got ignored.

---

> And it is ok that config is not matching what is in file. This is the intention with env vars support, for example. You don't want your secrets in yaml. And you have EG debug logs that will show what pipelines are processed, what policies are loaded.

It is ok, no doubt. What it's not ok is that the Admin API propagates the changes back to the files, overwriting what should be there. I made this mistake with env variables as well, and we should fix it.

For instance; if I have

```yml
http:
  port: ${PORT:-8000}
# other things
```

and I run `PORT=3000 npm start` — the yml will get processed and `3000` gets used as port.

If I now use the admin API to make some modifications, the new configuration gets written and what I have is:

```yml
http:
  port: 3000
# other things
```

I start the gateway again:

`PORT=8080 npm start` — guess what? The gateway listens on a different port (3000) — because the plain change has been propagated. This is wrong, and it's partially my fault also (I completely forgot about this scenario).


How do we fix these two issues and — the need of making code changes to the configuration?

My proposal would be:

* Environment variables take the precedence, as it should be.
* The gateway has a "file-only" configuration loading method. This should be generalized making sure we can start the gateway using a plain Javascript object. In this way, loading from file would be a byproduct and it would improve these "programmability scenarios" dramatically
* Providing a `server.js` file just for the sake of starting the gateway is not cool in my opinion. We should provide an `eg run` command that will run the gateway with the specified configuration directory. Then — if you want to run the gateway programmatically because you want to do other things — you require it from your code and you start it `require('express-gateway').run()`. Regarding this, we still have #422 tracking it. 
* EG should keep a double representation of the configuration. One that's the "current running" and one that's the "raw unprocessed" one, so we can keep what the user put into in the time (such as environment variables).

These are all big changes, but we can get there. Just matter of time.

lib/index.js
kevinswiber: Don't quite remember now.  I know our config singleton is a fragile beast.  If all the tests are passing, it's probably no longer an issue.

kevinswiber:
> the server.js is a customer's property.
> 
> I get it, but it's not something they do because they want. It's something they have because we force them to have (I would do the thing a bit differently, but that's another story).

I think we should look at this a different way.  Allowing programmable access gives them opportunities they would otherwise not have.  In other words: It's not a bug; it's a feature.  A more rigid model means having a lot more forks instead of `npm installs`.  That gets much harder to manage, IMHO.

kevinswiber:
Let's not forget that there is a very real use case here: Allowing secrets that are resolved programmatically at runtime, in-process, without leaking into environment variables or an on-disk representation.

If there's a better proposal to achieve this, I'm all ears.  We could extend the plugin system to allow config-loaded/config-resolved hooks, for example.

DrMegavolt:
closing this as review unpassable

patrickleet:
curious - was there another method for allowing this introduced?

XVincentX:
@patrickleet Unfortunately no, we haven't been able to set up on a strategy and we also realized it's not a top priority. We will definitely revisit this, but not in the super short term.

The idea would be to make sure that loading configuration by file would be a by-product of a serious configuration loading strategy.

Cheers!

PR #633, opened 2018-02-10
Pass in logger so it may be consumed by policies

jbalsip:
Some thoughts I was having while reading the activity on ExpressGateway/express-gateway-plugin-rewrite#9 and it seems like there should be a logger being passed into policy middleware.  I can see where the pluginContext receives a logger, but I saw nothing for policies.  Does an approach like what I have in this PR make sense?

Without this, I suppose it could be possible to do `require('express-gateway/lib/logger')` but I'm not sure if that is necessarily optimal.

I also haven't quite got worked out how to put tests around this yet, just putting this out for discussion for now though.

### DIFF
--- a/lib/gateway/pipelines.js
+++ b/lib/gateway/pipelines.js
@@ -2,6 +2,7 @@ const express = require('express');
 const vhost = require('vhost');
 const log = require('../logger').gateway;
 const policies = require('../policies');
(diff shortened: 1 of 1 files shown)


jbalsip:
On second thought, I'll open an issue with some thoughts on this.  Doing this in a PR is getting a little ahead of myself.

Issue #630, opened 2018-02-07
redirect to client redirect_uri after oAuth2 session cleanup 


DrMegavolt:
current implementation of `site.logout` just removes the session of the user. 

the idea is to somehow before clearing the session identify client_id
find redirect_uri of the app and return user there

cliend_id  most likely will be in session

@StickNitro

DrMegavolt:
After some investigation this scenario will be unstable due to session expiration. 
Instead client should send returnTo parameter with url 

if client_id is provided. then EG must validate if the passed url is allowed by app 

more info https://auth0.com/docs/logout

DrMegavolt:
and default redirect uri should be made configurable for both login and logout (`/` currently)

PR #636, opened 2018-02-19
connect #630 redirect to returnTo url after logout

DrMegavolt:


DrMegavolt:
@StickNitro can you please verify it works as you expect

lib/policies/oauth2/site.js
```diff
@@ -1,13 +1,25 @@
...
+const services = require('../../services/index');
 
 module.exports.loginForm = (request, response) => response.render(path.join(__dirname, 'views/login'));
 
 module.exports.login = passport.authenticate('local', { successReturnToOrRedirect: '/', failureRedirect: '/login' });
 
-module.exports.logout = (request, response) => {
-  request.logout();
-  response.redirect('/');
+module.exports.logout = (req, res) => {
```
XVincentX: 👍 for the renaming

lib/policies/oauth2/site.js
```diff
@@ -1,13 +1,25 @@
...
+  req.logout();
+  if (!req.query.returnTo) {
+    return res.redirect('/'); // TODO: implement settings section for default view path and return redirects
+  }
+  const returnUrl = decodeURI(req.query.returnTo);
```
XVincentX: Triple check: I think that express is already decoding the parameters for us under the hood.

lib/policies/oauth2/site.js
```diff
@@ -1,13 +1,25 @@
...
+  if (!req.query.clientId) {
+    return res.redirect(returnUrl);
+  }
+
+  const client = services.application.get(req.query.clientId);
+  if (client && client.redirectUri && client.redirectUri === returnUrl) {
```
XVincentX: This is returning a `Promise` that should be resolved before accessing its data. These lines will probably make EG crash.

lib/policies/oauth2/site.js
```diff
@@ -1,13 +1,25 @@
...
+    return res.redirect(returnUrl);
+  }
+
+  res.status(400).send(`Invalid return URI ${returnUrl} for client id ${req.query.clientId}`);
```
XVincentX: `req.query.clientId` might be undefined — we should check for it.

lib/policies/oauth2/site.js
```diff
@@ -1,13 +1,25 @@
...
-  request.logout();
-  response.redirect('/');
+module.exports.logout = (req, res) => {
+  req.logout();
+  if (!req.query.returnTo) {
+    return res.redirect('/'); // TODO: implement settings section for default view path and return redirects
+  }
+  const returnUrl = decodeURI(req.query.returnTo);
+  if (!req.query.clientId) {
+    return res.redirect(returnUrl);
```
XVincentX: This is a bit concerning to me. 

What's happening here is basically: If you want to redirect to any URL you have, just do not provide the `clientID` and we're good to go.

This seems to be a way to skip the security check — something that usually you do not want to do. I'd not provide this feature at all. No clientId? You get an error, and I won't log you out.

Also — the clientId should be also verified whether it's still active.

.dockerignore
```diff
@@ -1 +1,7 @@
 node_modules/
+.vscode/
+.idea/
+.git/
+.circleci/
```
XVincentX: This is also part of #628 — but that's fine I guess.

lib/policies/oauth2/site.js
DrMegavolt: replicated from Auth0 behaviour

XVincentX: Can you point the documentation link? I just want to check that out.

DrMegavolt: see the ticket and https://auth0.com/docs/logout
there is actually a bigger problem here even with this PR it still works incorrectly 
If you have time you can investigate this 

Gitter conversation text ---

Neil Stevens @StickNitro Feb 20 16:04
@DrMegavolt Ok so in full,
Navigate to application (user not logged in so redirects to http://localhost:8080/oauth2/authorize?response_type=token&clientId=...&redirectUrl=...
User is prompted to login (using oauth2 credentials)
User is prompted to authorize application
User is redirected to the application (redirectUrl along with token which is cached in sessionStorage and used to access the backend)
Click logout in application (this clears the persisted token and redirects user to http://localhost:8080/logout`)
Logout url above currently redirects to /
At this point I manually navigate to my application and get Step 3

Serhii Kuts @DrMegavolt Feb 20 16:07
and it should be step 2 ? (assuming step 1 will happen automatically)

Neil Stevens @StickNitro Feb 20 16:07
Checking in Chrome dev tools there is a cookie `connect.sid’ that is still present after logout and redirect to /, if i manually remove this cookie after Step 6 above and manually navigate to my application then I get prompted to login (effectively starting at Step 1)

XVincentX: Ok thanks. I'll see if we can prioritize this and look into.

kevinswiber:
@XVincentX @DrMegavolt This has been sitting for 19 days.  Is there a resolution here?

XVincentX:
This is partially my fault, I haven't been really looking into this actively.

It's also missing tests. I can take over and bring this over the finish line, if required.

DrMegavolt:
@XVincentX I cannot work on this at the moment. If it is relevant please take over

XVincentX:
@DrMegavolt I think the feature is worth continuing. I'll prepare a card and discuss it during next sprint planning. Thanks for pinging me on this.

PR #639, opened 2018-02-24
Pass policy logger as an action argument

### LINKED ISSUE #634: Question about passing the logger down to the policy
Some thoughts I was having while reading the activity on ExpressGateway/express-gateway-plugin-rewrite#9 and it seems like there should be a logger being passed into policy middleware.  I could possibly pass the logger in pluginContext down to the policy, but it's the plugin logger and not the policy logger.  But that is of course if I want to hook into EG's logger instead of choosing my own.

I'm pretty sure if the official plugins are going to ultimately reside within their own repos so that does make me curious if there's a plan for this yet or not.

jbalsip:
Closes #634 

Per the conversation with @DrMegavolt this is my understanding of how the policy logger could be sent to the policy itself.  Then in usage I believe the logger could be accessed like this:

```JavaScript
module.exports = {
  policy: (params, config, logger) => {
  }
  // etc, etc
}
```

What I worry about with this, however, is with how this code could affect `...args`
https://github.com/ExpressGateway/express-gateway/blob/39ca89e4d31b2f9e0cc7d0898683c555b9d0ae2b/lib/gateway/pipelines.js#L134

A change there could then mean that the third parameter could no longer be the logger and could be a breaking change at that point.  Hence why I was thinking it may be better to pass it through pipelines.js instead.  But I leave it as a matter of discussion within this PR for consideration.

### DIFF
--- a/lib/policies/index.js
+++ b/lib/policies/index.js
@@ -27,7 +27,7 @@ class Policies {
     policy.policy = (params, ...args) => {
       const validationResult = validate(params);
       if (validationResult.isValid) {
-        return action(params, ...args);
+        return action(params, ...args, logger);
       } else {
         logger.error(`Policy ${name} params validation failed`, validationResult.error);
         throw new Error(`Policy ${name} params validation failed`);


DrMegavolt:
@XVincentX  take a look pls

DrMegavolt:
@XVincentX waiting for your decision

jbalsip:
It's been a while on this one so I think if I want to do this I should revisit it from a different angle.

XVincentX:
I'm sorry for the late reply, but I haven't been able to come here earlier.

We're about to start revisiting the plugin framework a bit, so this might change a lot over the short term. In meantime, I'd suggest you to grab the `logger` reference directly from the pluginContext.

I know this is an undocumented feature, but I'm not expecting to change it anytime soon.

`this.logger` should do the job.

Thanks, and once again apologize for the delay.

christiansaiki:
@XVincentX can you give an example, please?
I'm not sure how use logging inside a custom plugin with a policy.

```
module.exports = {
  name: 'authorizer',
  policy: () => (req, res, next) => {
    this.logger.warn('hello');
   next()
};
```
But I'm getting `TypeError: Cannot read property 'warn' of undefined`

PR #640, opened 2018-02-27
 Reword policy loading error messages and test the scenario

XVincentX:
This PR will make sure to use `log.error` to output error messages and reword them so it should be hopefully more clear.

In particular, we were not mentioning that a policy error might came because of a misconfigured (or missing plugin).

This has been raised after a discussion in the Gitter channel.

### DIFF
--- a/lib/gateway/pipelines.js
+++ b/lib/gateway/pipelines.js
@@ -197,7 +197,8 @@ function validatePipelinePolicies (policies, avaiablePolicies) {
     const policyNames = Object.keys(policyObj);
 
     if (avaiablePolicies.indexOf(policyNames[0]) === -1) {
-      throw new Error(`${policyNames[0]} policy not declared`);
+      log.error(`${policyNames[0]} policy not declared in the 'policies' gateway.config section`);
+      throw new Error('POLICY_NOT_DECLARED');
     }
   });
 }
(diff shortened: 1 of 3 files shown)


lib/gateway/pipelines.js
```diff
@@ -197,7 +197,8 @@ function validatePipelinePolicies (policies, avaiablePolicies) {
     const policyNames = Object.keys(policyObj);
 
     if (avaiablePolicies.indexOf(policyNames[0]) === -1) {
-      throw new Error(`${policyNames[0]} policy not declared`);
+      log.error(`${policyNames[0]} policy not declared in the 'policies' gateway.config section`);
```
iwko: Why do You use log here and logger in https://github.com/ExpressGateway/express-gateway/pull/640/files#diff-6dd40befa5e020684f04fdd9e88783b9R44

XVincentX: @iwko Because of different name in the top requires. See [here](https://github.com/ExpressGateway/express-gateway/blob/master/lib/gateway/pipelines.js#L3) and [here](https://github.com/ExpressGateway/express-gateway/blob/f04df4711b5400ec36d9b0f4dfe85d4c930654dc/lib/policies/index.js#L3).

Does that clarify?

iwko: ah, okay I missed that.

PR #648, opened 2018-03-05
Emit event before and after executing policy

dotBits:
With @XVincentX help I came up with this, which is related to its feature proposal https://feathub.com/ExpressGateway/express-gateway/+51

IMHO, this can be discussed and improved but I've found it very useful even on this first draft.
Any plugin or app module has to import the same shared eventBus file and attach listeners to it.

Any naming convention for events firing? Other suggestions?

### DIFF
--- a/lib/policies/index.js
+++ b/lib/policies/index.js
@@ -2,6 +2,7 @@ const fs = require('fs');
 const path = require('path');
 const logger = require('../logger').policy;
 const schemas = require('../schemas');
+const bus = require('../eventBus');
 
 class Policies {
   constructor () {
@@ -27,7 +28,12 @@ class Policies {
     policy.policy = (params, ...args) => {
       const validationResult = validate(params);
       if (validationResult.isValid) {
-        return action(params, ...args);
+        const act = action(params, ...args);
+        return (req, res, next) => {
+          bus.emit('EG::policy::beforePolicy', params, req, res, next);
+          act(req, res, next);
+          bus.emit('EG::policy::afterPolicy', params, req, res, next);
+        };
       } else {
         logger.error(`Policy ${name} params validation failed`, validationResult.error);
         throw new Error(`Policy ${name} params validation failed`);


altsang:
@dotBits this is great! I'm asking the team to do an overall architectural review and spec for hooks/events on our pipeline, would like to understand the big picture before we had down this path formally

kevinswiber:
FYI: This proposal really only works for the request flow.  Since Express doesn't have a legitimate response pipeline, it's much more difficult to track policy start and end events with respect to response modifications.

altsang:
yup good point, we need to figure out our strategy around shaping the response

dotBits:
@kevinswiber 
> This proposal really only works for the request flow

I think I got your point but, as far as I see, res.next will trigger subsequent beforePolicy event and res.end will stop any other policy to be executed, therefore pipeline is matching request and response flow and their modifications even in asynchronous scenarios.

Maybe the problem affects afterPolicy event, which is not guaranteed to be fired when response modifications are ready.
What do you think?

lib/policies/index.js
```diff
@@ -27,7 +28,12 @@ class Policies {
     policy.policy = (params, ...args) => {
       const validationResult = validate(params);
       if (validationResult.isValid) {
-        return action(params, ...args);
+        const act = action(params, ...args);
+        return (req, res, next) => {
+          bus.emit('EG::policy::beforePolicy', params, req, res, next);
+          act(req, res, next);
+          bus.emit('EG::policy::afterPolicy', params, req, res, next);
```
DrMegavolt: this part will be unstable. as mentioned in comments the moment you policy does any async call and return act will be released and bus.emit after will fire. 

so timming will be 
EG::policy::beforePolicy
some async action start
EG::policy::afterPolicy
async action finish
res.send

why do you need after event? What if I suggest to just leave before and remove after?

dotBits: @DrMegavolt would be fine to just leave before event. Thank you

XVincentX:
I'm going to close this for now, as we haven't been able to settle down on a story around this yet.

PR #652, opened 2018-03-06
Feat/oauth2 jwt consumer

leekeal:
Can you consider the default consumerId, Instead of using headers policy?

### DIFF
--- a/lib/policies/jwt/jwt.js
+++ b/lib/policies/jwt/jwt.js
@@ -21,6 +21,7 @@ module.exports = function (params) {
     }
 
     if (!params.checkCredentialExistence) {
+      jwtPayload.consumerId && (jwtPayload.id = jwtPayload.consumerId);
       return done(null, jwtPayload);
     }
 


XVincentX:
Thanks for the PR.

Can you explain a little bit more your use case? This is already handled in another PR and I do not really get why that line of code to be honest.

lib/policies/jwt/jwt.js
```diff
@@ -21,7 +21,8 @@ module.exports = function (params) {
     }
 
     if (!params.checkCredentialExistence) {
-      return done(null, { id: 'anonymous' });
+      jwtPayload.consumerId && (jwtPayload.id = jwtPayload.consumerId);
```
XVincentX: I think this is wrong.

The use case here is that, if we want to bypass the express-gateway identity system, we should just forward the payload and not make any assumption on it. Who's telling you that `jwtPayload.id` is always valorized and that its value corresponds to the consumer Id? That depends on your token provider.

Also, you're modifying the payload, while I think we should not touch it at all.

XVincentX:
I'll close this until @leekeal will provide further informations.

lib/policies/jwt/jwt.js
leekeal: Understand what you mean.
I think you are right.
thank you.

PR #654, opened 2018-03-06
oauth2 generate token can configure algorithm

leekeal:


lib/policies/jwt/jwt.js
```diff
@@ -21,7 +21,7 @@ module.exports = function (params) {
     }
 
     if (!params.checkCredentialExistence) {
-      return done(null, { id: 'anonymous' });
+      return done(null, jwtPayload);
```
XVincentX: When creating a pull request, please start from the master branch, not from another one.

lib/services/tokens/token.service.js
```diff
@@ -160,7 +160,8 @@ s.createJWT = function (payload) {
       issuer: config.systemConfig.accessTokens.issuer,
       audience: config.systemConfig.accessTokens.audience,
       expiresIn: config.systemConfig.accessTokens.timeToExpiry,
-      subject: config.systemConfig.accessTokens.subject
+      subject: config.systemConfig.accessTokens.subject,
+      algorithm: config.systemConfig.accessTokens.algorithm || "HS256"
```
XVincentX: We do not put default values in the code as long it is possible. Please modify the relative [json schema](https://github.com/ExpressGateway/express-gateway/blob/master/lib/config/schemas/system.config.json#L55) where you can put all the choices and the default value.

XVincentX:
Superseded by #662

Issue #647, opened 2018-03-02
Proxy policy — changeOrigin can be defined in two different places


XVincentX:
The changeOrigin parameter can be defined in two different places. This is confusing.

```yaml
- proxy:
    - action:
        serviceEndpoint: backend
        changeOrigin: true
        proxyOptions:
          changeOrigin: false
```

XVincentX:
This needs more work, unfortunately.

XVincentX:
Thjs will require a documentation update as well.

PR #661, opened 2018-03-08
Support PR for Proxy Policy

XVincentX:
Connect #647 

This PR will adjust couple of things in our Proxy policy that I found while digging for #647. 

In particular:

* Replaced some run time checks into the JSON Schema
* Using the right logger (`policy` instead of `logger`)
* Rearranged parameter so that the created `balancer` is constant — hopefully immutable
* Removed unused parameters in `prepareHeaders`
* Delete foreign parameters in `…File` parameters
* Make sure the balancer strategies do not modify the current object (this was really a pain in the ass to find — was breaking a lot of things)
* Create most of the proxy options upfront

In particular, I found out that the `proxyOption` parameter was being modified and overwritten by multiple parts of the code, footstep each other's work. When trying to DRY the code, it started to break.

Moreover I suspect we have a bug in our proxy policy but the current code is hard to mentally parse and reason about. Hopefully this will be a base for investigation.

### DIFF
--- a/lib/config/schemas/gateway.config.json
+++ b/lib/config/schemas/gateway.config.json
@@ -95,15 +95,29 @@
         "type": "object",
         "properties": {
           "url": {
-            "type": "string"
+            "type": "string",
+            "format": "uri"
           },
           "urls": {
             "type": "array",
             "items": {
-              "type": "string"
+              "type": "string",
+              "format": "uri"
             }
           }
-        }
+        },
+        "oneOf": [
+          {
+            "required": [
+              "url"
+            ]
+          },
+          {
+            "required": [
+              "urls"
+            ]
+          }
+        ]
       }
     },
     "policies": {
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -18,7 +18,7 @@ module.exports = function (params, config) {
   }
 
   const proxyOptions = parseProxyOptions(params.proxyOptions, endpoint);
-  const proxy = httpProxy.createProxyServer({ changeOrigin: params.changeOrigin });
+  const proxy = httpProxy.createProxyServer({ changeOrigin: params.changeOrigin, ...proxyOptions });
 
   proxy.on('error', (err, req, res) => {
     logger.warn(err);
@@ -51,27 +51,28 @@ module.exports = function (params, config) {
   }
 
   return function proxyHandler (req, res) {
-    const targetUrl = balancer.nextTarget();
-    proxyOptions.target = Object.assign(proxyOptions.target, targetUrl);
+    const target = balancer.nextTarget();
+    const headers = prepareHeaders(req.egContext);
 
-    logger.debug(`proxying to ${targetUrl.href}, ${req.method} ${req.url}`);
+    logger.debug(`proxying to ${target.href}, ${req.method} ${req.url}`);
 
-    prepareHeaders(params, proxyOptions, req.egContext);
-    proxy.web(req, res, proxyOptions);
+    proxy.web(req, res, { target, headers });
   };
 
   // multiple urls will load balance, defaulting to round-robin
 };
 
-function prepareHeaders (params, proxyOptions, egContext) {
-  proxyOptions.headers = proxyOptions.headers || {};
+function prepareHeaders (egContext) {
+  const headers = {};
   // Default headers always sent to downstream
   // TODO: allow configuration
   if (egContext.requestId) {
-    proxyOptions.headers['eg-request-id'] = egContext.requestId;
+    headers['eg-request-id'] = egContext.requestId;
   }
-  proxyOptions.headers['eg-consumer-id'] = egContext.consumer && egContext.consumer.id
+  headers['eg-consumer-id'] = egContext.consumer && egContext.consumer.id
     ? egContext.consumer.id : 'anonymous';
+
+  return headers;
 }
 
 // Parse endpoint URL if single URL is provided.
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -14,7 +14,7 @@ module.exports = function (params, config) {
   const endpoint = config.gatewayConfig.serviceEndpoints[serviceEndpointKey];
 
   if (!endpoint) { // Note: one day this can be ensured by JSON Schema, when $data keyword will be avaiable.
-    throw new Error(`service endpoint ${serviceEndpointKey} (referenced in 'proxy' policy configuration) does not exist`);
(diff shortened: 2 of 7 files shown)


lib/policies/proxy/proxy.js
```diff
@@ -12,73 +12,67 @@ const createStrategy = (strategy, proxyOptions, endpointUrls) => {
...
+    strategy = 'static';
+    urls = [endpoint.url];
   } else {
-    const strategy = params.strategy || 'round-robin';
-    balancer = createStrategy(strategy, proxyOptions, endpoint.urls);
+    strategy = params.strategy || 'round-robin';
+    urls = endpoint.urls;
   }
 
-  const intermediateProxySettings = process.env.http_proxy || params.proxyUrl;
```
DrMegavolt: Where is this part in the new code?

DrMegavolt: ah, found

lib/policies/proxy/proxy.js
```diff
@@ -12,73 +12,67 @@ const createStrategy = (strategy, proxyOptions, endpointUrls) => {
...
+  if (!endpoint) { // Note: one day this can be ensured by JSON Schema, when $data keyword will be avaiable.
+    throw new Error(`service endpoint ${serviceEndpointKey} (referenced in proxy policy configuration) does not exist`);
   }
 
-  if (!endpoint.url && !endpoint.urls) {
-    throw new Error(
-      `service endpoint ${serviceEndpointKey} (referenced in 'proxy' ` +
-      'policy configuration) does not contain a `url` or `urls` property');
+  const proxyOptions = parseProxyOptions(params.proxyOptions, endpoint);
+  const intermediateProxyUrl = process.env.HTTP_PROXY || params.proxyUrl;
```
DrMegavolt: process.env.http_proxy is not the same as process.env.HTTP_PROXY
add support for both. 

if you are refactoring it anyway add no_proxy support

lib/config/schemas/gateway.config.json
```diff
@@ -95,15 +95,29 @@
...
-              "type": "string"
+              "type": "string",
+              "format": "uri"
             }
           }
-        }
+        },
+        "oneOf": [
+          {
+            "required": [
```
DrMegavolt: do you plan next PR with "pattern" support? will it be `url`  property that is a pattern (then no need for format) or it will be a separate property ?

lib/policies/proxy/proxy.js
```diff
@@ -12,73 +12,67 @@ const createStrategy = (strategy, proxyOptions, endpointUrls) => {
...
+    proxy.web(req, res, { target, headers });
   };
 
   // multiple urls will load balance, defaulting to round-robin
 };
 
-function prepareHeaders (params, proxyOptions, egContext) {
-  proxyOptions.headers = proxyOptions.headers || {};
+function prepareHeaders (egContext) {
+  const headers = {};
```
DrMegavolt: what if I want some custom predefined headers? `proxyOptions.headers`

test/e2e/proxy-through-proxy.test.js
```diff
@@ -24,33 +24,43 @@ describe('@e2e @proxy through proxy', () => {
...
 
-      const server = srv.listen(0, () => {
-        process.env.http_proxy = 'http://localhost:' + server.address().port;
-        gwHelper.startGatewayInstance({ dirInfo, gatewayConfig }).then(({gatewayProcess}) => {
+      const server = srv.listen(0, (err) => {
+        if (err) {
+          return done(err);
+        }
+
+        process.env.HTTP_PROXY = `http://localhost:${server.address().port}`;
```
DrMegavolt: duplicate test for both env vars

lib/config/schemas/gateway.config.json
XVincentX: I'm not entirely sure I get what you mean, could you elaborate?

lib/policies/proxy/proxy.js
DrMegavolt: you can define some custom static headers via `proxyOptions.headers` option. It looks like it is no longer supported

XVincentX: You're right. Great catch! I'll fix this right now and maybe add a test as well for this.

lib/policies/proxy/proxy.js
XVincentX: Are we sure we want to support both?

You can define environment variables in any format you like, that's for sure. However I haven't been able to find any other meaningful project that's using environment variables in lower case format and, if we support it here, we either should support this scenario everywhere in the gateway (and it's a lot of places)  or we'd be inconsistent (like — some variables are ok to be lowercase, some are ok to be both, some should be just uppercase).

I think in this case is way better to be opinionated (it's uppercase, period) rather than trying to support every single use case. (camel case, kebab case, something_new case). Plus, uppercase it's _really_ the standard way.

@kevinswiber what's your thought here?

---

Let's postpone `no_proxy` for now — I do not know what's that about and I want to move forward with #647 and this is blocking me now.

XVincentX: It turns out minikube is reasoning with such lowercase environment variable. So maybe it should stay lowercase? I really do not know…

image

XVincentX: https://unix.stackexchange.com/questions/212894/whats-the-right-format-for-the-http-proxy-environment-variable-caps-or-no-ca

So it really seems like nobody knows.

lib/config/schemas/gateway.config.json
DrMegavolt: I mean what is the point in having `url` if soon you'll add another property like `pattern` that will be used to build url from expression. like `$1$2`
if you assume it will be same property `url` then it is ok to make it required. otherwise, you would soon have to revert this schema change

lib/policies/proxy/proxy.js
DrMegavolt: Why not have both? Ubuntu supports both variants.

lib/config/schemas/gateway.config.json
XVincentX: As far as I know, there's no short term plan to make that thing a pattern yet. We've been discussing this with Al but I think we have other priorities. Either way, I'd try to keep the new behavior in this same property, rather than adding a new one.

lib/policies/proxy/proxy.js
XVincentX: Yeah — it probably makes sense — I just want to make sure everybody agrees. Let's discuss it with @altsang and @kevinswiber during the standup and we'll go from there

XVincentX:
@DrMegavolt I added the support for `http_proxy` and added a test as well. Please re-review.

PR #667, opened 2018-03-12
Put some order on the credential service

### LINKED ISSUE #666: Inserting existing scopes throws "one or more scopes don't exists"
When trying to insert existing scopes, you get an error "one or more scopes don't exists".

flow - 
Add some scopes > try to add the same scopes again > Error.

problem could be the syntax / the error throwing itself (based on wanted behavior)

XVincentX:
The following pull request will put some orders in our credentials->scopes parts.

In particular:

* There was some confusion on `scope` and `scopes` parameters. The first one was misused and, moreover, undocumented. I've removed it and made sure all the occurrences are using always `scopes` as an array.
* The message when creating the same scope multiple times was broken. Fixed with the correct one.

Connect #666 
Closes #666

### DIFF
--- a/test/services/credential-service/credentials.test.js
+++ b/test/services/credential-service/credentials.test.js
@@ -181,7 +181,7 @@ describe('Credential service tests', () => {
 
     it('should not insert a credential with scopes if the scopes are not defined', () => {
       return should(credentialService.insertCredential(username, 'oauth2', _credential))
-        .be.rejectedWith('one or more scopes don\'t exist');
+        .be.rejectedWith('One or more scopes don\'t exist');
     });
 
     it('should insert a credential with scopes if the scopes are defined', () => {
@@ -247,7 +247,7 @@ describe('Credential service tests', () => {
 
     it('should not add scopes to existing credential if the scopes are not defined', () => {
       return should(credentialService.addScopesToCredential(username, 'oauth2', ['undefinedScope']))
-        .be.rejectedWith('one or more scopes don\'t exist');
+        .be.rejectedWith('One or more scopes don\'t exist');
     });
 
     it('should use default property if not defined', () => {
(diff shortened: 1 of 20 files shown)


kevinswiber:
Looks okay to me.  I would like @DrMegavolt to take a look, as well.  In general, I have a lot of concerns about our modifying our Identity Server functionality.  Changes always seem to have big cascading effects.

XVincentX:
@kevinswiber 

I know these are changes hard to review and maybe a bit harmful and I apologise for that, but I really want to fix these things when I end up spending half of a day to solve a simple bug because the code is mentally hard to follow and goes in unpredictable paths. 

Spend again half a day to rebuild the "flow-map" of the code on the scopes next time I'll have to work is something I definitely would like to avoid.

test/services/credential-service/credentials.test.js
```diff
@@ -158,13 +158,13 @@ describe('Credential service tests', () => {
...
       secret: 'password',
-      scopes: 'someScope',
+      scopes: ['someScope'],
       someProperty: 'propVal'
     };
 
     before(() => {
       Object.assign(config.models.credentials.properties.oauth2.properties, {
-        scopes: { type: 'string' },
+        scopes: { type: 'array', items: { tyep: 'string' } },
```
DrMegavolt: typo `tyep` , and if it works with this error I would just remove the line, since it is not needed

DrMegavolt:
Except for the typo, I also see `grabScopesAndExecute` method is now not covered by tests. not sure what it does but it would be good to restore coverage 

other than that looks ok

test/services/credential-service/credentials.test.js
XVincentX: Correct, this line is useless and can be removed. Thanks for spotting this.

PR #668, opened 2018-03-13
Update all depedencies

XVincentX:
Our package are a bit behind. This PR will bring us to the present again.

I've gone through all the changelogs. Things should be fine (TM)


bin/environment.js
```diff
@@ -140,10 +140,6 @@ exports.bootstrap = (eg, adapter) => {
     .demandCommand()
     .recommendCommands()
     .strict()
-    .alias('h', 'help')
```
kevinswiber: What's the impact of this block of deletions?

XVincentX: According to the `yargs` changelog, `--help, -h and --version` are now automatically added. It was a boilerplate for everybody and they decided to put it as default.

kevinswiber: Oh, I just saw that help and version are included by default.  However, I don't think `-h` is aliased by default, so we may need that.

XVincentX: That's true. However I also realised that alias is automatically added [here](https://github.com/ExpressGateway/express-gateway/blob/master/bin/eg-generator.js#L60)

image

kevinswiber: Just tried locally.  I get this:

```
➜ ./bin/index.js -h    
<snip>

Not enough non-option arguments: got 0, need at least 1
```

`--help` works, though.

XVincentX: The output on my machine is the same. Yargs is complaining of the "non-option" arguments.

image

XVincentX:
@kevinswiber You were right, I made the changes in the wrong part. Thanks for spotting this.

3e45da9 should fix the issue you were pointing.

PR #674, opened 2018-03-23
Add client certificate condition

XVincentX:
This PR will add a client certificate policy condition that can be used to check whether the current request has a valid client certificate, or not.

### DIFF
--- a/lib/conditions/predefined.js
+++ b/lib/conditions/predefined.js
@@ -156,10 +156,10 @@ module.exports = [
     }
   },
   {
-    name: 'clientAuthorized',
+    name: 'tlsClientAuthenticated',
     handler: (req, conditionConfig) => req.client.authorized,
     schema: {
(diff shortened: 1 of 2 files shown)


lib/conditions/predefined.js
```diff
@@ -154,5 +154,13 @@ module.exports = [
     schema: {
       $id: 'http://express-gateway.io/schemas/conditions/anonymous.json'
     }
+  },
+  {
+    name: 'clientAuthorized',
```
DrMegavolt: 1) what is `clientAuthorized`? Can it have a more descriptive name? It is confusing with auth policies we have. 
2) it is really common use case ? why it can't be own module?

XVincentX: I'm fine with both

1. A better descriptive name (I was just replicating the logic `client.authorized`
2. Have it as a separate module. @altsang is that ok with you?

PR #675, opened 2018-03-26
Set proxy policy parameter changeOrigin to false by default

XVincentX:
This PR will set the default parameter to `false` as it's stated in the documentation and as it should be — so that by default the proxy policy is transparent.

Merge #661 first


XVincentX:
This is not needed anymore — I'll update the docs accordingly.

PR #681, opened 2018-03-28
Deprecate the proxyOptions object

### LINKED ISSUE #647: Proxy policy — changeOrigin can be defined in two different places

XVincentX:
Connect #647 
Closes #647

### DIFF
--- a/lib/services/credentials/credential.service.js
+++ b/lib/services/credentials/credential.service.js
@@ -1,5 +1,6 @@
 const uuid62 = require('uuid62');
 const uuidv4 = require('uuid/v4');
+const clone = require('clone');
 const refParser = require('json-schema-ref-parser');
 const mergeAllOf = require('json-schema-merge-allof');
 const utils = require('../utils');
@@ -292,7 +293,7 @@ function validateNewCredentialScopes (credentialConfig, credentialDetails) {
 // This function validates all user defined properties, excluding scopes
 function validateNewCredentialProperties (credentialConfig, credentialDetail) {
   // Tmp — horrible hack to remove.
-  const credentialDetails = JSON.parse(JSON.stringify(credentialDetail));
+  const credentialDetails = clone(credentialDetail);
   delete credentialDetails.scopes;
 
   const validationResult = validate(credentialConfig, credentialDetails);
--- a/lib/services/credentials/credential.service.js
+++ b/lib/services/credentials/credential.service.js
@@ -1,6 +1,5 @@
 const uuid62 = require('uuid62');
 const uuidv4 = require('uuid/v4');
-const clone = require('clone');
 const refParser = require('json-schema-ref-parser');
 const mergeAllOf = require('json-schema-merge-allof');
 const utils = require('../utils');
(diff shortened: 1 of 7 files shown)


lib/services/credentials/credential.service.js
```diff
@@ -292,7 +293,7 @@ function validateNewCredentialScopes (credentialConfig, credentialDetails) {
 // This function validates all user defined properties, excluding scopes
 function validateNewCredentialProperties (credentialConfig, credentialDetail) {
   // Tmp — horrible hack to remove.
-  const credentialDetails = JSON.parse(JSON.stringify(credentialDetail));
+  const credentialDetails = clone(credentialDetail);
```
DrMegavolt: Are there deep properties? Why not object.assing?

test/policies/oauth/jwt-token.js
```diff
@@ -14,8 +15,8 @@ const serverHelper = require('../../common/server-helper');
 const config = require('../../../lib/config');
 const testHelper = require('../../common/routing.helper')();
 
-const originalGatewayConfig = JSON.parse(JSON.stringify(config.gatewayConfig));
-const originalSystemConfig = JSON.parse(JSON.stringify(config.systemConfig));
+const originalGatewayConfig = clone(config.gatewayConfig);
```
DrMegavolt: just use new Config({}) to create your own config. copying some implicit config can result in hard track error if this default config will be changed

XVincentX: Oh interesting. Let's see if I can make it work with your suggestion.

lib/services/credentials/credential.service.js
XVincentX: Unfortunately yes, it has got deep properties and I am forced to make a clone of this object because functions here and there modify this object and make the whole flow break.

I hope I can get rid of this once we'll refactor the entire credential service, but I am afraid we will have to live with this for a while.

DrMegavolt: actually, the strange part is that `validate` modifies some object.  

if this is a default behavior of `ajv.validate` then this clone should probably be inside lib/schemas generic method. unlikely that this is the only place that has a complex structure to validate  

In any case, are you sure that it is really required to modify the object?

XVincentX: The reason why `ajv.validate` is modifying the object is because:

1. It puts default values on properties that haven't been provided
2. Performs coercing (automatic conversion) on the passed object.

So it's something that we effectively want.

Now to be honest I do not really remember why I had to put such a bad hack in the code (it's completely my fault, I should have commented it better; I'll make sure to do that next time), but I recall I was spending a lot of time on it.

If that's ok with you, I'd skip this for now and open a separate issue; otherwise I could get stuck trying to remember the details and delaying this pull request a lot.

Does that make sense to you?

XVincentX:
> check for new Config usage in other tests. it will make test configuration explicit

@DrMegavolt is it ok to create a separate issue for it? I had a look into but it seems like the way the config singleton is managed is basically breaking half of the tests when I try to use a different instance. That'd drain a significant amount of time and prevent this from getting merged.

DrMegavolt:
@XVincentX  I would not change the code without clear understanding. So I suggest to not use `clone`, just keep the previous hack. It does not prevent task completion. 

I'm ok with a new task to clean up this code and fix not needed cloning of config

XVincentX:
Oh ok sure, it was just to remove that `JSON.parse(JSON.stringify`, but we can do that as a separate step. I'll revert.

XVincentX:
@DrMegavolt Done!

PR #683, opened 2018-04-04
Explicitly send expires_in field

### LINKED ISSUE #679: oauth2 expiry time is not present
per Davendra, - needs to be verified as a bug

XVincentX:
Connect #679
Closes #679

The following PR will make sure to explicitly set the `expires_in` field in all the oAuth2 responses.

While developing this change, I realized how much code for the tests scenario was copied and pasted here and there, making changes almost impossible (copy and paste the things around 5-6 times). I've extracted such code in a shared file that's now used around.

### DIFF
--- a/test/oauth/testUtils.js
+++ b/test/oauth/testUtils.js
@@ -1,4 +1,5 @@
 const should = require('should');
+const uuid = require('uuid62');
 
 const services = require('../../lib/services');
 const db = require('../../lib/db');
@@ -16,7 +17,7 @@ module.exports = {
 
   createOAuthScenario: () => {
     const user1 = {
-      username: 'irfanbaqui',
+      username: uuid.v4(),
       firstname: 'irfan',
       lastname: 'baqui',
       email: 'irfan@eg.com'
--- a/test/oauth/testUtils.js
+++ b/test/oauth/testUtils.js
@@ -1,0 +1,58 @@
+const should = require('should');
+
+const services = require('../../lib/services');
+const db = require('../../lib/db');
+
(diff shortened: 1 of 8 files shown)


test/oauth/testUtils.js
```diff
@@ -0,0 +1,58 @@
...
+module.exports = {
+  checkTokenResponse: (response, additionalProps = []) => {
+    should(response).have.properties('access_token', 'expires_in', ...additionalProps);
+    should(response.token_type).be.eql('Bearer');
+    should(response.access_token.length).be.greaterThan(15);
+  },
+
+  createOAuthScenario: () => {
+    const user1 = {
+      username: 'irfanbaqui',
```
DrMegavolt: how about using a unique name? Same data across different tests can produce hard to track issues. Especially, if cleanup for some reason will not work.

XVincentX: Like `uuid.v4()`?

DrMegavolt: I've used base62 id, it is more readable. uuid is ok, probably some phrase generator like `little-brown-fox` can work better. up to you

XVincentX: I'm going to use uuid62 for now, but maybe it's time to introduce https://github.com/marak/Faker.js/ — it's able to generate decent fake data as well as data based on JSON schema (that's really handy for us)

DrMegavolt: faker looks good. but please one by one as part of other tasks. no need to refactor entire code base at one step

PR #687, opened 2018-04-06
Always put the credential ID and delete the hashed data

### LINKED ISSUE #678: credential generated not equal to credential looked up
per discussed with Davendra

needs to be verified as a bug

XVincentX:
## WIP

Connect #678 
Closes #678


lib/services/credentials/credential.service.js
```diff
@@ -254,6 +247,21 @@ s.setScopesForCredential = function (id, type, scopes) {
     }).then(() => true);
 };
 
+function processCredential (credential, options = { includePassword: false }) {
+  if (credential.scopes && credential.scopes.length > 0) {
+    credential.scopes = JSON.parse(credential.scopes);
+  }
+  const credentialModel = config.models.credentials.properties[credential.type];
+  if (!options.includePassword && credentialModel.properties && credentialModel.properties.passwordKey) {
```
iwko: IMO better idea is to move these conditions to function

XVincentX: Can you elaborate a little bit more here? How would you rephrase it?

lib/services/credentials/credential.dao.js
```diff
@@ -114,6 +114,7 @@ dao.insertCredential = function (id, type, credentialObj) {
   }
   const key = buildIdKey(type, id);
   if (type === 'key-auth') {
+    if (credentialObj.keyId) credentialObj.id = credentialObj.keyId;
```
DrMegavolt: it looks suspicious to me. isn't keyId some separate property? 
there is `id` incoming. As I understand it is used to reference key-auth credentials. so it could break listing creds 

The related task is about password hash, why there are changes to key-auth at all ?

lib/services/credentials/credential.service.js
```diff
@@ -254,6 +247,21 @@ s.setScopesForCredential = function (id, type, scopes) {
...
+    delete credential[credentialModel.properties.passwordKey.default];
+    delete credential.passwordKey;
```
DrMegavolt: what is ` credential.passwordKey`?

test/cli/credentials/list.test.js
```diff
@@ -44,58 +46,60 @@ describe('eg credentials list -c ', () => {
...
+    env.prepareHijack();
+  });
+
   afterEach(() => {
     env.resetHijack();
   });
 
+  it('should have the id property, but not the secret ones', () => {
+    createdTypes.raw.forEach(cred => {
+      should(cred).not.have.properties('secret', 'password');
```
DrMegavolt: If you remove them from response how would I know autogenerated values for a password? Insert op should still have a secret in response

XVincentX: This is not an insert operation, it's a listing one. See the [ran command](https://github.com/ExpressGateway/express-gateway/pull/687/files#diff-c7a7013b987baf8de10bb621bf51006eR160)

lib/services/credentials/credential.dao.js
XVincentX: So the problem here is that — we have yet another inconsistency when returning the credentials. Some of them return an `id` as an identifier, some of them return `keyId`.

So when you use the API you don't know where the id is stored. You need to look into `id` and if it does not exist, you look for `keyId` — this is confusing (it was confusing an user).

I agree that's suspicious *and* an horrible solution, but I haven't found anything better without breaking the API. Open to better solutions!

PR #690, opened 2018-04-11
Admin API: improve query capabilities for Users and Applications

### LINKED ISSUE #689: Admin API: Apps / Users improvements
- Return username in apps
- Fix `nextKey` format to be always numeric.
- Return accurate counts on API response.
- Allow setting `count` on API request.

XVincentX:
Connect #689 
Closes #689 

This PR will:

* Make sure the `nextKey` parameter is always an integer
* Correctly extract and send to Redis the `start` parameter so that the client can decide the starting point for the function
* Move default parameters into the `dao` part of the data storage, so services do not have to worry about it

### DIFF
--- a/admin/users.js
+++ b/admin/users.js
@@ -32,7 +32,7 @@ module.exports = function (client) {
         .get(`${baseUrl}${encodeURIComponent(id)}`)
         .then(res => res.body);
     },
-    list (params) { // TODO: add pagination
+    list (params) {
       return client
         .get(baseUrl)
         .query(params)
--- a/admin/users.js
+++ b/admin/users.js
@@ -32,9 +32,10 @@ module.exports = function (client) {
         .get(`${baseUrl}${encodeURIComponent(id)}`)
         .then(res => res.body);
     },
-    list () { // TODO: add pagination
+    list ({ start, count }) { // TODO: add pagination
       return client
         .get(baseUrl)
+        .query({ start, count })
         .then(res => {
           return res.body;
         });
--- a/admin/users.js
+++ b/admin/users.js
@@ -32,10 +32,10 @@ module.exports = function (client) {
         .get(`${baseUrl}${encodeURIComponent(id)}`)
         .then(res => res.body);
     },
-    list ({ start, count }) { // TODO: add pagination
+    list (params) { // TODO: add pagination
       return client
         .get(baseUrl)
-        .query({ start, count })
+        .query(params)
         .then(res => {
           return res.body;
         });
--- a/lib/services/consumers/application.dao.js
+++ b/lib/services/consumers/application.dao.js
@@ -53,9 +53,9 @@ dao.update = function (id, props) {
     });
 };
 
-dao.findAll = function ({ start = 0, count = 100 } = {}) {
+dao.findAll = function ({ start = 0, count = '100' } = {}) {
   const key = appHashKey('');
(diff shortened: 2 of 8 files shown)


admin/users.js
```diff
@@ -32,9 +32,10 @@ module.exports = function (client) {
         .get(`${baseUrl}${encodeURIComponent(id)}`)
         .then(res => res.body);
     },
-    list () { // TODO: add pagination
+    list (params) { // TODO: add pagination
```
DrMegavolt: pls rm TODO, nextKey is for pagination

lib/services/consumers/application.dao.js
```diff
@@ -53,11 +53,10 @@ dao.update = function (id, props) {
     });
 };
 
-dao.findAll = function (query) {
+dao.findAll = function ({ start = 0, count = '100' } = {}) {
```
DrMegavolt: a little strange that `count` is a string. Is it required by the ioredis library? 
What will happen if I pass not a number? No validation as I understand?

test/rest-api/applications.js
```diff
@@ -1,58 +1,56 @@
 const should = require('should');
 const idGen = require('uuid62');
 const adminHelper = require('../common/admin-helper')();
+const username = idGen.v4();
 
 describe('REST: Applications', () => {
```
DrMegavolt: Can you add the same tests for users? 
And I've seen the following behavior:
count=100 actually returns 14 users. Each user has 7 fields so it is 98 <100;  user 15 is on the next page
Interesting that app tests seem to behave differently. you request count 3 and check for 3 apps

Is it the difference between real redis vs mock or apps are implemented differently than users?

admin/users.js
XVincentX: Oh yeah, that's right!

lib/services/consumers/application.dao.js
XVincentX: Good point. I set it as string because Redis is accepting strings only and — given that http query parameters are strings as well, I saw a perfect match. I'll look into, if it accepts numbers as well then better.

test/rest-api/applications.js
XVincentX: Sure, I'm going to replicate the test.

Regarding the count — it seems to be a bit complicated.

 I've been digging into the docs this morning and it seems to be working a little bit differently compared to what you said yesterday (or maybe I didn't get what you said correctly).

From Redis Docs:

> It is possible to empirically adjust the behavior of SCAN using the COUNT option

> Basically with COUNT the user specified the amount of work that should be done at every call in order to retrieve elements from the collection. This is just a hint for the implementation.

So the `count` parameter is not really about "how big you want the resultset to be" but rather "how many keys I'm going to process per each iteration"

I've also checked the `ioredis-mock` implementation for the `scan` command and what they do is
1. Fetch **all** the keys
2. Select a window of `count` keys
3. Filter the result using the match statement

So — if in the particular range there are no matching keys, you'll get an empty result, which is what the docs is stating as well: 

> The commands are also allowed to return zero elements, and the client should not consider the iteration complete as long as the returned cursor is not zero.

Also — but I might be wrong here — the `scan` command is iterating exclusively on all the keys in the database, it never fetches their values. So I'm not sure how the number of field that a key has is involved in the discussion.

Does that clarify?

DrMegavolt: aha, that is interesting. so some pages might be empty. looks good for now

PR #691, opened 2018-04-12
Enable username resolution in apps route in Admin API

### LINKED ISSUE #689: Admin API: Apps / Users improvements

XVincentX:
* Return the username from apps calls

Connect #689 
Closes #689 

Note: there's a dangerous circular dependency that is basically a code smell. Yet another sign that the credential service is accumulating technical debt :(

### DIFF
--- a/lib/services/consumers/application.service.js
+++ b/lib/services/consumers/application.service.js
@@ -28,6 +28,7 @@ s.insert = function (_app, userId) {
 };
 
 s.get = function (id) {
+  const userService = require('./user.service');
   return applicationDao.get(id)
     .then(app => {
       if (!app) {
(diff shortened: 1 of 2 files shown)


lib/services/consumers/application.service.js
```diff
@@ -51,11 +60,20 @@ s.find = function (appName) {
     });
 };
 
-s.findAll = function (query = {}) {
+s.findAll = function (query) {
+  const userService = require('./user.service');
```
DrMegavolt: I'm not sure that this preload is required all the time. And what if one needs more info, like Firstname?  

how about `/users` will allow query params to filter by ids
something like `/users?ids=12,34,45`=> List of Users

XVincentX:
@DrMegavolt Shall we close this for now maybe or keep it here around?

XVincentX:
Closed until further decisions.

Issue #672, opened 2018-03-21
Fix base url definition for CLI


XVincentX:
Follow up to #669 

The current solution I implemented was kind of an hack and won't be working with all the use cases.

If an user is writing something such as:

```yml
cli:
  url: http://localhost:9876/
```

We'll have the same problem again. This is because of the poor way superagent is handling the baseURL with regards of the single paths.

On this regards, `axios` does a way better job, and it might be worth actually to switch to it.

PR #693, opened 2018-04-16
Use gateway output instead of timers to signal the gateway is ready

### LINKED ISSUE #694: eg tokens revoke is throwing an internal warning
When running `eg tokens revoke` command, yargs is throwing an internal warning for a bad function call.

image

XVincentX:
Connect #672 
Connect #694 
Closes #694 

This PR will change the way we signal to the test framework that the gateway is ready. Instead of making multiple attempts separated by a timeout, we'll wait till the gateway will output the ready sentences.

This should decrease a bit the flakyness of our tests.

### DIFF
--- a/test/common/gateway.helper.js
+++ b/test/common/gateway.helper.js
@@ -50,7 +50,7 @@ module.exports.startGatewayInstance = function ({ dirInfo, gatewayConfig }) {
         const gatewayProcess = fork(modulePath, [], {
           cwd: dirInfo.basePath,
           env: childEnv,
-          stdio: 'pipe'
+          stdio: ['pipe', 'pipe', 'pipe', 'ipc']
         });
 
         gatewayProcess.on('error', reject);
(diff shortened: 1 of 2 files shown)


test/common/gateway.helper.js
```diff
@@ -49,30 +49,23 @@ module.exports.startGatewayInstance = function ({ dirInfo, gatewayConfig }) {
           'lib', 'index.js');
         const gatewayProcess = fork(modulePath, [], {
           cwd: dirInfo.basePath,
-          env: childEnv
+          env: childEnv,
+          stdio: ['pipe', 'pipe', 'pipe', 'ipc']
```
DrMegavolt: do you need `ipc`? As I understand you are just waiting for log output to understand that the process has started (which btw is great idea)

XVincentX: It seems like it's mandatory to specify all the parameters, however starting with Node8 you can just put `stdio: 'pipe'` and that'll do the job for you. See [here](https://github.com/nodejs/node/commit/3268863ebc40d1f0beee61b044c492b43fa57fa5#diff-1dea15b19e826b37828eef6c41166ffeR20).

As long we'll support Node 6 — we'll have to supply the whole array.

PR #698, opened 2018-04-18
Add log line for sent headers

XVincentX:


lib/policies/headers/headers.js
```diff
@@ -1,10 +1,12 @@
...
+
 module.exports = function (params) {
   return (req, res, next) => {
     for (const key in params.forwardHeaders) {
       const val = params.forwardHeaders[key];
-      // key is new header name that will be prefixed with `headersPrefix`
-      // val is some JS expression to execute against egContext
-      req.headers[params.headersPrefix + key] = req.egContext.run(val);
+      const headerName = params.headersPrefix + key;
+      log.debug(`Adding ${headerName} header to the request with value: ${val}`);
```
DrMegavolt: It can reveal sensitive info in logs, `val` should not be in logs

PR #700, opened 2018-04-18
Expose server

jinasonlin:
Expose gateway server and admin server.

```javascript
const eg = gateway()
  .load(path.join(__dirname, 'config'))
  .run();

Promise
  .all([eg.gateway, eg.admin])
  .then(([{ app, httpsApp }, admin]) => {
    // gateway server and admin server
    //
    // app.close();
  });
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,6 +9,8 @@ if (require.main === module) {
   class Main {
     constructor () {
       this.configPath = null;
+      this.gateway = null;
+      this.admin = null;
     }
 
     load (configPath) {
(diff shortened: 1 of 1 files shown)


XVincentX:
This is part of #422 and I do not think we should be doing it in this way.

Ideally, the `run` method should return a Promise that, once resolved, would return the _various_ express apps (http server, https server, admin api).

jinasonlin:
I agree with you.
In cluster mode, when `uncaughtException` emit in worker, it should be graceful exit express apps.
Is there have any way?

XVincentX:
Unfortunately right now no, and your use case is more than legit. I'll bump up this issue in one of our next sprints.

If you could give me a little bit more details on your use case, I can try to craft a story around it. Feel free to write it here or stop by in our Gitter channel

jinasonlin:
Case with cluster

server.js
```javascript
const cfork = require('cfork');

// fork and restart
cfork({
  exec: './worker.js',
});
```

worker.js
```javascript
const gateway = require('express-gateway');
const path = require('path');
const graceful = require('graceful');

gateway()
  .load(path.join(__dirname, 'config'))
  .run()
  .then(([{ app, httpsApp }, admin]) => {
    const servers = [];

    if (app) servers.push(app);
    if (httpsApp) servers.push(httpsApp);
    if (admin) servers.push(admin);

    // graceful exit
    graceful({
      servers,
      killTimeout: '10s',
    });
  });

```

XVincentX:
Perfect, it now makes sense and it's a small step forward. @DrMegavolt what do you think?

PR #706, opened 2018-04-24
Spell thrown error correctly on apps routes

XVincentX:
Unfortunately some parts of our error handling is depending on the message content.

The lack of the final `s` was making our Admin API to return a 500 error message instead of a 409.


XVincentX:
@DrMegavolt I've added the missing test for this.

PR #707, opened 2018-04-24
Make sure cli start URL does not end with a slash

### LINKED ISSUE #672: Fix base url definition for CLI

XVincentX:
Connect #672 
Closes #672  

This PR will workaround a limitation in superagent, where its `prefix` plugin is breaking in case the CLI url ends with a slash.

This is temporary. The long term solution here would be to replace the client entirely with axios, which is better maintained as well as more complete by default.

### DIFF
--- a/bin/eg-generator.js
+++ b/bin/eg-generator.js
@@ -77,6 +77,12 @@ module.exports = class EgGenerator extends Generator {
       baseURL = `http://${hostname}:${port}`;
     }
 
+    /*
+      This bad hack is required because of the weird way superagent is managing urls. Hopefully this is not
+      going to be here forever — we'll replace the client with Axios hopefully.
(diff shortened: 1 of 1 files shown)


XVincentX:
Testing this small piece seems to be kind of a pain in the ass. As I'd like to refactor this part in the following sprints, I'd hold off on this for now. :trollface:

XVincentX:
Sure. Will add and merge.

PR #713, opened 2018-05-01
fix server start method change in the test

taewdy:
* hot-reload.test.js : starts server and waits for certain time for server to start.
* changed to Promise/resolve on server message produced


test/e2e/hot-reload.test.js
```diff
@@ -25,6 +25,19 @@ describe('hot-reload', () => {
...
     let watcher = null;
 
+    function startServer (options) {
+      return new Promise(
+        (resolve, reject) => {
+          childProcess = spawn('node', [options.modulePath], {cwd: options.cwd, env: options.env});
+          childProcess.stdout.on('data', (data) => {
+            const chk = /gateway http server listening on/g;
+            if (chk.test(data)) {
+              resolve('server started');
```
XVincentX: I'd just resolve the promise, with no returned data.

test/e2e/hot-reload.test.js
```diff
@@ -71,23 +84,24 @@ describe('hot-reload', () => {
...
+                    .then(res => {
+                      request
+                        .get(`http://localhost:${originalGatewayPort}`)
+                        .end((err, res) => {
+                          should(err).not.be.undefined();
+                          should(res.unauthorized).not.be.undefined();
+                          return done();
+                        });
+                    })
+                    .catch(e => {
```
XVincentX: `.catch(done)`

test/e2e/hot-reload.test.js
```diff
@@ -71,23 +84,24 @@ describe('hot-reload', () => {
...
-                        done();
-                      });
-                  }, GATEWAY_STARTUP_WAIT_TIME);
+                  };
+                  startServer(options)
+                    .then(res => {
+                      request
+                        .get(`http://localhost:${originalGatewayPort}`)
+                        .end((err, res) => {
+                          should(err).not.be.undefined();
```
XVincentX: Please use `if (err) { return done(err) }` so Mocha can tell us what the error was.

test/e2e/hot-reload.test.js
XVincentX: Also — maybe we shuold reject the promise if the reg exp does not pass, so we do not waste our time waiting for a resolve that'll never happen?

test/e2e/hot-reload.test.js
taewdy: Hi @XVincentX, 
Thanks for the good work on Express Gateway, and thanks for the comment. 
I agree with you, specially the Promise.reject when RegExp fails.
I will do another PR. With a new PR, I'm also happy if you modify any short comings. 
Thanks

PR #714, opened 2018-05-01
Update bcrypt installation

XVincentX:
`bcrypt 1.0.3` is old and it seems like it has also got some security problems.

image

### DIFF
--- a/lib/services/utils.js
+++ b/lib/services/utils.js
@@ -1,8 +1,4 @@
-require('util.promisify/shim')(); // NOTE: shim for native node 8.0 uril.promisify
-
-const util = require('util');
 const crypto = require('crypto');
-
 const config = require('../config');
 
 let bcrypt;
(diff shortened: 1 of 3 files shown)


XVincentX:
> After this change, one will not be able to downgrade, right?

This is correct — in terms of bcrypt 2.x hashes are not compatible with it's 1.x version. The contrary is true though.

> And this means that bcryptjs will not be able to read values created with the new version?

The compatibility stays. From the official repository:
> Hashes generated in earlier version of bcrypt remain 100% supported in v2.x.x.

> And node10 support is not done yet, so why rushing? No real benefits, right? N-API or not, how it matters?

Hmm, this is not because of Node 10 — `bcrypt 1.x` works perfectly fine with Node 10 as well. The main reason for the upgrade is because the 1.x version has been declared as insecure

XVincentX:
@DrMegavolt Please see also #724 — maybe it's going to be way easier in this way.

PR #717, opened 2018-05-09
Add logger config

jinasonlin:
This pr support to solve the longer configuration for the current version.
Support the two configurations "timestamp" and "colorize" of winston consult transport.


XVincentX:
The changes look good and reasonable. The only thing we need to sort it out is whether we want to make this option to be per-logger or a global one. @DrMegavolt 

Thoughts?

XVincentX:
@jinasonlin Thanks! We'll make sure to put it on the next release.

PR #721, opened 2018-05-09
Use bcryptjs for Alpine Docker images

### LINKED ISSUE #719: Docker images have a segmentation fault problem with bcrypt
Alpine linux based images use musl libc instead of the standard glibc and thus are not compatible with compiled binaries. node-pre-gyp currently does not test the pre-compiled binaries to be ABI-compatible and thus you may see segfaults.

Use multi stage build to recompile the library and then copy the binares in a brand new image

XVincentX:
Connect #719 
Closes #719

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,26 @@
-FROM node:8-alpine
+FROM node:8-alpine AS base
 
-LABEL maintainer Vincenzo Chianese, vincenzo@express-gateway.io
-
-ARG EG_VERSION
 ENV NODE_ENV production
 ENV NODE_PATH /usr/local/share/.config/yarn/global/node_modules/
+
+FROM base as assets
+
+RUN apk --no-cache add --virtual builds-deps build-base python
+RUN yarn global add bcrypt --force
+
+FROM base
+LABEL maintainer Vincenzo Chianese, vincenzo@express-gateway.io
 ENV EG_CONFIG_DIR /var/lib/eg
+
 # Enable chokidar polling so hot-reload mechanism can work on docker or network volumes
 ENV CHOKIDAR_USEPOLLING true
+ARG EG_VERSION
 
 VOLUME /var/lib/eg
 
 RUN yarn global add express-gateway@$EG_VERSION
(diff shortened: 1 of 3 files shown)


XVincentX:
@kevinswiber Sure, we can definitely use `bcryptjs` and make the day.

Just one thing more: this is not complicating the flow for the final user. They'll keep downloading the ready to use image but I agree that this is complicated. I'll modify the PR.

Dockerfile
```diff
@@ -11,7 +11,7 @@ ENV CHOKIDAR_USEPOLLING true
 
 VOLUME /var/lib/eg
 
-RUN yarn global add express-gateway@$EG_VERSION
+RUN yarn global add express-gateway@$EG_VERSION --ignore-optional
```
kevinswiber: I asked this before, but I'm not sure I got a response.  Why are we using `yarn` instead of `npm`?

Dockerfile-master
```diff
@@ -1,4 +1,4 @@
-FROM node:8-alpine
+FROM node:8
```
kevinswiber: Why are we not using the Alpine distribution for master builds?

kevinswiber:
> To be honest here I was wondering whether wouldn't it be better to simply drop bcrypt and have bcryptjs only. It is not a speed-critical part of the system (user creation) and it would remove a lot of hassle here.

@XVincentX Can you put it in a new issue? We can close this one out.  Thanks!

XVincentX:
https://github.com/ExpressGateway/express-gateway/issues/724

Dockerfile
kevinswiber: Ah, okay.  You can change the global path in NPM, though.  `npm config set prefix "/chicken/noodle/soup"`

XVincentX:
I do not really care that much about the package manager — but at the time it was not working with NPM. It's surely because I was missing something with Docker Users and permissions, but with yarn it was working out of the box, it's included in the Node Image so I went in this direction.

I'm ok revisiting it later!

PR #730, opened 2018-05-16
Remove bcrypt native dependency

XVincentX:
Need #726 nested first.


bin/generators/gateway/templates/basic/package.json
```diff
@@ -1,5 +1,8 @@
 {
-  "name": "",
+  "name": "express-gateway-instance",
```
kevinswiber: Doesn't the CLI overwrite this with the gateway name?

bin/generators/gateway/templates/basic/package.json
```diff
@@ -1,5 +1,8 @@
...
+  "description": "Express Gateway Instance Bootstraped from Command Line",
+  "repository": {},
+  "license": "MIT",
```
kevinswiber: Don't need a license here.  How the generated gateway gets licensed is up to the user.  Having this here could actually present an internal governance problem.

bin/generators/gateway/templates/getting-started/package.json
```diff
@@ -1,5 +1,8 @@
 {
-  "name": "",
+  "name": "express-gateway-instance",
```
kevinswiber: Same as the other template.

bin/generators/gateway/templates/getting-started/package.json
```diff
@@ -1,5 +1,8 @@
...
+  "description": "Express Gateway Instance Bootstraped from Command Line",
+  "repository": {},
+  "license": "MIT",
```
kevinswiber: Same as the other template.

bin/generators/gateway/templates/basic/package.json
XVincentX: If so, then it's overwritten in the wrong moment, and npm is throwing a warning. I need to verify this.

bin/generators/gateway/templates/basic/package.json
XVincentX: Npm will throw a warning without. In this case, maybe the best thing would be to ask the license using the CLI?

kevinswiber: We could use `UNLICENSED` or something.

bin/generators/gateway/templates/basic/package.json
kevinswiber: It's supposed to happen here: https://github.com/ExpressGateway/express-gateway/blob/master/bin/generators/gateway/create.js#L121

XVincentX: The code seems correct, the change is happening *before* the install process.

I can dig into more tomorrow.

PR #733, opened 2018-05-20
adding url prefix for default login route

taewdy:
Hi, 
I needed this to use '/login' url for myself.
Plz have a look. Thanks

### DIFF
--- a/lib/config/schemas/system.config.json
+++ b/lib/config/schemas/system.config.json
@@ -118,6 +118,10 @@
     },
     "plugins": {
       "type": "object"
+    },
+    "oauth2": {
+      "type": "object"
+      ,"default": {}
     }
   },
   "required": [
--- a/lib/policies/oauth2/oauth2-routes.js
+++ b/lib/policies/oauth2/oauth2-routes.js
@@ -28,9 +28,10 @@ module.exports = function (app, config) {
     passport.initialize(),
     passport.session()
   ];
-  app.get('/login', site.loginForm);
-  app.post('/login', middlewares, site.login);
-  app.get('/logout', site.logout);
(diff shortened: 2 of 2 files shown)


XVincentX:
Hey,

thanks for the PR.

Unfortunately I'm not able to understand nor the scope, what's trying to solve and so on.

If you could provide a little bit of more context, I can probably tackle down this and see if we can solve the issue without having to modify the gateway or concordate a change together.

Thanks!

XVincentX:
@taewdy I'm going to close this. I'll be more than happy to go through once you'll clarify the use case and the reason behind the PR. Thanks!

theothermattm:
@XVincentX I believe I hit this same issue (even though the original description is vague).  To provide some more details:  I need to proxy endpoints that use `/login` in their path, and at the express gateway API level, I also want to use the root `/login`. But, what's happening is that express-gateway is intercepting those for it's own use.  Looking through the code briefly it looks like the login route is hardcoded in if you're using the `oauth` policies.  I was able to disable the oauth policy and get around it for now.  If you're able to configure the login/logout URL's this would likely get around this problem if you need to use the built in oauth policies.

XVincentX:
Oh ok I understand what the gentleman meant now. 

PRs are welcomed, I have no bandwidth to make this happen at the moment.

theothermattm:
@XVincentX I might be up for that but if you have any guidance that would be good.  The PR linked to this issue is similar to what I'd implement, but perhaps there are other places that need to be touched as well?  I haven't looked deeply into the login functionality. I think i would just allow for a fully configurable login URL.

XVincentX:
That's the same way I would do. Add a configuration parameter in the schema and then take it back and use it in the oauth2 policy. Do not forget to put it to `/login` as default value

PR #747, opened 2018-06-11
Introduce oAuth2 introspection support

### LINKED ISSUE #703: OAuth2: add third party introspection URL
https://tools.ietf.org/html/rfc7662

XVincentX:
The following PR will introduce a new policy called `oauth2-introspection`.

The OAuth 2.0 Introspection Policy checks for token validity using the **introspection** endpoint, following the [RFC-7662 standard](https://tools.ietf.org/html/rfc7662).

### Implementation notes
1. Removed an huge chunk of comments that are unhelpful.
2. I put in our codebase parts of `node-token-inspector` because the original library is using async/await as well as a library conflicting a bit with ES6/Token. I made a pull request that fixes the issue which has been merged also. However, that module was doing way more than needed (checking for JWTKS as well)
3. I had some troubles because the `urlEncoded` body parser is (rightly) reading the request stream which, once it reaches the end, is not readable anymore. For that reason, the proxy policy was hanging and not working correctly. Basically we'd need to restream the content here as well but I found this [great part](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-incoming.js#L121) that basically lets us override the sent content. So, through a `PassThrough` stream, I'm copying the data in memory while parsing it and once ready, I'll simply use such property to send the data back to the underlying service. This should help us with `bodyModifier` policy as well (see the relative branch).

I choose `bodyStream` as a property to indicate something that the proxy server will send as body instead of the original request content. Let me know in case you disagree with the property name.
P.S: Maybe I should add dedicated tests for this feature.

Connect #703 
Closes #703 

#### Outstanding issues
- [x] https://github.com/schibsted/node-token-introspection/issues/13
- [x] `node-token-introspection` is using `async/await` that Node6 does not support
- [x] Get answers from Auth0 to have a test `introspect` compliant url
- [x] Tests  

## Note
In case you're good with the changes please do not merge, just approve.


lib/policies/oauth2/oauth2-server.js
```diff
@@ -10,23 +10,8 @@ const authCodeService = services.authorizationCode;
 const authService = services.auth;
 
 const expiresIn = config.systemConfig.accessTokens.timeToExpiry / 1000;
-
-// Create OAuth 2.0 server
 const server = oauth2orize.createServer();
 
-// Register serialialization and deserialization functions.
```
DrMegavolt: I suggest reverting comments. They explain magic in this example

DrMegavolt:
1. I suggest reverting comments. 
2. why would you need to parse something and restream it? As I understand you just need to get token, send a request to third-party service to validate the token. And if the token is valid stream request as in standard proxy policy. Why is this trick needed?

XVincentX:
> I suggest reverting comments.

Done.


> Why would you need to parse something and restream it? As I understand you just need to get token, send a request to third-party service to validate the token. And if the token is valid stream request as in standard proxy policy. Why is this trick needed?

Very good question, I've been fighting with this for at least 2-3 hours. Here's the story:

`IncomingMessage` (that is the `req` object) is a `Readable` stream. When you read it, the content is basically the request body.

When using `bodyParser.urlEncoded()` to parse the request body, the bodyParser reads the stream which, once it's over, can't be re-read. For this reason, when the `node-http-proxy` tries to pipe the request stream to the `proxyReq` object, the request stream results to be already read to the end — so no body gets forwarded but, as the `Content-Lenght` header is still set to the original size, the receiving downstream service can either wait forever (because it's expecting a body) — or either respond anyway with something.

In any case, the original request body does not get delivered.

The way I resolved this was to create a `PassThrough` stream (a stream that sends back what's sent) and use this one as the new body to send, through the `buffer` property.

DrMegavolt:
the question is why do you need to parse body. `bodyParser.urlEncoded()`

As I understand the flow is the following: 

1. Somehow obtain oauth2 token from 3rd party
2. Make a request to API protected by EG with the token in the header with body
3. Verification stage:
   3.a EG reads token from the header  (Does not read body, it is still a stream)
   3.b EG sends separate request to 3rdparty to check the token
   3.c If not valid return 401\403
   3.d continue
4. Calls unmodified proxy policy which will manage `req` directly

At which point I need body parsing?

XVincentX:
@DrMegavolt The flow you mentioned is correct, but the token is not in the header.

The oAuth 2.0 introspection endpoint requires the `token` and `token_type` fields to be in the request body encoded as application/x-www-form-urlencoded.

See the [RFC Section 2.1](https://tools.ietf.org/html/rfc7662#section-2.1) for more details on the request shape.

DrMegavolt:
`oAuth 2.0 introspection endpoint ` is the `/introspect` url spec is describing. I called it a 3rd party. 

this is what *3.b* does. You somehow extract token from request going to the protected resource. then form *new* request to  `my3rdparty.oauth2.server/introspect`, parse response and depending on result proxy or not proxy original request

step 2 will look like 
Headers:
`Authorization: bearer 2YotnFZFEjr1zCsicMWpAA`
`Content-type: ..json`
Body:
{"hello":true}


3.a 3b will make a new request 
POST
URL `my3rdparty.oauth2.server/introspect`
Body `token=2YotnFZFEjr1zCsicMWpAA`
Content-type - url-encoded 

Apigee calls steps in 3.* as ServiceCallOut
https://docs.apigee.com/api-platform/security/oauth/use-third-party-oauth-system

XVincentX:
@DrMegavolt Ah I see what you mean!

If I understand correctly you propose to provide the token to the gateway using an Header anyway, and then move these values in the body of the request that the gateway will do to the `/introspect` endpoint, right?

Hmm, sounds reasonable. At first glance I thought to replicate the same mechanism of the `/introspect` endpoint but perhaps your idea is better.

The question would be — how does one provide the `token_type` field? A custom header? `X-Token-Type`?

DrMegavolt:
*optional* token_type:  If I pass in Authorization header `bearer ...` it means access_token automatically. 
And as I understand EG would not handle refresh tokens etc. This would be clients concern. 

So we can always assume `token_type` as `access_token` or even skip this parameter. 

And check the ServiceCallOut policy Apigee. this can be a generalization of call to service. 

---

Nevertheless, I still believe that supporting arbitrary passport policy would be by far superior approach and enable all kind of third-party auth support. 

---

@kevinswiber please check the conversation probably you'll have some additional input

XVincentX:
> And as I understand EG would not handle refresh tokens etc. This would be clients concern.

Ok it makes sense.

> I still believe that supporting arbitrary passport policy would be by far superior approach and enable all kind of third-party auth support.

Sure — one approach does not preclude the other. If people want to use the `introspect` endpoint, good for them. If not, I think we'll offer the mechanism you're talking about in the (near?) future

> We didn't touch scopes

Correct — this is something I haven't handled/thought about yet. 

Thanks for all the inputs!

XVincentX:
@DrMegavolt 

- I've changed the code so the source token comes from the `Authorization` header (`Bearer`)
- I've added scopes support according to the RFC — that is — `scopes` property with space separated strings.
- Modified the tests to cover both things here

Enjoy!

lib/policies/oauth2-introspect/oauth2-introspect.js
```diff
@@ -0,0 +1,42 @@
...
+      }
+    };
+
+    if (scannedTokens[accessToken] && ((Date.now() - scannedTokens[accessToken].lastCheck) / 1000) < actionParams.ttl) {
+      return scopeCheck(scannedTokens[accessToken].tokenData, done);
+    }
+
+    tokenIntrospection(accessToken, 'access_token')
+      .then(tokenData => {
+        scannedTokens[accessToken] = { lastCheck: Date.now(), tokenData };
```
XVincentX: To be honest I'm not entirely sure if it's cool to keep this data in memory //cc @kevinswiber @DrMegavolt

lib/policies/oauth2-introspect/oauth2-introspect.js
```diff
@@ -0,0 +1,42 @@
...
+  const scannedTokens = [];
+  const tokenIntrospection = tokenIntrospectionGenerator(actionParams);
+
+  passport.use('bearer-introspect', new BearerStrategy({ passReqToCallback: true }, (req, accessToken, done) => {
+    const requestedScopes = req.egContext.apiEndpoint.scopes;
+
+    const scopeCheck = (tokenData, done) => {
+      const avaiableScopes = tokenData.scopes ? tokenData.scopes.split(' ') : [];
+
+      if (!requestedScopes.some(requestedScope => {
```
DrMegavolt: Are you trying to check this condition `requestedScopes.every(avaiableScopes.includes)` ?

`requsted [r, w] with available [r,w,admin]` -> ok because r and w both included in `available` 

`requsted [r, w] with available [r,admin]` -> not ok because w is missing in `available`

lib/policies/oauth2-introspect/oauth2-introspect.js
DrMegavolt: it's ok as long as the cache expiration (ttl) is reasonable. we can always iterate later  to include redis or file storage

DrMegavolt:
Looks better now, see comments above. 
So next point is what is user_id in this case? 
and what will be inside `req.user`

lib/policies/oauth2-introspect/oauth2-introspect.js
XVincentX: Yeah, that's the intention. Let me try your code.

XVincentX:
@DrMegavolt `req.user` will be the content of the returned deciphered `tokenData`, whose content is unknown (it depends on the oAuth2 server).

Same with the user_id I guess.

PR #753, opened 2018-06-18
Introduce the body-modifier policy — connect LunchBadger/marketing#13

XVincentX:
The following PR is a proposal to have a request/response body modifier in the core.

This feature has been required by multiple users and it should help a lot people to move forward and adopting the gateway.

## Example usage

```yaml
body-modifier:
  action:
    request:
      add:
      - name: fullname
        value: req.body.name + ' ' + req.body.surname
      remove:
      - name
      - surname
    response:
      add:
      - name: createdBy
        value: req.user.id
      remove:
      - uselessParam
```

You can see that basically you have a `add` and `remove` property. The first one is an array of object containing `name` and `value` —  the value is evaluated as a code property, so you can grab data from the `req` object.

The remove property is a simple string array that'll be simply removed. 

## Outstanding issues
- [x] Tests
- [ ] The current code is trying to first parse the body as json, then as form/url encoded. I'm not too sure if we should go with this brute force approach or maybe set this as a configuration property. You could tell in the policy to parse the body as JSON or as UrlEnconded and restream it as something else.
- [ ] Is `body-modifier` a meaningful name?
- [x] Request/Response transformation logic is the same. Try to DRY it.
- [ ] Maybe we can use destructuring and `object.assign` instead of `delete`?

### DIFF
--- a/lib/policies/body-modifier/index.js
+++ b/lib/policies/body-modifier/index.js
@@ -47,7 +47,7 @@ module.exports = {
 
           if (params.request) {
             params.request.add.forEach(addParam => { req.body[addParam.name] = req.egContext.run(addParam.value); });
-            params.request.remove.forEach(removeParam => { delete req.body[removeParam.name]; });
+            params.request.remove.forEach(removeParam => { delete req.body[removeParam]; });
 
             req.egContext.requestStream = new PassThrough();
             const bodyData = JSON.stringify(req.body);
@@ -62,7 +62,7 @@ module.exports = {
                 const body = JSON.parse(data);
 
                 params.response.add.forEach(addParam => { body[addParam.name] = req.egContext.run(addParam.value); });
-                params.response.remove.forEach(removeParam => { delete body[removeParam.name]; });
+                params.response.remove.forEach(removeParam => { delete body[removeParam]; });
 
                 const bodyData = JSON.stringify(body);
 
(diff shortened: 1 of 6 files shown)


kevinswiber:
This is good.  We should document `req.egContext.requestStream` as a separate feature.  Middleware could potentially replace this with their own custom stream.

lib/policies/body-modifier/index.js
```diff
@@ -0,0 +1,82 @@
...
+          if (err) return next(err);
+
+          if (params.request) {
+            const bodyData = JSON.stringify(transformBody(params.request, req.egContext, req.body));
+            req.headers['content-length'] = Buffer.byteLength(bodyData);
+            req.egContext.requestStream = new PassThrough();
+            req.egContext.requestStream.write(bodyData);
+          }
+
+          if (params.response) {
```
kevinswiber: Feels a little weird to have a `requestStream` but no `responseStream`.

XVincentX:
> We should document req.egContext.requestStream as a separate feature. Middleware could potentially replace this with their own custom stream.

I agree. The only caveat is that you need to remember to reset the `Content-Lenght` header for the request, or the things might go wrong. If we do a really good job in documenting this, we might really have a killer feature here.

XVincentX:
> not possible to dynamically determine the `Content-Length` after body manipulation right?

XVincentX:
Closed until we figure out what we want to do with this.

PR #757, opened 2018-06-19
Fix plugin schema validation.

ouvtk:
Will throw an error ("Failed to validate settings") in case if settings validation fail. Therefore will stop loading of current plugin.
Will pass schema validation in case of malformed schema itself.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -33,18 +33,11 @@ module.exports.load = function ({ config }) {
       }
 
       // register schema and validate settings
-      let validate;
-      try {
-        validate = schemas.register('plugin', pluginName, plugin.schema);
-      } catch (e) {
-        logger.error(`Wrong schema definition for ${pluginName}`, e);
-      }
-
-      if (validate !== undefined) {
-        const {isValid, error} = validate(settings);
-        if (!isValid) {
-          throw new Error('Failed to validate settings: ' + error);
-        }
+      const validate = schemas.register('plugin', pluginName, plugin.schema);
+      const { isValid, error } = validate(settings);
+      if (!isValid) {
+        // Stop loading plugin.
+        throw new Error(`Failed to validate settings: ${error}`);
       }
 
       const services = require('./services');
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -33,10 +33,18 @@ module.exports.load = function ({ config }) {
       }
 
       // register schema and validate settings
+      let validate;
       try {
-        schemas.register('plugin', pluginName, plugin.schema)(settings);
+        validate = schemas.register('plugin', pluginName, plugin.schema);
       } catch (e) {
         logger.error(`Wrong schema definition for ${pluginName}`, e);
+      }
+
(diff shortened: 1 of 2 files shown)


lib/plugins.js
```diff
@@ -33,12 +33,20 @@ module.exports.load = function ({ config }) {
...
       // register schema and validate settings
+      let validate;
       try {
-        schemas.register('plugin', pluginName, plugin.schema)(settings);
+        validate = schemas.register('plugin', pluginName, plugin.schema);
       } catch (e) {
         logger.error(`Wrong schema definition for ${pluginName}`, e);
       }
 
+      if (validate !== undefined) {
```
XVincentX: Do not check for such value — I think it's a bad fallback and a situation that should never happen. I'd rather crash as soon as possible and checkout what's wrong instead of silently swallowing this exception.

lib/plugins.js
```diff
@@ -33,12 +33,20 @@ module.exports.load = function ({ config }) {
...
+        const {isValid, error} = validate(settings);
```
XVincentX: Nit: please put spaces
```diff
- const {isValid, error}
+ const { isValid, error }
```

lib/plugins.js
```diff
@@ -33,12 +33,20 @@ module.exports.load = function ({ config }) {
...
+        if (!isValid) {
+          throw new Error('Failed to validate settings: ' + error);
```
XVincentX: Nip: please use ES6 literal strings.
Also, make sure to log the error (`logger.error`)

ouvtk:
This is what I'm end up with:
- Added unit-test to cover whole plugin loading process (both successful and failed), but focusing on schema validation. Basically covering plugins.js::load method only.
- Plugin load will fail now in any unexpected case, error will be caught locally and logged.
- Styling.

test/plugins/schema.test.js
```diff
@@ -0,0 +1,53 @@
+const path = require('path');
+const should = require('should');
+
+const pluginsLoader = require('../../lib/plugins');
+
+// Caching fixture plugin, but redefining its properties for testing plugin load.
+const pluginName = 'express-gateway-plugin-test';
+const pluginDirectory = path.join(__dirname, '../fixtures', pluginName);
+const testPlugin = require(pluginDirectory);
+testPlugin.schema = {
```
XVincentX: Nit: please put a new line here

test/plugins/schema.test.js
```diff
@@ -0,0 +1,53 @@
...
+  '$id': `http://express-gateway.io/schemas/plugin/${pluginName}.json`,
+  'required': ['schema-test-param']
+};
+testPlugin.init = (pluginContext) => {
```
XVincentX: Nit: put a new line here.

XVincentX:
Thank you very much for the contribution @ouvtk — in particular for taking the time to test this piece properly.

ouvtk:
@XVincentX Thank you for your guidance. Was a pleasure to help

PR #762, opened 2018-06-27
Make the CLI follow the nextKey for app and users

### LINKED ISSUE #749: eg apps list and eg users list return incomplete results or empty
It seems that the SCAN in the findAll method for both the application dao and the user dao only does a single call instead of a full iteration which works when testing but not when your database grows outside of the default query range for SCAN.

The SCAN should be called again and again until the return cursor is 0 to guarantee the search is complete (called a 'full iteration'). From the [docs](https://redis.io/commands/scan):

> As you can see the SCAN return value is an array of two values: the first value is the new cursor to use in the next call, the second value is an array of elements.
> 
> Since in the second call the returned cursor is 0, the server signaled to the caller that the iteration finished, and the collection was completely explored. Starting an iteration with a cursor value of 0, and calling SCAN until the returned cursor is 0 again is called a full iteration.

I will put in a PR for what is working as a fix in our environment.

XVincentX:
This PR introduces an additional flag for `eg apps list` and `eg users list` called `-a` or `--all`. When this is provided, the cli will navigate through the paged results from Redis in order to show all the users and not just the first page.

Connect #749 
Closes #749 

P.S: The CLI and the tests for the CLI have *a lot* of copy/paste code that could be removed and reused across the whole part of the system. However doing this now would be a significant amount of time. I still want to do that, but maybe in another iteration.

### DIFF
--- a/admin/users.js
+++ b/admin/users.js
@@ -35,13 +35,13 @@ module.exports = function (client) {
     list (params) {
       let results = [];
 
-      const fetchNext = (res, queryParams = {}) => {
+      const fetchNext = (res) => {
         results = results.concat(res.body.users);
         if (res.body.nextKey !== 0) {
           return client
             .get(baseUrl)
-            .query(Object.assign(queryParams, { start: res.body.nextKey }))
-            .then(res => fetchNext(res, queryParams));
+            .query(Object.assign({}, params, { start: res.body.nextKey }))
+            .then(fetchNext);
         }
         return { users: results };
       };
@@ -49,7 +49,7 @@ module.exports = function (client) {
       return client
         .get(baseUrl)
         .query(params)
-        .then(res => fetchNext(res, params));
+        .then(fetchNext);
     },
 
     remove (id) {
--- a/admin/users.js
+++ b/admin/users.js
@@ -33,12 +33,23 @@ module.exports = function (client) {
         .then(res => res.body);
     },
     list (params) {
+      let results = [];
+
+      const fetchNext = (res, queryParams = {}) => {
+        results = results.concat(res.body.users);
+        if (res.body.nextKey !== 0) {
+          return client
+            .get(baseUrl)
+            .query(Object.assign(queryParams, { start: res.body.nextKey }))
+            .then(res => fetchNext(res, queryParams));
+        }
+        return { users: results };
+      };
+
       return client
         .get(baseUrl)
         .query(params)
-        .then(res => {
-          return res.body;
-        });
+        .then(res => fetchNext(res, params));
     },
 
     remove (id) {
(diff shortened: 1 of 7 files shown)


bin/generators/apps/list.js
```diff
@@ -10,12 +10,13 @@ module.exports = class extends eg.Generator {
       builder: yargs =>
         yargs
           .usage(`Usage: $0 ${process.argv[2]} list [options]`)
+          .boolean('a').alias('a', 'all').describe('a', 'Show all the elements instead of stopping to the fist page')
```
kevinswiber: Suggest change in description:
> List all apps

bin/generators/users/list.js
```diff
@@ -10,11 +10,12 @@ module.exports = class extends eg.Generator {
       builder: yargs =>
         yargs
           .usage(`Usage: $0 ${process.argv[2]} list [options]`)
+          .boolean('a').alias('a', 'all').describe('a', 'Show all the elements instead of stopping to the fist page')
```
kevinswiber: Suggest change in description:
> List all users.

XVincentX:
I agree with your points — the CLI is way behind what the API offers. I do hope we'll have some time to modernize the CLI and catch up with the API features.

bin/generators/apps/list.js
XVincentX: Done in f42d8d1

bin/generators/users/list.js
XVincentX: Done in f42d8d1

XVincentX:
@kevinswiber I addressed the comments. If there's anything else here, we could bring this in.

XVincentX:
@DrMegavolt @kevinswiber Ping

admin/users.js
```diff
@@ -32,13 +32,24 @@ module.exports = function (client) {
         .get(`${baseUrl}${encodeURIComponent(id)}`)
         .then(res => res.body);
     },
-    list (params) {
+    list (params = {}) {
+      let results = [];
+
+      const fetchNext = (res) => {
```
DrMegavolt: why do you need this page by page loader if you can request all users\apps at once 
by passing `count` parameter as something big like Number.MAX_SAFE_INTEGER 
https://github.com/ExpressGateway/express-gateway/blob/f42d8d1fe61c74779a70ede452255583fa6a430f/lib/services/consumers/application.dao.js#L56

XVincentX: Using a large number in theory guarantees that I get all the users — in practice we can't never be completely sure. We do not know how large the data base is.

Also — if I recall correctly the count is about the number of fields, not the number of records.

I do not feel really comfortable in making a single call with a large number and cross the fingers that all will work accordingly. If it's a paginated resource, we should handle pagination.

@kevinswiber Thoughts?

PR #763, opened 2018-06-28
Add requestStream property to the context. Connect LunchBadger/general#381 closes LunchBadger/general#381

XVincentX:
This PR will forward the `requestStream` property to the http proxy, if provided. It would enable interesting plugins manipulation tricks.

### DIFF
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -83,7 +83,7 @@ module.exports = function (params, config) {
 
     logger.debug(`proxying to ${target.href}, ${req.method} ${req.url}`);
 
-    proxy.web(req, res, { target, headers });
+    proxy.web(req, res, { target, headers, buffer: req.egContext.requestStream });
   };
 
   // multiple urls will load balance, defaulting to round-robin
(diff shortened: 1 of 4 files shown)


lib/policies/proxy/proxy.js
```diff
@@ -83,7 +83,7 @@ module.exports = function (params, config) {
 
     logger.debug(`proxying to ${target.href}, ${req.method} ${req.url}`);
 
-    proxy.web(req, res, { target, headers });
+    proxy.web(req, res, { target, headers, buffer: req.egContext.requestStream });
```
kevinswiber: There's no type-checking here, and I'm not sure what error http-proxy might throw, but it would be good if it was informative actionable by the user in an Express Gateway context.  Still, going to move it forward for now.  Thanks!

XVincentX: Good point. Basically, the code is `(buffer || req).pipe()` — so I guess it'd be a runtime error or something like "the object does not have `pipe` function"

XVincentX: Confirmed, it's exactly the error that you'd see in case the provided object is not a `Stream`

PR #777, opened 2018-08-07
Add strip path capabilities to Proxy — Connect LunchBadger/general#425

XVincentX:
This PR adds a `stripPath` option to the proxy policy that, when activated, will effectively strip out the matched part from the target url.

This is particularly useful when proxying something to something else and you want to both own the url space as well as preserving the query string parameters and the following path parts when proxying to the selected service endpoint.

### DIFF
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -77,13 +77,13 @@ module.exports = function (params, config) {
 
   const balancer = createStrategy(strategy, proxyOptions, urls);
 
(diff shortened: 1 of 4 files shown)


lib/policies/proxy/proxy.js
```diff
@@ -77,10 +77,14 @@ module.exports = function (params, config) {
 
   const balancer = createStrategy(strategy, proxyOptions, urls);
 
+  const stripPathFn = params.stripPath ? (req) => { req.url = `${req.params[0]}${req._parsedUrl.search || ''}`; } : () => { };
```
XVincentX: I'm open to better names in both function name *and* parameter name.

PR #779, opened 2018-08-08
Add passThrough parameter to oauth2-introspect policy and key-auth policy name fix

### LINKED ISSUE #781: keyauth should be key-auth in schemas
This
`http://express-gateway.io/schemas/policies/keyauth.json`
should be
`http://express-gateway.io/schemas/policies/key-auth.json`

### LINKED ISSUE #783: Add label TTL in oauth2-infrospect schemas
As we want to render this property as `TTL`, please add for `http://express-gateway.io/schemas/policies/oauth2-introspect.json` extra `label` field:
```
"ttl": {
  "type": "integer",
  "default": 60,
  "label": "TTL"
},
```

XVincentX:
Kristof, while syncinc the JSON Schemas for the UI discovered the missing passThrough in the oauth2-introspection policy.

This PR fixes that.

### Why is this bug happening?

Because we do not have a standard way to define auth policies. #746 is part of the solution.

---

Also

Connect #781 
Closes #781 

Connect #783 
Closes #783


lib/policies/key-auth/index.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
-  policy: require('./keyauth'),
+  policy: require('./key-auth'),
```
kevinswiber: The policy file is still named `keyauth.json`.

XVincentX: image
It has been renamed @kevinswiber — the Github diff doesn't really show it that clearly enough

XVincentX: Also – in case tests would pass with such thing, I'd really go and find another job 😱

kevinswiber: @XVincentX Oh, my bad.  I was browsing the code at the commit.  But instead of picking the last commit, I picked the first one, which had not included these changes.  We're good!

PR #782, opened 2018-08-09
Validate configuration on API Calls before disk saving

### LINKED ISSUE #778: Admin API should validate the provided data before writing it to the disk
Currently the Admin API is persisting the config changes on the disk directly, without validating them first.

So if you provide an invalid config change through the API, it will be persisted and then the hot reload method will trigger — validating the changes and refusing them.

The final result is that you've got an invalid config on the disk while running EG with the old config. On a server restart, the gateway won't be able to start at all.

Kudos to @kolbinski for spotting this.

### LINKED ISSUE #788: Error: policies should NOT have less than 1 items
On prod, if I deployed a gateway with no policies, admin api failed with error:
```
PUT on https://admin-gateway-kristofolbinski-dev.lunchbadger.io/pipelines/f9b1bc62-fc6f-4084-8929-41ff61376ea2
Error 422: data.pipelines['f9b1bc62-fc6f-4084-8929-41ff61376ea2'].policies should NOT have less than 1 items
```

Serhii published different EG version to fix this problem, and suggested it was most likely due to bug somewhere here:
https://github.com/ExpressGateway/express-gateway/pull/782/files#diff-5122ec46a4f7ff0f1909a379117ff89bR138

XVincentX:
Connect #778
Closes #778

The original intention of this pull request was to simply make sure that, when an Admin API call that is modifying the configuration is done, the validation should be triggered so that we do not persist on disk a broken configuration file.

It then become a dragoon because doing this uncovered several bad things in the codebase that I have decided to fix (see image below)

image

In particular:

1. I've removed the `preserveStructureOnUpdates` support. It's complicating the code for an undocumented flag that nobody is using with a sometimes-broken-parser-dumper.
2. Invoking the generic `gateway.config` file validator before persist the changes
3. Loop through all the proposed policy and invoke their validators one by one before persist the changes to the disk. This is not ideal, it's replicating a logic that's elsewhere already but isolating it in an reasonable amount of time it's tricky. I can do that, though.
4. Adding a `INVALID_CONFIG` error code so that the Admin Express App can detect such condition
6. The Admin API is returning an appropriate status code (`422`)

---

This new behavior made fail some tests because they were just sending strange/random payloads — that wouldn't pass validation. I've fixed these tests and during the thing I discovered that some JSON Schemas in `gateway.config` that I fixed.

Two points in particular:

1. We have got a circular dependency between `policy` and `config`. This is particularly bad, and needs to be fixed as soon as possible.
2. This thing that some parts can be both a single object or an array it's preventing us to use some interesting JSON Schema features. This needs to be fixed as a breaking change, if required.

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -118,16 +118,23 @@ class Config {
     const path = this[this.configTypes[type].pathProperty];
     return readFile(path, 'utf8').then(data => {
       let text;
+      let json;
       const isJSON = path.toLowerCase().endsWith('.json');
       if (this.systemConfig && this.systemConfig.preserveStructureOnUpdates && !isJSON) {
         const yawn = new YAWN(data);
-        yawn.json = modifier(yawn.json);
+        json = yawn.json = modifier(yawn.json);
         text = yawn.yaml;
       } else { // js-yaml can handle JSON as well as YAML if no need for structure\comments save
-        const json = yamlOrJson.load(data);
-        const result = modifier(json);
+        const currentJson = yamlOrJson.load(data);
+        const result = json = modifier(currentJson);
         text = yamlOrJson.dump(result);
       }
+      const { isValid, error } = this.configTypes[type].validator(json);
+
+      if (!isValid) {
+        throw new Error(error);
+      }
+
       return writeFile(path, text);
     });
   }
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -115,21 +115,22 @@ class Config {
   // due to fragile nature of YAWN it may not work properly in that mode.
   // safe way is to use it in rewrite file mode
   _updateConfigFile (type, modifier) {
-    const path = this[this.configTypes[type].pathProperty];
+    const configType = this.configTypes[type];
+
+    const path = this[configType.pathProperty];
     return readFile(path, 'utf8').then(data => {
       let text;
-      let json;
       const isJSON = path.toLowerCase().endsWith('.json');
       if (this.systemConfig && this.systemConfig.preserveStructureOnUpdates && !isJSON) {
         const yawn = new YAWN(data);
-        json = yawn.json = modifier(yawn.json);
+        yawn.json = modifier(yawn.json);
         text = yawn.yaml;
       } else { // js-yaml can handle JSON as well as YAML if no need for structure\comments save
-        const currentJson = yamlOrJson.load(data);
-        const result = json = modifier(currentJson);
+        const json = yamlOrJson.load(data);
+        const result = modifier(json);
         text = yamlOrJson.dump(result);
       }
-      const { isValid, error } = this.configTypes[type].validator(json);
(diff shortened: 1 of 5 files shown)


lib/config/config.js
```diff
@@ -111,24 +110,50 @@ class Config {
...
-    const path = this[this.configTypes[type].pathProperty];
+    const configType = this.configTypes[type];
+    const path = this[configType.pathProperty];
+
     return readFile(path, 'utf8').then(data => {
-      let text;
-      const isJSON = path.toLowerCase().endsWith('.json');
-      if (this.systemConfig && this.systemConfig.preserveStructureOnUpdates && !isJSON) {
-        const yawn = new YAWN(data);
-        yawn.json = modifier(yawn.json);
```
kevinswiber: I had not recalled that we introduced `preserveStructureOnUpdates`.  I thought we were still preserving it by default.  The YAML spec calls comments a "presentation detail," and most parsers/serializers don't maintain information regarding comments.  For more on this ugly wart, see here: http://yaml.org/spec/1.2/spec.html#id2766150

YAWN is the only Node.js-based YAML library that attempts to do this.  And it's a valid use case.  If we're not comfortable with the maturity of that project and we _are_ comfortable overwriting presentation details of the YAML configuration whenever Express Gateway automation touches the config files, this change is fine.  But let's explicitly draw that line in the sand before proceeding.

lib/config/config.js
```diff
@@ -111,24 +110,50 @@ class Config {
...
+            if (!isValid) {
+              throw new Error(error);
+            }
+          });
+        }
+
+        return writeFile(path, text);
+      } catch (err) {
+        log.error('Invalid pipelines configuration.', err);
+        err.code = 'INVALID_CONFIG';
```
kevinswiber: Should we reuse this logic in `loadConfig`?  In our create-gateway templates, we can also look for this error and display a warning instead of crashing the process with an uncaught promise rejection.

XVincentX: In theory I'd like to make the opposite — this part should be using the `loadConfig`. Unfortunately this is not possible in a timely manner because of the coupled code between load config and config source — which is probably going to be a different story.

I can try to reword this piece a little bit, but it might take a little bit of time before having something good; we can create a tracking card for that. Your call.

P.S: I'm using https://github.com/ExpressGateway/express-gateway/projects to track things I'd like to change in Express Gateway for the better which might result in a breaking change. Feel free to have a look into!

lib/config/config.js
kevinswiber: Let's yank it for now.  The Go YAML parser has done some work regarding presentation, but I think we're far off from other libraries getting there.  Most of them rely on the C implementation.  Is this something that would be considered a breaking change?

lib/config/config.js
kevinswiber: Okay, let's move forward in the interest of time.  Thanks!

PR #786, opened 2018-08-13
Add edge case handling for strip Path — Connect LunchBadger/general#425 Connect LunchBadger/general#428 Closes LunchBadger/general#425 Closes LunchBadger/general#428

XVincentX:
This PR will handle the edge case when the path specified in the ApiEndpoint is simply something like /api instead of /api*.

Tests included.

### DIFF
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -77,7 +77,13 @@ module.exports = function (params, config) {
 
   const balancer = createStrategy(strategy, proxyOptions, urls);
 
-  const stripPathFn = params.stripPath ? (req) => { req.url = `${req.params[0]}${req._parsedUrl.search || ''}`; } : () => { };
+  const stripPathFn = params.stripPath
+    ? req => {
+      req.url = req.params[0]
+        ? `${req.params[0]}${req._parsedUrl.search || ''}`
+        : '';
+    }
+    : () => { };
 
   return function proxyHandler (req, res) {
     const target = balancer.nextTarget();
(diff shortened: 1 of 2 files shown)


lib/policies/proxy/proxy.js
```diff
@@ -77,7 +77,13 @@ module.exports = function (params, config) {
...
   const balancer = createStrategy(strategy, proxyOptions, urls);
 
-  const stripPathFn = params.stripPath ? (req) => { req.url = `${req.params[0]}${req._parsedUrl.search || ''}`; } : () => { };
+  const stripPathFn = params.stripPath
+    ? req => {
+      req.url = req.params[0]
+        ? `${req.params[0]}${req._parsedUrl.search || ''}`
+        : '';
+    }
+    : () => { };
```
XVincentX: In case you're wondering how I came out with this formatting — I gave the code to Prettier.io and I pasted here the result :trollface:

lib/policies/proxy/proxy.js
```diff
@@ -77,7 +77,10 @@ module.exports = function (params, config) {
 
   const balancer = createStrategy(strategy, proxyOptions, urls);
 
-  const stripPathFn = params.stripPath ? (req) => { req.url = `${req.params[0]}${req._parsedUrl.search || ''}`; } : () => { };
+  const stripPathFn = params.stripPath
+    ? req => {
+      req.url = `${req.params[0] || ''}${req._parsedUrl.search || ''}`;
```
kevinswiber: `req.params[0] || ''`

Should this be `req.params[0] || '/'`?  I see that the test is passing, so `/` might be assumed somewhere, but I think `req.url = ''`--which is what this can appear to be--is confusing.

XVincentX: Tests pass with both things (`''` and `/`). I definitely recall there's some normalization somewhere but I can't remember where unfortunately. It should be good.

PR #791, opened 2018-08-15
Return 404 when service endpoint is not provided

XVincentX:


XVincentX:
I'll split this into multiple PRs

PR #806, opened 2018-08-27
Allow empty serviceEndpoint

### LINKED ISSUE #789: proxy policy: should allow empty serviceEndpoint

XVincentX:
Connect #789 
Closes #789

### DIFF
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -14,7 +14,7 @@ module.exports = function (params, config) {
   const serviceEndpointKey = params.serviceEndpoint;
 
   if (!serviceEndpointKey) {
-    return (req, res) => res.sendStatus(404);
+    return (req, res) => res.sendStatus(502);
   }
 
   const endpoint = config.gatewayConfig.serviceEndpoints[serviceEndpointKey];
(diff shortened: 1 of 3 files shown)


lib/policies/proxy/proxy.js
```diff
@@ -12,6 +12,11 @@ const createStrategy = (strategy, proxyOptions, endpointUrls) => {
 
 module.exports = function (params, config) {
   const serviceEndpointKey = params.serviceEndpoint;
+
+  if (!serviceEndpointKey) {
+    return (req, res) => res.sendStatus(404);
```
DrMegavolt: how about 
502 Bad Gateway

   The 502 (Bad Gateway) status code indicates that the server, while
   acting as a gateway or proxy, received an invalid response from an
   inbound server it accessed while attempting to fulfill the request.

no response from the nonexisting server is an invalid response, right?

XVincentX: Yes I think that makes sense. I'll change the returned status code

XVincentX: Added a warning.

> Most likely this is some misconfiguration issue

I can't believe we're making this happen :trollface:

PR #812, opened 2018-09-10
Insert JWT credential correctly so that CLI and AdminAPI can catch them

### LINKED ISSUE #759: [Bug] Issues with `eg credentials info` and `eg credentials list` and JWT creds + Redis
[Bug] 
- Version 1.9.1
- Containerized Redis

#### User & Credential creation
- Startup eg with Redis enabled
- Run `eg user create -p "username=system" -p "firstname=system" -p "lastname=system"`
  - Returns user rec with id, assume its exported in bash as ${eg_id}
- Run `eg credentials create -c system -t jwt`
  - Returns credential

If you check Redis at this point, you can see all of the records that were created including the JWT creds.

#### info and list fail when type is jwt 
- Run `eg credentials info system -t jwt`
  - Expected: credential
  - Returned: Empty response
- Run `eg credentials list -c system`
  - Expected: List with one entry
  - Returned: Empty response

#### info and list fail when using user id and type is jwt
- Run `eg credentials info ${eg_id} -t jwt`
  - Expected: credential
  - Returned: Empty response
- Run `eg credentials list -c ${eg_id}`
  - Expected: List with one entry
  - Returned: Empty response

I also tested this using the ID when creating the credential. No change from the observed behavior above.

I also tested this with type oauth2. List works. Info works if you use the id. Fails when you use the username.

XVincentX:
This PR will make sure that the JWT credentials will be shown correctly when asked through the Admin API.

Unfortunately this is yet another monkey-patch thing, but I have no better options at the moment.

Connect #759 
Closes #759

### DIFF
--- a/lib/services/credentials/credential.dao.js
+++ b/lib/services/credentials/credential.dao.js
@@ -111,7 +111,7 @@ dao.insertCredential = function (id, type, credentialObj) {
     return Promise.resolve(null);
   }
   const key = buildIdKey(type, id);
-  if (type === 'key-auth') {
+  if (type === 'key-auth' || type === 'jwt') {
     if (credentialObj.keyId) credentialObj.id = credentialObj.keyId;
     return Promise.all([
       // build relation consumerId -> [key1, key2]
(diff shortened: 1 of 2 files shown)


XVincentX:
Well it's not cool at all, but unfortunately there's no other way to make it work without refactoring half of the services code. 👎

PR #818, opened 2018-09-14
Logs chores

XVincentX:
1. Winston 3 has a better default log level `info`. Using that we can avoid using `console.log` in production code (which is not redirectable)
2. Given point 1, update all references in the code to use `verbose` to hide statements that do not matter
3. Give color to log context so we can differentiate between timestamp, context, level and message.
4. It turns out we have different functions that aren't used anywhere that are harming our general test coverage. I've deleted these. They're dealing with system.config reload that we do not support.

I'd love to restate a little bit all the log statements in future. This is a start, at least.

image

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -62,8 +62,6 @@ class Config {
 
   loadGatewayConfig () { this.loadConfig('gateway'); }
 
-  loadSystemConfig () { this.loadConfig('system'); }
-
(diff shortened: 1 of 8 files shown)


XVincentX:
There might be something wrong with the coverage; I've even added an additional test to cover a case that wasn't before, but still the code coverage is reported to be insufficient.

PR #823, opened 2018-09-26
Support regular expression on cors policy configuration 

### LINKED ISSUE #821: CORS policy configuration doesn't support Regular Expressions
CORS library support regular expression on Origin option.  So as js-yaml (yaml parser) by using tag !!js/regexp. However if you look at express-gateway/lib/policies/cors/index.js, the type only allows string, Boolean and array where array items are only strings. This in turn treats a regular expression value on the Origin as a literal string so the response header Access-Control-Allow-Origin value is the pattern itself.

For example: partial gateway config is

```
policies:
  cors:
    action: 
      origin: /\.domain\.com$/
      methods: 'GET, PUT, POST, DELETE, OPTIONS'
      allowedHeaders: ['Content-Type', 'Authorization']
      credentials: true
      optionsSuccessStatus: 200
```

The response header Access-Control-Allow-Origin value becomes the pattern itself '/\.domain\.com$' 

I can't do this
`origin: !!js/regexp /\.domain\.com$/`
It errors our saying the tag !!js/regexp is not recognized or not one of the acceptable types in the schema.    

In express-gateway/lib/config/config.js the line for js-yaml to load the config file is this 
`config = yamlOrJson.load(envReplace(fs.readFileSync(configPath, 'utf8'), process.env));`

According to the js-yaml documentation https://github.com/nodeca/js-yaml

load (string [ , options ])
Use with care with untrusted sources. The same as safeLoad() but uses DEFAULT_FULL_SCHEMA by default - adds some JavaScript-specific types: !!js/function, !!js/regexp and !!js/undefined.

FYI @XVincentX  thanks.

albertlinga:
This allows cors origin configuration to have regular expressions as values. 

Closes #821

### DIFF
--- a/lib/policies/cors/index.js
+++ b/lib/policies/cors/index.js
@@ -5,8 +5,8 @@ module.exports = {
     type: 'object',
     properties: {
       origin: {
-        type: ['string', 'boolean', 'array'],
-        items: { type: 'string' },
+        type: ['string', 'boolean', 'array', 'object'],
+        items: { type: ['string', 'object'] },
         default: '*',
         description: 'Configures the Access-Control-Allow-Origin CORS header.'
       },
--- a/lib/policies/cors/index.js
+++ b/lib/policies/cors/index.js
@@ -6,7 +6,17 @@ module.exports = {
     properties: {
       origin: {
         type: ['string', 'boolean', 'array', 'object'],
-        items: { type: ['string', 'object'] },
+        items: {
+          type: ['string', 'object'],
+          if: { type: 'object' },
+          then: {
+            instanceof: 'RegExp'
+          }
+        },
+        if: { type: 'object' },
+        then: {
+          instanceof: 'RegExp'
+        },
         default: '*',
         description: 'Configures the Access-Control-Allow-Origin CORS header.'
       },
--- a/lib/schemas/index.js
+++ b/lib/schemas/index.js
@@ -5,10 +5,9 @@ const ajv = new Ajv({
   useDefaults: true,
   coerceTypes: true
 });
+require('ajv-keywords')(ajv, 'instanceof');
 
 const registeredKeys = [];
-
-require('ajv-keywords')(ajv, 'instanceof');
 
 function register (type, name, schema) {
   /*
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "express-gateway",
-  "version": "1.12.1",
+  "version": "1.12.2",
```
XVincentX: Do not advance the `package.json` version, we take care of it when it is time to release it.

lib/policies/cors/index.js
```diff
@@ -5,8 +5,8 @@ module.exports = {
     type: 'object',
     properties: {
       origin: {
-        type: ['string', 'boolean', 'array'],
-        items: { type: 'string' },
+        type: ['string', 'boolean', 'array', 'object'],
+        items: { type: ['string', 'object'] },
```
XVincentX: This will do, but it is not exactly the solution I would like to employ. I think we should add yet another layer of validation on that to make sure the provided object is a `RegExp` object. Maybe we could use [instanceOf](https://github.com/epoberezkin/ajv-keywords#instanceof) plugin to make it happen?

package.json
albertlinga: Noted

lib/policies/cors/index.js
albertlinga: Sorry fat fingered resolve button. I understand, I have read js-yaml that we need to validate should we pass types other than the primitive ones.

albertlinga:
@XVincentX updated schema for appropriate regexp validation. Added more tests.

lib/schemas/index.js
```diff
@@ -8,6 +8,8 @@ const ajv = new Ajv({
 
 const registeredKeys = [];
 
+require('ajv-keywords')(ajv, 'instanceof');
```
XVincentX: This should be happening just after `ajv` has been instantiated.

albertlinga: Noted.

albertlinga:
@XVincentX refactored the instanceof initialization. Added negative test case for setting origin as an object other than a regexp.

PR #824, opened 2018-09-26
Replace policy from class to functions — Connect LunchBadger/general#487 Closes Connect LunchBadger/general#487

XVincentX:
This PR is a small code refactor to transform the current policy to be a singleton class to an object which exports 3 functions.

As we're using almost functions everywhere and not classes, I think unifying the things here and there is beneficial to lower the contribution barrier.

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -155,7 +155,7 @@ class Config {
 
         return writeFile(path, text);
       } catch (err) {
-        log.error('Invalid pipelines configuration.', err);
+        log.error(`Invalid pipelines configuration: ${err}`);
         err.code = 'INVALID_CONFIG';
 
         throw err;
--- a/lib/policies/index.js
+++ b/lib/policies/index.js
@@ -4,64 +4,60 @@ const path = require('path');
 const logger = require('../logger').policy;
 const schemas = require('../schemas');
 
-class Policies {
-  constructor () {
-    this.policies = {};
+const register = (policy) => {
+  const { name, schema } = policy;
+  const action = policy.policy;
 
-    schemas.register('internal', 'baseAuth', {
-      $id: 'http://express-gateway.io/schemas/base/auth.json',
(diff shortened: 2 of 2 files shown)


DrMegavolt:
There are no advanced concepts in current version. Class is standard ES6 feature, no difference to have instance of the class or export functions.

XVincentX:
I agree that this is not bringing any new feature or improvement. My aim was not that one indeed. My aim was to make the code less complex — which I feel the class construct brings.

When you see a class you mentally start to think to an internal state, multiple instances, maybe inheritance — all things that in this case aren't happening.

Reducing the cognitive load is important to onboard new people on the project and lower the contribution barrier.

This change, _alone_, won't change the things. Complexity is something you accumulate with small steps and dismantle with small steps. This — I think, is one of these.

@kevinswiber What's your thought?

kevinswiber:
I understand this in terms of consistency less than I do in terms of complexity.  

What I like about the class approach is that the `policies` array is mutable, which might be handy for test scenarios.  (I'm currently working on plugin testing, so maybe that's why this is top of mind.)  With the approach in this PR, that `policies` list is inaccessible.

I agree with closing this for now.

XVincentX:
@kevinswiber The `policies` array is exposed [here](https://github.com/ExpressGateway/express-gateway/pull/824/files#diff-6dd40befa5e020684f04fdd9e88783b9R63) — or am I missing something?

kevinswiber:
@XVincentX Ah, nope.  I missed something!  In that case, I'd be in favor of this PR for consistency reasons.  I don't think the class adds any complexity, but if it's one of few (if any) classes in the codebase, that's a bit weird.

kevinswiber:
@DrMegavolt Can you agree to this PR for consistency reasons (as opposed to complexity)?

DrMegavolt:
ok

PR #826, opened 2018-09-27
 Remove custom keyAuth policy — connect LunchBadger/general#486

XVincentX:
Closes LunchBadger/general#486


DrMegavolt:
@XVincentX @kevinswiber  EG supports custom schema for api key in header. In proposed plugin you can set the name of the header, but it expects that entire value would be a key. 

In current version you can set 
`options.apiKeyHeaderScheme` 

`Authorisation: my_magic_prefix_for_api_key the_key`

also it allows to disable specific sources. 
`disableHeadersScheme`
`disableQueryParam`

I agree that unlikely this is something people are using that often. But switching to proposed plugin would mean removing those options and a breaking change.

XVincentX:
Yes, this would be a breaking change. I was hoping to make it retro compatible but unfortunately I haven't succeeded. I guess this is a 2.0 change.

PR #831, opened 2018-10-17
initial template configuration to include rewrite plugin - connect LunchBadger/general#496

DrMegavolt:


DrMegavolt:
It is not mandatory to update system.config. it is only if you need rewrite, so if they miss it it's fine 
The good part is that people will get hands on experience adding plugin. 
Doc update would be needed. 

>Also I do not really know what could happen if somebody installed
Latest will be used, but you can specify exact version in the config file if you want oldest for some reason.

policy from plugin is no different from internal 
 https://github.com/ExpressGateway/express-gateway/blob/30b0bf12f8b62b2f08d633bf730a736137699968/lib/gateway/index.js#L16

https://github.com/ExpressGateway/express-gateway/blob/e2055e921debd6c5819457e3103e0a11ed32ad47/lib/policies/index.js#L60

both just call `register` method. Only security related policies use consumer management in some hackie way.  

> create a new internal policy called rewrite
at some point I would rather extract policies as plugins rather than hardcoding them in EG

PR #832, opened 2018-10-17
Update dependencies

XVincentX:
Updating all dependencies before the minor release.

*NOTE:* The rate limit middleware has changed its behavior significantly — it has been split into two packages. I'll save this update for the next minor.


XVincentX:
Hold on — tests aren't being ran at all.

PR #840, opened 2018-11-07
Add userid to oauth2 jwt

duongvanba:


XVincentX:
@duongvanba This pull request has no description and no related issue. I basically do not understand what problem you're trying to solve as well as any test coverage.

In the current state, there's nothing I can do with this. Please elaborate a little bit more so I can have some context and then we'll move forward.

Thanks!

PR #842, opened 2018-11-08
Add request/response modifier policy

XVincentX:
This PR will add a new policy in the core called `modifier` that offers the following features:

1. Modify the request headers before they get sent to the upstream server
2. Modify the request body (if JSON) before they get sent to the upstream server
1. Modify the response headers before they get returned to the client
2. Modify the response body (if JSON) before they get returned to the client

The syntax to use such policy is currently like this:

```yaml
modifier:
- action:
    request:
      headers:
        add:
          r-test: 'baffino'
    response:
      body:
        add:
          hello: 'world'
        remove:
        - url
      headers:
        add:
          res: 'correct'
        remove:
        - x-test
```

Note: the left expression is executed on the `egContext`, so you can do `req.params.otherThing`; in general any evaluable expression will work.

```yaml
add:
  fullname: req.body.name + ' ' + req.body.surname
```

### Outstanding parts
* Missing complete code coverage
* Settle down on policy parameters
* Come up with the right policy name
* Figure out if request/response should be all in the same policy, or split these into two policies (`requestModifier`,`responseModifier`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,13 +7,7 @@ jobs:
 
     steps: &eg-checkout-install
       - checkout
-      - restore_cache:
-          key: dependency-cache-{{ checksum "package.json" }}
       - run: npm ci
-      - save_cache:
-          key: dependency-cache-{{ checksum "package.json" }}
-          paths:
-            - ./node_modules
       - run: npm test
 
   "node-10":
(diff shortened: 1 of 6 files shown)


DrMegavolt:
This is very limited, what if I want to rename property, join two properties in one , somehow change value ?

XVincentX:
@DrMegavolt Fail point. Can you list all the requirements this policy should have with some examples? That'd help to clarify what else should be included.

PR #845, opened 2018-11-21
Add support for hostname binding for http/https gateway

### LINKED ISSUE #843: Cannot configure bind address for http/https gateway server
(Note: I reported this on gitter and @kevinswiber suggested I file a bug.)

screen shot 2018-11-13 at 7 11 05 pm

There's a way to specify the bind address for the admin http server (via admin->hostname in gateway.config.yml) but you can't currently configure the bind address for the main gateway http/https servers. This is specifically problematic when trying to run in a Docker container where you need it to bind to all interfaces (0.0.0.0) when running in "bridge" network mode, for example.

I took a stab at filing a PR for this. While adding support for the new config value is trivial, the problem I ran into is maintaining backwards compatibility such that folks that currently don't have the hostname defined get the same behavior (binds to '::' for IPv6, etc), but when I *do* specify the hostname, the tests break with errors like:

```
  5) exact host name configuration
       should not serve sub.abc.acme.com :
     TypeError: Cannot read property 'address' of undefined
      at Test.serverAddress (node_modules/supertest/lib/test.js:55:18)
      at new Test (node_modules/supertest/lib/test.js:36:12)
      at Object.obj.(anonymous function) [as get] (node_modules/supertest/index.js:25:14)
      at prepareScenario (test/common/routing.helper.js:16:35)
      at Context.<anonymous> (test/common/routing.helper.js:73:30)

```

I *think* this is because node's `net.server.address()` returns an object if you don't specify a hostname (current behavior) or a String if you do (new behavior). (See https://nodejs.org/api/net.html#net_server_address), but I can't tell if this is just an issue with the tests, or if the rabbit hole goes deeper.

This may be easier than I think and I'm just not familiar with the code base. Any pointers in the right direction so I can file a PR would be much appreciated. Thanks!

XVincentX:
Connect #843
Closes #843

### DIFF
--- a/lib/gateway/index.js
+++ b/lib/gateway/index.js
@@ -20,12 +20,19 @@ module.exports = function ({ plugins, config } = {}) {
   const { httpServer, httpsServer } = bootstrap({ plugins, config });
 
   if (config.gatewayConfig.http && httpServer) {
+    // config.gatewayConfig.http.hostname = config.gatewayConfig.http.hostname || 'localhost';
     appPromises.push(
       new Promise(resolve => {
-        const runningApp = httpServer.listen(config.gatewayConfig.http.port, () => {
-          const { address, port } = runningApp.address();
+        const runningApp = httpServer.listen(config.gatewayConfig.http.port, config.gatewayConfig.http.hostname, () => {
+          // This can either be an object or a String depending on whether you specify a host or not
+          // https://nodejs.org/api/net.html#net_server_address
+          if (config.gatewayConfig.http.hostname) {
+            log.info(`gateway http server listening on ${config.gatewayConfig.http.hostname}:${config.gatewayConfig.http.port}`);
+          } else {
+            const { address, port } = runningApp.address();
+            log.info(`gateway http server listening on ${address}:${port}`);
+          }
 
-          log.info(`gateway http server listening on ${address}:${port}`);
           eventBus.emit('http-ready', { httpServer: runningApp });
 
           apps.httpApp = runningApp;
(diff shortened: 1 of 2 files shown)


lib/gateway/index.js
```diff
@@ -20,12 +20,19 @@ module.exports = function ({ plugins, config } = {}) {
   const { httpServer, httpsServer } = bootstrap({ plugins, config });
 
   if (config.gatewayConfig.http && httpServer) {
+    // config.gatewayConfig.http.hostname = config.gatewayConfig.http.hostname || 'localhost';
```
XVincentX: You can safely remove this line. We do not need to provide a default value since nodejs will handle that for us.

lib/gateway/index.js
```diff
@@ -20,12 +20,19 @@ module.exports = function ({ plugins, config } = {}) {
...
     appPromises.push(
       new Promise(resolve => {
-        const runningApp = httpServer.listen(config.gatewayConfig.http.port, () => {
-          const { address, port } = runningApp.address();
+        const runningApp = httpServer.listen(config.gatewayConfig.http.port, config.gatewayConfig.http.hostname, () => {
+          // This can either be an object or a String depending on whether you specify a host or not
+          // https://nodejs.org/api/net.html#net_server_address
+          if (config.gatewayConfig.http.hostname) {
```
XVincentX: Instead of checking our configuration values, I'd rather check the values returned by the `address()` function to decide what to print or not.

```javascript
            const addressInfo = runningApp.address();
            if (typeof addressInfo === 'string') {
              log.info(`gateway http server listening on ${addressInfo}`);
            } else {
              log.info(`gateway http server listening on ${addressInfo.address}:${addressInfo.port}`);
            }
```

XVincentX:
@bploetz 

I've added some comments on how to structure the code a little bit better. We'll also need couple of of tests to make sure EG is then listening on the right hostname. Let me know your thoughts and whether you want to proceed with this or I should take over. I'm happy to guide you through all the steps; we really need more contributors on the project!

bploetz:
@XVincentX I'm happy to carry this over the finish line, but I could use some guidance as to where to put the test that asserts the gateway is listening on the right hostname. It wasn't readily apparent when I looked at how the tests were structured. If you could nudge me in the right direction that would be great!

XVincentX:
I'm glad to know that. I'll definitely provide you some help in writing these. I'll get back to you hopefully later today.

I'm sorry for the back and forth but I'm really on crazy times. :)

bploetz:
@XVincentX No worries. Have a 🍺 and relax.

XVincentX:
@bploetz I've added a commit to your branch which is simplifying the code a little bit and more importantly deduplicating it. If you could go on your pull request and enable the "edit from maintainers" option then Github would show it here and we can work together on that!

XVincentX:
@bploetz As for testing, I think the simplest way would be to create another file not that different from https://github.com/ExpressGateway/express-gateway/blob/master/test/config-https-sni.test.js where you could make the gateway listen on a particular interface and then making sure that it proxies the things exclusively from that ip address. Does that make sense to you?

XVincentX:
Moved on #846

bploetz:
@XVincentX sorry, I don't see where to enable "edit from maintainers". I think because you were actually the one to open this PR? I only filed an issue.

PR #847, opened 2018-11-22
Remove swagger docs and middleware

XVincentX:
It turns out this is a feature that's not documented, outdated and not even meeting any standard decency. Plus, the gateway should not serve such document directly.

Let's remove this feature.


XVincentX:
I will never understand how removing code is decreasing code coverage.

Issue #833, opened 2018-10-19
Evaluate/replace chokidar with nsfw


XVincentX:
Chokidar works great with OSX — but it's falling back to the watch function provided by NodeJS for other platform.

It turns out there's another library that goes native for all of them. This can be useful especially for Docker container, where the events might work correctly without having to use the chokidar pool hack.

https://github.com/Axosoft/nsfw

Evaluate, verify, and replace in case it makes sense.

XVincentX:
Evaluation done, it's not worth the effort. As long they do not provide a pre built binary, this is not suitable for EG.

PR #848, opened 2018-11-23
Replace chokidar with nsfw

XVincentX:
# WIP

Connect #833 
Closes #833


XVincentX:
Not worth at this moment. This would require too many changes in docker images and docker official images cannot go with multistage builds. Deferred.

PR #852, opened 2018-12-05
Expose internal express app instance

brandonros:
I need this to be able to instrument health check middleware.


XVincentX:
I understand your use case here but I do not believe this is the right way to make this happen. 

This should be exposed potentially in the plugin framework. Right now, the best way to do this is to use the `registerGatewayRoute` from a plugin. You will receive the app instance directly from there where you can attach your middleware.

PR #853, opened 2018-12-05
Add support for invoking third-party Express middlewares at startup

brandonros:
Some middlewares (like https://github.com/roylines/node-epimetheus) require that they are the first middleware added to an Express app instance.

This function allows `express-gateway`'s internal `app` instance to be configured against third-party middleware.

### DIFF
--- a/lib/gateway/index.js
+++ b/lib/gateway/index.js
@@ -7,7 +7,7 @@ const policies = require('../policies');
 const conditions = require('../conditions');
 const passport = require('passport');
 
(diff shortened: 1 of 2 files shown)


XVincentX:
I understand your use case here but I do not believe this is the right way to make this happen. 

This should be exposed potentially in the plugin framework. Right now, the best way to do this is to use the `registerGatewayRoute` from a plugin. You will receive the app instance directly from there where you can attach your middleware.

Issue #856, opened 2018-12-20
insertCredential fails when adding the second credential


devzsolt:
Credential service's `insertCredential()` calls `getCredential()` which fails to consider the scenario when a credential is not a single hash but a set of hashes in the DB.

First it causes a warning: `WRONGTYPE Operation against a key holding the wrong kind of value`
which is followed by the following error:

```
TypeError: Cannot read property 'properties' of undefined at processCredential (/path-to-project/node_modules/express-gateway/lib/services/credentials/credential.service.js:255:51) 
  at s.getCredential.credentialDao.getCredential.then.credential (/path-to-project/node_modules/express-gateway/lib/services/credentials/credential.service.js:138:14) 
```

It's possible to reproduce always by creating a user with a credential and then deactivating the credential and creating a new one for the same user. Only key-auth and jwt type credentials are affected by the bug.

XVincentX:
Thanks for the fix. I hope to be able to jump on this soon, I just didn't have the time.

PR #857, opened 2018-12-20
fix: avoid failure when inserting another credential for a user

devzsolt:
resolves https://github.com/ExpressGateway/express-gateway/issues/856

insertCredential will call dao's getAllCredentials() which can handle key-auth and jwt type credentials that were affected by the bug

some extra error handling were also added

### DIFF
--- a/lib/services/credentials/credential.dao.js
+++ b/lib/services/credentials/credential.dao.js
@@ -124,7 +124,18 @@ dao.insertCredential = function (id, type, credentialObj) {
 };
 
 dao.getCredential = function (id, type) {
-  return db.hgetall(buildIdKey(type, id)).then(credential => {
+  const key = buildIdKey(type, id);
+  return db.type(key).then(type => {
+    if (type === 'hash') {
+      return db.hgetall(key)
+    } else if (type === 'set') {
+      throw Error(`Implementation error: ${key} is a set, therefore getAllCredentials() should be called instead of getCredential()`);
+    } else if (type === 'none') {
+      return null;
+    } else {
+      throw Error(`Corrupt data found at ${key}`);
+    }
+  }).then(credential => {
     if (!credential || Object.keys(credential).length === 0) return null;
     credential.isActive = credential.isActive === 'true'; // Redis has no bool type, manual conversion
     credential.type = type;
--- a/lib/services/credentials/credential.dao.js
+++ b/lib/services/credentials/credential.dao.js
@@ -127,7 +127,7 @@ dao.getCredential = function (id, type) {
   const key = buildIdKey(type, id);
   return db.type(key).then(type => {
     if (type === 'hash') {
-      return db.hgetall(key)
+      return db.hgetall(key);
     } else if (type === 'set') {
       throw Error(`Implementation error: ${key} is a set, therefore getAllCredentials() should be called instead of getCredential()`);
     } else if (type === 'none') {
--- a/lib/services/credentials/credential.service.js
+++ b/lib/services/credentials/credential.service.js
@@ -45,10 +45,24 @@ s.insertCredential = function (id, type, credentialDetails) {
     throw new Error(`Invalid credential type: ${type}`); // TODO: replace with validation error
   }
 
-  return this.getCredential(id, type) // check if credential already exists
+  return credentialDao.getAllCredentials(id) // check if credential already exists
     .then((cred) => {
-      if (cred && cred.isActive) {
-        throw new Error('Credential already exists and is active'); // TODO: replace with validation error
+      if (Array.isArray(cred)) {
+        cred.forEach((c) => {
+          if (c && typeof c.isActive !== 'boolean') {
+            throw new Error('Failed to get existing credentials'); // TODO: replace with validation error
+          }
+          if (c && c.isActive) {
+            throw new Error(`Credential ${c.keyId} already exists and is active`); // TODO: replace with validation error
+          }
+        });
+      } else {
+        if (cred && typeof cred.isActive !== 'boolean') {
+          throw new Error('Failed to get existing credentials'); // TODO: replace with validation error
+        }
+        if (cred && cred.isActive) {
+          throw new Error(`Credential ${cred.keyId} already exists and is active`); // TODO: replace with validation error
+        }
       }
       return dereferencePromise;
     })


devzsolt:
The failing test (https://github.com/ExpressGateway/express-gateway/blob/master/test/services/credential-service/credentials.test.js#L50) is contradicting with the existing logic in the code (https://github.com/ExpressGateway/express-gateway/blob/master/lib/services/credentials/credential.service.js#L51)

What is the expected behavior?
- **A**: allow multiple active credentials (in this case we need to change the code)
- **B**: don't allow multiple active credentials (in this case the now failing test needs to be removed imho)

lib/services/credentials/credential.dao.js
```diff
@@ -124,7 +124,18 @@ dao.insertCredential = function (id, type, credentialObj) {
...
+  const key = buildIdKey(type, id);
+  return db.type(key).then(type => {
+    if (type === 'hash') {
+      return db.hgetall(key);
+    } else if (type === 'set') {
+      throw Error(`Implementation error: ${key} is a set, therefore getAllCredentials() should be called instead of getCredential()`);
+    } else if (type === 'none') {
+      return null;
+    } else {
+      throw Error(`Corrupt data found at ${key}`);
```
XVincentX: I believe this should be a new instance

```suggestion
      throw new Error(`Corrupt data found at ${key}`);
```

lib/services/credentials/credential.dao.js
```diff
@@ -124,7 +124,18 @@ dao.insertCredential = function (id, type, credentialObj) {
 };
 
 dao.getCredential = function (id, type) {
-  return db.hgetall(buildIdKey(type, id)).then(credential => {
+  const key = buildIdKey(type, id);
+  return db.type(key).then(type => {
+    if (type === 'hash') {
+      return db.hgetall(key);
+    } else if (type === 'set') {
+      throw Error(`Implementation error: ${key} is a set, therefore getAllCredentials() should be called instead of getCredential()`);
```
XVincentX:
```suggestion
      throw new Error(`Implementation error: ${key} is a set, therefore getAllCredentials() should be called instead of getCredential()`);
```

lib/services/credentials/credential.dao.js
```diff
@@ -124,7 +124,18 @@ dao.insertCredential = function (id, type, credentialObj) {
 };
 
 dao.getCredential = function (id, type) {
```
XVincentX: Maybe this method could be renamed in `getActiveCredential` — so that instead of throwing an error in case it's a `Set`, it could further filter the returned data and return the active credential.

lib/services/credentials/credential.service.js
```diff
@@ -45,10 +45,24 @@ s.insertCredential = function (id, type, credentialDetails) {
...
     .then((cred) => {
-      if (cred && cred.isActive) {
-        throw new Error('Credential already exists and is active'); // TODO: replace with validation error
+      if (Array.isArray(cred)) {
+        cred.forEach((c) => {
+          if (c && typeof c.isActive !== 'boolean') {
+            throw new Error('Failed to get existing credentials'); // TODO: replace with validation error
+          }
+          if (c && c.isActive) {
+            throw new Error(`Credential ${c.keyId} already exists and is active`); // TODO: replace with validation error
```
XVincentX: I know this is weird — but the credential can have an `id` or a `keyId`

```suggestion
            throw new Error(`Credential ${c.id || c.keyId} already exists and is active`); // TODO: replace with validation error
```

lib/services/credentials/credential.service.js
```diff
@@ -45,10 +45,24 @@ s.insertCredential = function (id, type, credentialDetails) {
     throw new Error(`Invalid credential type: ${type}`); // TODO: replace with validation error
   }
 
-  return this.getCredential(id, type) // check if credential already exists
+  return credentialDao.getAllCredentials(id) // check if credential already exists
     .then((cred) => {
-      if (cred && cred.isActive) {
-        throw new Error('Credential already exists and is active'); // TODO: replace with validation error
+      if (Array.isArray(cred)) {
+        cred.forEach((c) => {
```
XVincentX:
```suggestion
        cred.forEach(c => {
```

lib/services/credentials/credential.service.js
```diff
@@ -45,10 +45,24 @@ s.insertCredential = function (id, type, credentialDetails) {
...
+          if (c && typeof c.isActive !== 'boolean') {
```
XVincentX: Is this a condition that can even happen? I'm usually in for defensive programming but in this case I really do not see this happening. 🤔

lib/services/credentials/credential.service.js
```diff
@@ -45,10 +45,24 @@ s.insertCredential = function (id, type, credentialDetails) {
...
+            throw new Error('Failed to get existing credentials'); // TODO: replace with validation error
+          }
+          if (c && c.isActive) {
+            throw new Error(`Credential ${c.keyId} already exists and is active`); // TODO: replace with validation error
+          }
+        });
+      } else {
+        if (cred && typeof cred.isActive !== 'boolean') {
```
XVincentX: ditto.

lib/services/credentials/credential.service.js
```diff
@@ -45,10 +45,24 @@ s.insertCredential = function (id, type, credentialDetails) {
...
+          throw new Error('Failed to get existing credentials'); // TODO: replace with validation error
+        }
+        if (cred && cred.isActive) {
+          throw new Error(`Credential ${cred.keyId} already exists and is active`); // TODO: replace with validation error
```
XVincentX:
```suggestion
          throw new Error(`Credential ${cred.id || cred.keyId} already exists and is active`); // TODO: replace with validation error
```

lib/services/credentials/credential.service.js
```diff
@@ -45,10 +45,24 @@ s.insertCredential = function (id, type, credentialDetails) {
...
 
-  return this.getCredential(id, type) // check if credential already exists
+  return credentialDao.getAllCredentials(id) // check if credential already exists
     .then((cred) => {
-      if (cred && cred.isActive) {
-        throw new Error('Credential already exists and is active'); // TODO: replace with validation error
+      if (Array.isArray(cred)) {
+        cred.forEach((c) => {
+          if (c && typeof c.isActive !== 'boolean') {
+            throw new Error('Failed to get existing credentials'); // TODO: replace with validation error
```
XVincentX: We might want to remove all these comments with a TODO which is not likely to be executed anytime soon.

XVincentX:
Closed for lack of interest. In case you want to resume this, feel free to reopen it!

PR #861, opened 2019-01-02
Honor content negotiation in 404 responses from gateway

bploetz:
Regardless of what the client specifies in the `Accept` header, when a 404 is returned from express gateway, the response content type and body is always HTML, which is a little odd since most folks will be using this with JSON based APIs.

Example:

```
curl -vv -H "Host: api.mydomain.com" -H "Content-Type: application/json" -H "Accept: application/json" -d '{"foo":"bar"}' http://localhost:8080/v1/something
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> POST /v1/something HTTP/1.1
> Host: api.mydomain.com
> User-Agent: curl/7.54.0
> Content-Type: application/json
> Accept: application/json
> Content-Length: 51
> 
* upload completely sent off: 51 out of 51 bytes
< HTTP/1.1 404 Not Found
< Content-Security-Policy: default-src 'self'
< X-Content-Type-Options: nosniff
< Content-Type: text/html; charset=utf-8
< Content-Length: 147
< Date: Wed, 02 Jan 2019 14:46:31 GMT
< Connection: keep-alive
< 
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot POST /v1/something</pre>
</body>
</html>
* Connection #0 to host localhost left intact
```

With this change, 404s will now honor JSON requests. If we wanted to add support for XML or something else, that's trivial to add as well.

Note that the version of express-rate-limit that express-gateway currently depends on (2.x) is using `res.format()` to toggle between various response content types, and you can't customize the error response, but the latest version uses `res.send()`, and you could then in theory send custom 404 response bodies, but you'd likely need to change the schema for the rate limit policy to support this. See: https://github.com/nfriedly/express-rate-limit/issues/63.

There may be other areas where eg is blindly returning HTML. I'm happy to address those as well if you know of examples off the top of your head.

Thanks.

### DIFF
--- a/lib/policies/terminate/terminate.js
+++ b/lib/policies/terminate/terminate.js
@@ -1,3 +1,10 @@
 module.exports = (params) =>
-  (req, res, next) =>
-    res.status(params.statusCode).send(params.message);
+  (req, res, next) => {
+    let payload;
+    if (req.accepts('json')) {
+      payload = { message: params.message };
+    } else {
+      payload = params.message;
+    }
+    res.status(params.statusCode).send(payload);
+  };
--- a/lib/policies/terminate/terminate.js
+++ b/lib/policies/terminate/terminate.js
@@ -1,10 +1,8 @@
 module.exports = (params) =>
   (req, res, next) => {
-    let payload;
     if (req.accepts('json')) {
-      payload = { message: params.message };
+      res.status(params.statusCode).send({ message: params.message });
     } else {
-      payload = params.message;
+      res.status(params.statusCode).send(params.message);
     }
-    res.status(params.statusCode).send(payload);
   };
--- a/lib/gateway/index.js
+++ b/lib/gateway/index.js
@@ -89,7 +89,13 @@ function bootstrap ({ plugins, config } = {}) {
   // 404
   app.use((req, res, next) => {
     if (req.accepts('json')) {
-      res.status(404).json({ error: 'resource not found' });
+      res.setHeader('content-type', 'application/problem+json');
+      res.status(404)
+        .json({
+          type: 'https://express-gateway.io/errors#NOT_FOUND',
+          title: 'Resource not found',
+          status: 404
+        });
     } else {
       next();
     }
--- a/lib/gateway/index.js
+++ b/lib/gateway/index.js
@@ -86,6 +86,15 @@ function bootstrap ({ plugins, config } = {}) {
     rootRouter(req, res, next);
   });
 
+  // 404
+  app.use((req, res, next) => {
+    if (req.accepts('json')) {
+      res.status(404).json({ error: 'resource not found' });
+    } else {
+      next();
+    }
+  });
+
   eventBus.on('hot-reload', (hotReloadContext) => {
     const oldRootRouter = rootRouter;
     try {
--- a/test/common/routing.helper.js
+++ b/test/common/routing.helper.js
@@ -17,6 +17,7 @@ module.exports = function () {
     }
 
     testScenario.set('Content-Type', 'application/json');
+    testScenario.set('Accept', 'application/json');
 
     if (testCase.setup.host) {
       testScenario.set('Host', testCase.setup.host);
@@ -69,11 +70,14 @@ module.exports = function () {
       return this.validateError(testCase);
     },
     validateError: (testCase) => {
+      testCase.test = testCase.test || {};
+      // allow tests to override the default content type expected
+      testCase.test.contentType = testCase.test.contentType || 'application/json; charset=utf-8';
       return (done) => {
         const testScenario = prepareScenario(testCase);
         testScenario
           .expect(testCase.test.errorCode)
-          .expect('Content-Type', /text\/html/)
+          .expect('Content-Type', testCase.test.contentType)
           .end((err, res) => { done(err); });
       };
     },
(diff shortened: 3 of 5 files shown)


bploetz:
I tried to address the code coverage error, but it looks like that does not re-run on subsequent pushes?

bploetz:
FYI, output when pointing my api gateway to this change:

```
curl -vv -H "Host: api.mydomain.com" -H "Content-Type: application/json" -H "Accept: application/json" -d '{"foo":"bar"}' http://localhost:8080/v1/something
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> POST /v1/something HTTP/1.1
> Host: api.merryfield.com
> User-Agent: curl/7.54.0
> Content-Type: application/json
> Accept: application/json
> Content-Length: 51
> 
* upload completely sent off: 51 out of 51 bytes
< HTTP/1.1 404 Not Found
< Content-Type: application/json; charset=utf-8
< Content-Length: 30
< ETag: W/"1e-KJNPqC/+VRFVMPmZOIGw9PI8thY"
< Date: Wed, 02 Jan 2019 22:11:26 GMT
< Connection: keep-alive
< 
* Connection #0 to host localhost left intact
{"error":"resource not found"}
```

XVincentX:
Hey @bploetz,

I'd like to get some more context. I've briefly looked at the code and I think some changes might be needed.

Apart from that, I'd love to understand a little bit more the use case. Isn't 404 (as status code) good enough to understand what is going on? Would you really need a JSON-based response?

Moreover, what would be the perfect error shape? Would code and message properties be good enough? Would we need to adopt a vnd+json based error message?

At the current state, I'm not so sure this change would fit; but I'm willing to listen to your particular use case.

bploetz:
@XVincentX It's simple HTTP content negotiation semantics. If the client says (via the `Accept` header) that it wants a JSON response, and you send it back an HTML response (regardless of what response status code we're talking about), it's just wrong. If the server is UNABLE to generate JSON responses, then it should send back a 406 status code. But of course we know eg can speak JSON.

As for the actual JSON body we send back in response to 404's, I'm open to whatever. What I implemented was just a strawman really.

Hope that helps!

XVincentX:
Got it — it makes sense and we should do something on this.

I'll have a look at what ExpressJS does by default and if there's any internal mechanism that can handle this without any code intervention. If not, we'll proceed with your solution and polish it. Does that sound like a plan?

In meantime (just out of curiosity) is this particular issue preventing you in any use case?

bploetz:
Yeah I think because you didn't have a specific 404 handler, you were just getting the default express error handler behavior (https://expressjs.com/en/guide/error-handling.html) which is where that generic HTML response is coming from I believe.

This is NOT preventing any specific use case on my end, it's not breaking anything, it's just something I noticed that was wrong and wanted to help fix. It's extremely low priority, obviously.

XVincentX:
@bploetz As a starting point, I think it would be good enough to restrict this change to the `terminate` policy only. In that case, I think your change makes completely sense. While I look into the alternatives for the Express in general — I'd be ok merging a PR modifying that policy only. What do you think?

lib/policies/terminate/terminate.js
```diff
@@ -1,3 +1,10 @@
...
-  (req, res, next) =>
-    res.status(params.statusCode).send(params.message);
+  (req, res, next) => {
+    let payload;
+    if (req.accepts('json')) {
+      payload = { message: params.message };
+    } else {
+      payload = params.message;
+    }
+    res.status(params.statusCode).send(payload);
```
XVincentX: I believe, in case is a JSON response, to use `res.status(500).json({ message: params.message });` so that the content type header is set correctly.

lib/policies/terminate/terminate.js
```diff
@@ -1,3 +1,10 @@
 module.exports = (params) =>
-  (req, res, next) =>
-    res.status(params.statusCode).send(params.message);
+  (req, res, next) => {
+    let payload;
```
XVincentX: I wouldn't bother creating a `let` variable on the top — I'd construct the payload on the fly.

I always sweat when I see a `let` variable :)

lib/policies/terminate/terminate.js
bploetz: `send` does the right thing based on the argument.

https://expressjs.com/en/4x/api.html#res.send

bploetz:
@XVincentX It's obviously your call, but I'd rather take the time to fix this issue holistically than just putting a band aid on the terminate policy alone. It would be weird if just that policy did the right thing, but the rest of the system didn't.

But again, I defer to you......

lib/policies/terminate/terminate.js
bploetz: LOL, that's fine. I was following the lead of similar situations I saw throughout the code (for example https://github.com/ExpressGateway/express-gateway/blob/master/lib/policies/basic-auth/basic-auth.js#L11) and avoid repeating the `res.status().send()` line, but I'm happy to use whatever style you prefer.

XVincentX: I understand you're confusion. This code base has been written by multiple persons with the passing of the time before it ended up to me — and I have a completely different mindset compared to the guys that were working on this before.

bploetz: Fixed the style in ac8f74d

XVincentX:
I'm closing this in favour of generally using `application/problem+json` for all our errors. Thanks for the input!

PR #862, opened 2019-01-02
clean up logic for removing creds on delete user

mfrye:
The problem with the current logic is that the Redis Set containing the credentials isn't deleted on user delete. So when deleting the user and user owned credentials, the set still exists.

### DIFF
--- a/test/services/users.test.js
+++ b/test/services/users.test.js
@@ -6,13 +6,11 @@ const userService = services.user;
 const credentialService = services.credential;
 const db = require('../../lib/db');
 
-describe('User service tests', function () {
-  describe('Insert tests', function () {
-    before(function () {
-      return db.flushdb();
-    });
+describe('User service tests', () => {
+  describe('Insert tests', () => {
+    before(() => db.flushdb());
 
-    it('should insert a user', function () {
+    it('should insert a user', () => {
       const user = {
         username: 'irfanbaqui',
         firstname: 'irfan',
@@ -35,26 +33,18 @@ describe('User service tests', function () {
         });
     });
 
-    it('should throw an error when inserting a user with missing properties', function (done) {
+    it('should throw an error when inserting a user with missing properties', () => {
       const user = {
         username: 'irfanbaqui-1',
         lastname: 'baqui',
         email: 'irfan@eg.com'
       };
 
-      userService
-        .insert(user)
-        .then(function (newUser) {
-          should.not.exist(newUser);
-        })
-        .catch(function (err) {
-          should.exist(err);
-          err.message.should.eql('data should have required property \'firstname\'');
-          done();
-        });
+      return should(userService.insert(user))
+        .be.rejectedWith('data should have required property \'firstname\'');
     });
 
-    it('should throw an error when inserting a user with existing username', function (done) {
+    it('should throw an error when inserting a user with existing username', () => {
       const user = {
         username: 'irfanbaqui',
         firstname: 'irfan',
@@ -62,20 +52,12 @@ describe('User service tests', function () {
         email: 'irfan@eg.com'
       };
 
-      userService
-        .insert(user)
-        .then(function (newUser) {
-          should.not.exist(newUser);
-        })
-        .catch(function (err) {
-          should.exist(err);
-          err.message.should.eql('username already exists');
-          done();
-        });
+      return should(userService.insert(user))
+        .be.rejectedWith('username already exists');
     });
   });
 
-  describe('Get and Find User tests', function () {
+  describe('Get and Find User tests', () => {
     let user;
     before(() => db.flushdb().then(() => {
       user = createRandomUserObject();
@@ -85,8 +67,8 @@ describe('User service tests', function () {
       user.id = newUser.id;
     }));
 
-    it('should get user by userId', function (done) {
-      userService.get(user.id)
+    it('should get user by userId', () => {
+      return userService.get(user.id)
         .then(function (_user) {
           const expectedUserProps = ['firstname', 'lastname', 'email', 'isActive', 'username', 'id', 'createdAt', 'updatedAt'];
           should.exist(_user);
@@ -100,12 +82,10 @@ describe('User service tests', function () {
             email: user.email,
             username: user.username
           });
-          done();
-        })
-        .catch(done);
+        });
     });
 
-    it('should get user all users', function () {
+    it('should get user all users', () => {
       return userService.findAll()
         .then(function (data) {
           should.exist(data.users);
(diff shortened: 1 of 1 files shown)


mfrye:
@XVincentX could you take a look at this one? This bug is the last thing that's blocking us from deploying this.

XVincentX:
I need to find some bandwidth but yeah I'll look into //cc @altsang

mfrye:
I'm open to other ideas on the optional parameter I included. I figured there might be a case where you'd want to delete all credentials, but still keep the user "container".

mfrye:
Separate bug... Looks like the keys also still exist under each scope as well.

lib/services/credentials/credential.dao.js
```diff
@@ -145,20 +145,37 @@ dao.removeCredential = function (id, type) {
...
+ *
+ * @id {String}
+ * @deleteAll {Boolean} Optional - Whether to delete the key / set that contains all creds for users.
+ */
+dao.removeAllCredentials = function (id, deleteAll) {
   const dbTransaction = db.multi();
   const credentialTypes = Object.keys(config.models.credentials.properties);
-  const awaitAllPromises = [];
-  credentialTypes.forEach((type) => {
+  const awaitAllPromises = credentialTypes.map((type) => {
```
XVincentX: If it's a single parameter, you can avoid the parentheses

```suggestion
  const awaitAllPromises = credentialTypes.map(type => {
```

XVincentX: 👍 for the `map`

lib/services/credentials/credential.dao.js
```diff
@@ -145,20 +145,37 @@ dao.removeCredential = function (id, type) {
...
     if (type === 'key-auth') {
       // id in this call is actually consumerId, so we need to get all referenced keyIds
-      awaitAllPromises.push(db.smembers(buildIdKey(type, id)).then(ids => {
-        ids.map(keyId => {
-          dbTransaction.del(buildIdKey(type, keyId));
+      // This returns a list of all keys the user owns
+      return db.smembers(authKey)
+        .then(ids => {
+          // Delete each
+          const promises = ids.map(keyId => {
```
XVincentX: You can write this as a single line

`const promises = ids.map(keyId => dbTransaction.del(buildIdKey(type, keyId))`

lib/services/credentials/credential.dao.js
```diff
@@ -145,20 +145,37 @@ dao.removeCredential = function (id, type) {
...
+            return dbTransaction.del(buildIdKey(type, keyId));
+          });
+
+          // Now delete the main key for the user that lists all creds
+          if (deleteAll) {
```
XVincentX: I do not think we should provide such flag. This should always be happening.

XVincentX:
Also — can you revert the `package-lock.json` changes?

lib/services/credentials/credential.dao.js
mfrye: Cool.

lib/services/credentials/credential.dao.js
mfrye: Sounds good.

mfrye:
Yes on the package-lock and your other recommended changes. 

Where should the tests be added for this? The code coverage % is still going to be below your threshold even if I add tests for this.

XVincentX:
> Where should the tests be added for this? The code coverage % is still going to be below your threshold even if I add tests for this.

Yeah good point. The problem is that sometimes Codecov reveals areas that have never been tested, thous reporting something wrong with the coverage. Add the tests you feel are useful and I'll check that out for you later and integrate in case it's needed.

P.S: Nowadays I'm using Typescript too. I have opinions on it :)

lib/services/credentials/credential.dao.js
```diff
@@ -145,20 +145,47 @@ dao.removeCredential = function (id, type) {
...
+ * Remove all credentials
+ *
+ * @id {String}
+ */
 dao.removeAllCredentials = function (id) {
   const dbTransaction = db.multi();
   const credentialTypes = Object.keys(config.models.credentials.properties);
-  const awaitAllPromises = [];
-  credentialTypes.forEach((type) => {
+  const awaitAllPromises = credentialTypes.map(async type => {
```
XVincentX: I know that might be annoying to you but Express Gateway is using Promises and not any async/await constructs. This because it used to support Node 6 at the beginning. Although it would make sense to change, I do not think this is the right time to migrate.

package-lock.json
```diff
@@ -9200,4 +9200,4 @@
       }
     }
   }
-}
+}
```
XVincentX: Can you restore the new line?

lib/services/credentials/credential.dao.js
```diff
@@ -145,20 +145,47 @@ dao.removeCredential = function (id, type) {
...
-        });
-      }));
+      // This returns a list of all keys the user owns
+      const ids = await db.smembers(authKey);
+
+      // Get all the scope so we can remove key from them
+      const scopes = await dao.getAllScopes();
+
+      // Delete each key and remove key from scopes if they exist
+      const promises = [];
```
XVincentX: Nitpick: new line.

lib/services/credentials/credential.dao.js
```diff
@@ -145,20 +145,47 @@ dao.removeCredential = function (id, type) {
...
+      ids.forEach(keyId => {
```
XVincentX: Maybe this can become another `.map`?

lib/services/credentials/credential.dao.js
mfrye: Any chance you could ask Al about this? For this thing in particular, async/await really helps keep it clean.

I was debating on whether to include it, so I specifically looked at your package.json and docker file. Both say that the engine is `"node": ">= 8.0.0"`. Then technically speaking, current version of LTS node is 10.

I don't think you have to migrate all at once. I think adding it gradually where it makes sense is fine.

package-lock.json
mfrye: Fixed

lib/services/credentials/credential.dao.js
mfrye: Fixed

lib/services/credentials/credential.dao.js
mfrye: Map won't work here because we loop through the scopes for each key as well. So it has the potential to return more than one item.

In case you didn't notice, I also fixed the scopes thing since that wasn't being cleaned up on user delete either.

mfrye:
@XVincentX I attempted to add a test to `test/services/credential-service/credentials.test.js`. I'm not sure what happened though as it ended up blowing up other unrelated tests...

I can try to look at this again, but I really don't have a ton of time to spend on this if it's going to be a bigger refactor.

XVincentX:
Do not worry, leave the tests to me and I'll figure this out.

mfrye:
Cool, thanks.

XVincentX:
@mfrye Thanks a lot for the contribution. Writing the tests ended up being a little bit more difficult than anticipated because of the horrible state of some parts of the codebase. 

However, we should be good now. Thanks!

mfrye:
@XVincentX awesome. Thanks for fast tracking this.

PR #866, opened 2019-01-19
OAuth2 didnt authenticate if no redirectUri but client provided one.

wintersandroid:
I was having trouble authenticating Oauth2, both with google home actions, and the postman app.

I found that by not having a redirecturi which is optional for the user/app in express gateway, and the authenticating app providing one that express-gateway would fail.
The reason for not having the redirect uri set, is that the url given is not known.

This pull request resolves that problem by not testing equality if the user does not have one specified.

### DIFF
--- a/lib/policies/oauth2/oauth2-server.js
+++ b/lib/policies/oauth2/oauth2-server.js
@@ -267,7 +267,7 @@ module.exports.authorization = [
   server.authorization((req, done) => {
     return authService.validateConsumer(req.clientID)
       .then(consumer => {
-        if (!consumer || consumer.redirectUri !== req.redirectURI) return done(null, false);
+        if (!consumer || (consumer.redirectUri && consumer.redirectUri !== req.redirectURI)) return done(null, false);
 
         if (!req.scope) {
           return done(null, consumer, req.redirectURI);


XVincentX:
Thanks for the pull request. At first glance, it seems to make perfect sense. I just want to circulate it back and make sure that this is the right way to solve the problem.

Hopefully I'll be able to jump on EG soon :)

XVincentX:
@wintersandroid

wintersandroid:
Ok, I was testing (using postman & google home actions) to see if express-gateway was suitable, I didnt fill in the redirectUri's as the case with the google one, I didn't know what it will be. And testing with postman, well the redirect is not actually done.

This change can be dropped, I actually found that express-gateway will not complete authentication on a RaspberryPi either in our out of docker. I cant find the reason why so I can't use it for this project.

XVincentX:
I'll be happy to help you in case you need assistance, we have a Gitter channel where you can ask question and get some community support; you can find the link on our project website.
Express Gateway works in Docker with no problems and it's particularly well suited for IoT applications, so I invite you to reconsider.

That said, until the issue is verified, I'm going to close this.

PR #874, opened 2019-03-06
Update rate limit policy to set redis expiry inline with windowMs

ciaranashton:
Redis rate limiter expires by default at 60 seconds. 

Dividing the `windowMs` into seconds and then passing it through to replace the default `expiry: 60` will ensure you get correct rate limiting when storing counters in Redis.


XVincentX:
Good catch! Thanks a lot for this!

PR #876, opened 2019-03-11
fix: replace hostname with host in admin config

### LINKED ISSUE #432: host versus hostname: need consistency

ravikp7:
- improves consistency with apiEndpoints config
- generators now use host in admin config
- hostname is still supported, but warning is shown

fixes #432

### DIFF
--- a/bin/eg-generator.js
+++ b/bin/eg-generator.js
@@ -71,7 +71,7 @@ module.exports = class EgGenerator extends Generator {
       baseURL = systemConfig.cli.url;
     } else if (gatewayConfig && gatewayConfig.admin) {
       const adminConfig = gatewayConfig.admin;
-      const host = adminConfig.host || adminConfig.hostname;
+      const host = adminConfig.host || adminConfig.hostname || 'localhost';
       const port = adminConfig.port || 9876;
 
       baseURL = `http://${host}:${port}`;
--- a/bin/eg-generator.js
+++ b/bin/eg-generator.js
@@ -71,10 +71,10 @@ module.exports = class EgGenerator extends Generator {
       baseURL = systemConfig.cli.url;
     } else if (gatewayConfig && gatewayConfig.admin) {
       const adminConfig = gatewayConfig.admin;
-      const hostname = adminConfig.hostname || 'localhost';
+      const host = adminConfig.host || adminConfig.hostname;
       const port = adminConfig.port || 9876;
 
-      baseURL = `http://${hostname}:${port}`;
+      baseURL = `http://${host}:${port}`;
     }
 
     /*
--- a/lib/config/schemas/gateway.config.json
+++ b/lib/config/schemas/gateway.config.json
@@ -102,136 +102,130 @@
       }
     },
     "admin": {
+      "allOf": [
+        {
+          "properties": {
+            "port": {
+              "type": "number",
+              "default": 9876
+            },
+            "host": {
+              "type": "string"
+            },
+            "hostname": {
+              "type": "string"
+            }
+          },
+          "required": [
+            "port"
+          ]
+        }
+      ]
+    }
+  },
+  "apiEndpoints": {
+    "type": [
+      "object",
+      "null"
+    ],
+    "additionalProperties": {
+      "anyOf": [
+        {
+          "$ref": "#/definitions/baseApiEndpoint"
+        },
+        {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/baseApiEndpoint"
+          }
+        }
+      ]
+    }
+  },
+  "serviceEndpoints": {
+    "type": [
+      "object",
+      "null"
+    ],
+    "additionalProperties": {
       "type": "object",
       "properties": {
-        "port": {
-          "type": "number",
-          "default": 9876
+        "url": {
+          "type": "string",
+          "format": "uri"
         },
-        "host": {
-          "type": "string"
-        },
-        "hostname": {
-          "type": "string"
+        "urls": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "format": "uri"
+          }
         }
       },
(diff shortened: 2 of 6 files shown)


bin/eg-generator.js
```diff
@@ -71,10 +71,10 @@ module.exports = class EgGenerator extends Generator {
       baseURL = systemConfig.cli.url;
     } else if (gatewayConfig && gatewayConfig.admin) {
       const adminConfig = gatewayConfig.admin;
-      const hostname = adminConfig.hostname || 'localhost';
+      const host = adminConfig.host || 'localhost';
```
XVincentX: You need also to try to grab the property from the old property that we're not going to deprecate.
```suggestion
      const host = adminConfig.host || adminConfig.hostname || 'localhost';
```

If would be great if you could also delete the default value `localhost` and put it in the default value of the JSON Schema file.

lib/config/schemas/gateway.config.json
```diff
@@ -107,6 +107,9 @@
         "port": {
           "type": "number"
         },
+        "host": {
+          "type": "string"
+        },
```
XVincentX: You should update the JSON Schema to make sure only one of the two properties are present (host or hostname)

lib/rest/index.js
```diff
@@ -9,7 +9,11 @@ module.exports = function ({ plugins, config } = {}) {
     logger.verbose('Admin server is not configured, launch canceled');
     return;
   }
-  cfg.admin.hostname = cfg.admin.hostname || 'localhost';
+  if (cfg.admin.hostname) {
+    // eslint-disable-next-line no-console
+    console.log('Warning! use of hostname is depracated in admin section, use host instead');
```
XVincentX: Do not use `console.log`, use the `logger object`

```suggestion
    logger.warn('Warning! use of hostname is depracated in admin section, use host instead');
```

lib/config/schemas/gateway.config.json
ravikp7: - If we remove `hostname` from here, will it still be able to support old configurations?
- Btw, what's the purpose of this JSON Schema? I'm not sure of this

ravikp7:
I'm not sure of where to put tests or what to test for the changes made in this PR.
Can you please guide me on this?

lib/config/schemas/gateway.config.json
XVincentX: We're not going to remove it, we're going to put a constraint so that the user will only use `host` or `hostname`, using the JSON Schema.

The JSON Schema is basically a JSON file telling how an object should look like, so that you can validate it and make actions.

For instance, if I declare a JSON Schema like this:

```json
{
  "$id": "http://express-gateway.io/models/users.json",
  "type": "object",
  "properties": {
    "firstname": {
      "type": "string"
    },
    "lastname": {
      "type": "string"
    },
    "username": {
      "type": "string"
    },
    "email": {
      "type": "string",
      "format": "email"
    },
    "redirectUri": {
      "type": "string",
      "format": "uri"
    }
  },
  "required": [
    "username",
    "firstname",
    "lastname"
  ]
}
```

Then the object passing the validation will have al least     "username","firstname", "lastname"  properties defined and they're a string.

We use that to validate the configuration files `gateway.config.yml` and `system.config.yml`

ravikp7: Got it, thanks!

lib/rest/index.js
```diff
@@ -9,7 +9,11 @@ module.exports = function ({ plugins, config } = {}) {
...
+  }
+  cfg.admin.host = cfg.admin.hostname || cfg.admin.host || 'localhost';
```
XVincentX: Let's remove the default value and put it in the JSON Schema

```suggestion
  cfg.admin.host = cfg.admin.hostname || cfg.admin.host;
```

XVincentX:
> I'm not sure of where to put tests or what to test for the changes made in this PR.

Good point. Let's first make sure the code is ready and then CodeCov will guide us in the part that needs to be covered.

It's not exactly that mandatory to test this stuff (I can live with it) but it'd be great to onboard you appropriately :)

Keep up the good work!

lib/rest/index.js
ravikp7: If we remove `localhost` default value from here, then `28 tests` are breaking.
- One of them is https://github.com/ExpressGateway/express-gateway/blob/8d4e9646e795538ee2abf272b190cabff0ae7568/test/rest-api/api-endpoint.test.js#L24
- I think, in these tests, the `yaml config` is made up directly from object defined there and `host` is missing there, so it relies on this default value in the `lib/rest/index.js` file and not on the JSON schema default value.
___
If that's the case, then we've two options:
- Add `host` in the config objects for all the failing tests
-  or Leave this default value as it is here

Please let me know if I missed something

lib/config/schemas/gateway.config.json
XVincentX: Once addressed, do not forget to mark the conversation as resolved. This helps the reviews ;)

image

lib/rest/index.js
XVincentX: @ravikp7 The reason is a wrong JSON Schema. It can be tricky sometimes to modify it correctly so I went ahead and modified it for you. See the changes at ca50b189fb7fa1f961bbf0021194d52a6cdd9e47

Basically what is needed is to declare both property in the schema, and then use the `oneOf` feature to specify that only one of the two is required.

I've also modified the order — as we want to give the new property `host` precedence over `hostname`

XVincentX:
I also went ahead and checked what's up with the tests and the untested part do not concern this PR. Good to go! 🚀

PR #877, opened 2019-03-12
Json Schema Policy Validation

XVincentX:


XVincentX:
Keep as reference, try to implement as a condition.

PR #885, opened 2019-03-20
eg cli: header apikey secret gets truncated

ontehfritz:
There is a bug in the eg cli, this example:
`eg users list -H "Authorization:apikey 2YvpwOURTnNB0mDFhOpnxj:08o0ZHLD1pGAIiG61tXc9S`

does not work because the headers in the processHeaders function splits the string by ":" splitting the header into three values, and this piece of code only accepts the first two: `const [headerName, headerValue] = header.split(':');`

which truncates the key secret
the header value sent is `{ Authorization: 'apikey 2YvpwOURTnNB0mDFhOpnxj' }`

the above change only splits the first value` header.split(/:(.+)/)`, getting the correct header with keyid:secret
`{ Authorization: 'apikey 2YvpwOURTnNB0mDFhOpnxj:08o0ZHLD1pGAIiG61tXc9S' }`

without this the eg cli does not work

### DIFF
--- a/bin/eg-generator.js
+++ b/bin/eg-generator.js
@@ -106,7 +106,7 @@ module.exports = class EgGenerator extends Generator {
     const ArrayHeaders = Array.isArray(headers) ? headers : [headers];
 
     return ArrayHeaders.reduce((prev, header) => {
-      const [headerName, headerValue] = header.split(':');
+      const [headerName, headerValue] = header.split(/:(.+)/);
 
       if (headerValue) {
         prev[headerName] = headerValue;


ontehfritz:
Would be awesome to get this merged ASAP, using this on a big project :)

XVincentX:
Thanks, this makes a lot of sense @ontehfritz 

I'd be really curious where are you using Express Gateway — having some satisfaction for the project I invest my time into is always rewarding. Can we chat?

ontehfritz:
Sure we can chat, my email is on my profile. Shoot me an email and we will chat :)

ontehfritz:
Also a new npm package with the bug fix would be great! Thanks!

XVincentX:
Yeah I know, we'll try to get this released asap.

PR #886, opened 2019-03-21
Potential feature - adds a health end point

ontehfritz:
This adds a health end point outside the admin end points, that will always open for health pings (if you secure the admin endpoints), this also survives hot reloads. 

The only issue I find with this,  is /health would have to be a reserved endpoint, so it would not be open for reuse

### DIFF
--- a/lib/gateway/index.js
+++ b/lib/gateway/index.js
@@ -76,8 +76,12 @@ function bootstrap ({ plugins, config } = {}) {
       conditionEngine.register(cond);
     });
   }
+
+  const router = express.Router();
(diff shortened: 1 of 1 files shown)


XVincentX:
I'm pasting here my comment for future reference:

I saw your pull request, but I do not think we should go in this way. The exposed interface you're modifying is the "user space" which I do not want to touch. Such solution must be coming from the Admin API. I'd suggest to write a custom plugin and use the registerAdminRoute to make that happen and then proxy it using the gateway if you want.

PR #890, opened 2019-03-25
feat: lazy load and hot reload for policies

### LINKED ISSUE #870: Lazy load policies
Hi! I've been using Express Gateway on a project I am currently working on. It has been great to work with, thank you very much for the project!

I recently started implementing a custom policy for SAML authentication (using passport) and ran into a bit of an issue. It looks like when EG starts up it loads all of the policies that are included with it regardless if you are using them or not. It looks like the oauth2 policy adds a serializer and deserializer to passport when the module loads. Sadly, the deserializer was causing issues for me because Passport is invoking the oauth deserializer before mine and fails. 

I was wondering if there is a proper way to prevent modules from being loaded. If not, I was wondering if it would be worth it to add the ability to NOT load policies you're not using. Perhaps the loaded policies can be driven off of the `policies` array in the gateway's config file. 

I found a (hacky) workaround for now. Just wanted to check to make sure I am not missing anything.

Thanks!!!

ravikp7:
- only load policies defined in the gateway config

fixes #870

### DIFF
--- a/lib/gateway/index.js
+++ b/lib/gateway/index.js
@@ -53,9 +53,7 @@ const bootstrapPolicies = ({ app, plugins, config } = {}) => {
   }
 
   // Load policies present in config
-  if (config.gatewayConfig.policies) {
-    policies.load(config.gatewayConfig.policies);
-  }
+  policies.load(config.gatewayConfig.policies);
 
   // Load all routes from policies
   // TODO: after all complext policies will go to plugin this code can be removed
@@ -88,9 +86,7 @@ function bootstrap({ plugins, config } = {}) {
   app.use(passport.initialize());
   bootstrapPolicies({ app, plugins, config });
   rootRouter = pipelines.bootstrap({ app: express.Router(), config });
-  app.use((req, res, next) => {
-    rootRouter(req, res, next);
-  });
+  app.use(rootRouter);
 
   eventBus.on('hot-reload', (hotReloadContext) => {
     const oldConfig = config;
@@ -107,8 +103,7 @@ function bootstrap({ plugins, config } = {}) {
       rootRouter = pipelines.bootstrap({ app: express.Router(), config: newConfig });
       log.info('hot-reload config completed');
     } catch (err) {
-      log.error('Could not hot-reload gateway.config.yml. Configuration is invalid. ', err);
-      log.debug('Loading old configuration');
+      log.error(`Could not hot-reload gateway.config.yml. Configuration is invalid. ${err}`);
       bootstrapPolicies({ app, plugins: oldPlugins, config: oldConfig });
       rootRouter = oldRootRouter;
     }
--- a/lib/gateway/index.js
+++ b/lib/gateway/index.js
@@ -11,9 +11,9 @@ module.exports = function ({ plugins, config } = {}) {
   const appPromises = [];
   const apps = {};
   if (plugins && plugins.policies && plugins.policies.length) {
-    plugins.policies.forEach(p => {
-      log.debug('registering policy', p.name);
-      policies.register(p);
+    plugins.policies.forEach(policy => {
+      log.debug(`registering policy ${policy.name}`);
+      policies.register(policy);
     });
   }
   config = config || require('../config');
@@ -54,6 +54,12 @@ function bootstrap({ plugins, config } = {}) {
   app.set('x-powered-by', false);
 
   let rootRouter;
+
+  // Load policies present in config
+  if (config.gatewayConfig.policies) {
+    policies.load(config.gatewayConfig.policies);
+  }
+
   // Load all routes from policies
   // TODO: after all complext policies will go to plugin this code can be removed
   // NOTE: plugins have mechanism to provide custom routes
--- a/lib/gateway/index.js
+++ b/lib/gateway/index.js
@@ -6,16 +6,11 @@ const eventBus = require('../eventBus');
 const policies = require('../policies');
 const conditions = require('../conditions');
 const passport = require('passport');
+const pluginsLoader = require('../plugins');
 
 module.exports = function ({ plugins, config } = {}) {
   const appPromises = [];
   const apps = {};
-  if (plugins && plugins.policies && plugins.policies.length) {
-    plugins.policies.forEach(policy => {
-      log.debug(`registering policy ${policy.name}`);
-      policies.register(policy);
-    });
-  }
   config = config || require('../config');
   const { httpServer, httpsServer } = bootstrap({ plugins, config });
 
@@ -49,11 +44,13 @@ module.exports = function ({ plugins, config } = {}) {
     });
 };
 
-function bootstrap({ plugins, config } = {}) {
-  const app = express();
-  app.set('x-powered-by', false);
-
-  let rootRouter;
+const bootstrapPolicies = ({ app, plugins, config } = {}) => {
+  if (plugins && plugins.policies && plugins.policies.length) {
+    plugins.policies.forEach(policy => {
+      log.debug(`registering policy ${policy.name}`);
+      policies.register(policy);
+    });
+  }
 
   // Load policies present in config
   if (config.gatewayConfig.policies) {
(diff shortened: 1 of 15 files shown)


lib/gateway/index.js
```diff
@@ -11,9 +11,9 @@ module.exports = function ({ plugins, config } = {}) {
   const appPromises = [];
   const apps = {};
   if (plugins && plugins.policies && plugins.policies.length) {
-    plugins.policies.forEach(p => {
-      log.debug('registering policy', p.name);
-      policies.register(p);
+    plugins.policies.forEach(policy => {
+      log.debug(`registering policy ${policy.name}`);
+      policies.register(policy);
```
XVincentX: Good idea on the renaming.

test/e2e/oauth2-authorization-code.js
```diff
@@ -35,7 +35,7 @@ describe('oauth2 authorization code grant type', () => {
           path: '/not-found'
         }
       },
-      policies: ['oauth2', 'proxy'],
+      policies: ['cors', 'jwt', 'basic-auth', 'oauth2', 'proxy'],
```
XVincentX: These policies do not seem to be used in this test. Why did you need to load them?

test/e2e/policy-seq-oauth2-expression-log-ratelimit-proxy.js
```diff
@@ -39,7 +39,7 @@ describe('E2E: oauth2, proxy, log, expression, rate-limit policies', () => {
           scopes: ['authorizedScope']
         }
       },
-      policies: ['oauth2', 'proxy', 'log', 'expression', 'rate-limit'],
+      policies: ['jwt', 'oauth2', 'proxy', 'log', 'expression', 'rate-limit'],
```
XVincentX: Same here, why did you need to add this policy?

test/e2e/oauth2-authorization-code.js
ravikp7: The tests were failing
Logs:
```
  606 passing (2m)
  2 pending
  2 failing

  1) oauth2 authorization code grant type
       "before all" hook:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/ravi/oss/express-gateway/test/e2e/oauth2-authorization-code.js)
  

  2) oauth2 authorization code grant type
       "after all" hook:
     TypeError: Cannot read property 'kill' of null
      at Context.done (test/e2e/oauth2-authorization-code.js:131:20)
```
With `LOG_LEVEL=debug`
```
  5) oauth2 authorization code grant type
       "before all" hook:
     Error: connect ECONNREFUSED 127.0.0.1:33635
      at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1097:14)

  6) oauth2 authorization code grant type
       "after all" hook:
     TypeError: Cannot read property 'kill' of null
      at Context.done (test/e2e/oauth2-authorization-co
```
Since, only policies loading was changed, so best guess was to try including some to fix those.
I thought that in the test in proxy `changeOrigin` is set to true, so `cors` need to be included, and others `jwt` because of `oauth2`. And no idea about `basic-auth`, but without it also, tests was breaking.

XVincentX: Ok do not worry about these now, I can try to have a look into later.

test/e2e/policy-seq-oauth2-expression-log-ratelimit-proxy.js
ravikp7: For this one, with debug
```
  605 passing (53s)
  2 pending
  2 failing

  1) E2E: oauth2, proxy, log, expression, rate-limit policies
       "before all" hook: setup:
     can't resolve reference jwt.json from id http://express-gateway.io/schemas/policies/oauth2.json#
  

  2) E2E: oauth2, proxy, log, expression, rate-limit policies
       "after all" hook: cleanup:
     TypeError: Cannot read property 'close' of undefined
      at Context.after (test/e2e/policy-seq-oauth2-expression-lo
```
So added `jwt` policy

lib/gateway/index.js
```diff
@@ -49,11 +44,19 @@ module.exports = function ({ plugins, config } = {}) {
...
+const bootstrapPolicies = ({ app, plugins, config } = {}) => {
+  if (plugins && plugins.policies && plugins.policies.length) {
+    plugins.policies.forEach(policy => {
+      log.debug(`registering policy ${policy.name}`);
+      policies.register(policy);
+    });
+  }
+
+  // Load policies present in config
+  if (config.gatewayConfig.policies) {
```
XVincentX: This field is required in the JSON Schema, so there's no chance it will be undefined. You can remove this check.

lib/gateway/index.js
```diff
@@ -76,23 +79,37 @@ function bootstrap({ plugins, config } = {}) {
...
+        app,
+        plugins: pluginsLoader.load(newConfig),
+        config: newConfig
+      });
+      rootRouter = pipelines.bootstrap({ app: express.Router(), config: newConfig });
+      log.info('hot-reload config completed');
     } catch (err) {
-      log.error('Could not hot-reload gateway.config.yml. Configuration is invalid.', err);
+      log.error('Could not hot-reload gateway.config.yml. Configuration is invalid. ', err);
+      log.debug('Loading old configuration');
```
XVincentX: Let's ditch this.

lib/gateway/index.js
```diff
@@ -76,23 +79,37 @@ function bootstrap({ plugins, config } = {}) {
...
+      bootstrapPolicies({
+        app,
+        plugins: pluginsLoader.load(newConfig),
+        config: newConfig
+      });
+      rootRouter = pipelines.bootstrap({ app: express.Router(), config: newConfig });
+      log.info('hot-reload config completed');
     } catch (err) {
-      log.error('Could not hot-reload gateway.config.yml. Configuration is invalid.', err);
+      log.error('Could not hot-reload gateway.config.yml. Configuration is invalid. ', err);
```
XVincentX:
```suggestion
      log.error(`Could not hot-reload gateway.config.yml. Configuration is invalid. ${err}`);
```

lib/gateway/index.js
```diff
@@ -49,11 +44,19 @@ module.exports = function ({ plugins, config } = {}) {
     });
 };
 
-function bootstrap({ plugins, config } = {}) {
-  const app = express();
-  app.set('x-powered-by', false);
+const bootstrapPolicies = ({ app, plugins, config } = {}) => {
+  if (plugins && plugins.policies && plugins.policies.length) {
+    plugins.policies.forEach(policy => {
+      log.debug(`registering policy ${policy.name}`);
```
XVincentX: Can we add in this statement that this is coming from a plugin?

lib/gateway/index.js
```diff
@@ -49,11 +44,19 @@ module.exports = function ({ plugins, config } = {}) {
     });
 };
 
-function bootstrap({ plugins, config } = {}) {
-  const app = express();
-  app.set('x-powered-by', false);
+const bootstrapPolicies = ({ app, plugins, config } = {}) => {
```
XVincentX: The `app` parameter does not seem to be used. Let's get rid of it.
```suggestion
const bootstrapPolicies = ({ plugins, config } = {}) => {
```

lib/gateway/index.js
```diff
@@ -76,23 +79,37 @@ function bootstrap({ plugins, config } = {}) {
...
+};
+
+function bootstrap({ plugins, config } = {}) {
+  let rootRouter;
+  const app = express();
+  app.set('x-powered-by', false);
   app.use(passport.initialize());
+  bootstrapPolicies({ app, plugins, config });
   rootRouter = pipelines.bootstrap({ app: express.Router(), config });
   app.use((req, res, next) => {
```
XVincentX: I think we can refactor this to be `app.use(rootRouter)`

lib/policies/index.js
```diff
@@ -49,14 +49,25 @@ schemas.register('internal', 'baseAuth', {
...
-  const policy = require(path.resolve(__dirname, name));
-  policies[name] = Object.assign(policy, { name });
-  register(policy);
-});
+const load = (policiesToBeLoaded) => {
+  const policyNames = fs
+    .readdirSync(path.resolve(__dirname))
+    .filter(dir => fs.lstatSync(path.resolve(__dirname, dir)).isDirectory());
+
+  const policiesWithoutPlugins = policiesToBeLoaded.filter((name) => policyNames.includes(name));
```
XVincentX: Let's use a better name

```suggestion
  const corePolicies = policiesToBeLoaded.filter(name => policyNames.includes(name));
```

lib/policies/index.js
```diff
@@ -49,14 +49,25 @@ schemas.register('internal', 'baseAuth', {
...
+
+  policiesWithoutPlugins.forEach((name) => {
```
XVincentX:
```suggestion
  corePolicies.forEach(name => {
```

lib/gateway/index.js
```diff
@@ -76,23 +79,37 @@ function bootstrap({ plugins, config } = {}) {
...
 
   eventBus.on('hot-reload', (hotReloadContext) => {
+    const oldConfig = config;
+    const oldPlugins = plugins;
     const oldRootRouter = rootRouter;
     try {
-      rootRouter = pipelines.bootstrap({ app: express.Router(), config: hotReloadContext.config });
-      log.info('hot-reload router completed');
+      const newConfig = hotReloadContext.config;
+      policies.removeAll();
```
XVincentX: I do not think this is a good idea. What we should be doing here is to check out and load only the *new* policies instead of ditching everything and reload them from scratch. It should not be a big deal, but I believe it's important.

ravikp7: This was added to cater the case when a policy is removed from the config.
If we want to selectively add or remove policies according to the new config, then I think we would have to expose two functions to check if a policy exists and other to remove the policy removed from config.

What do you think about this?

lib/gateway/index.js
ravikp7: `app` parameter is being used to register policy routes

lib/gateway/index.js
ravikp7: sure, will do

lib/gateway/index.js
XVincentX: I'd go in this direction, as long it is not that difficult. The whole idea is to minimize the startup time and the reload time, so it's kind of important to make sure we do not waste time on that.

ravikp7: I agree with your point. I'm now working on to check and then load or remove core policies as well as plugins.

ravikp7:
@XVincentX Do we also need to remove policies which are removed from the config on hot reload?

XVincentX:
@ravikp7 No; they're on memory already so it does not make sense to unload them. In any case I took care of it and I'm adding an e2e test. See the code in case you're interested!

PR #893, opened 2019-03-27
Modified context.js to test if code sent to egContext is JS before executing

### LINKED ISSUE #892: Error with transform-object.js
I've been getting an error trying to pass a new header into my proxied request.

Here's the relevant section of my gateway.config.yml:
```
pipelines:
  default:
    apiEndpoints:
      - kifa
    policies:
      - key-auth:
      - request-transformer:
        - action:
            headers:
              remove: ['authorization']
              add:
                Authorization: Bearer ${GW_KIFA_TOKEN}
      - proxy:
        - action:
            serviceEndpoint: kifa
            changeOrigin: true
```

Here is the error it throws:
```
evalmachine.<anonymous>:1
Bearer eyJhbGciOiJ.etc...
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: Unexpected identifier
  at new Script (vm.js:85:7)
  at createScript (vm.js:266:10)
  at Object.runInNewContext (vm.js:307:10)
  at EgContextBase.run (/Users/daniel/Dev/selfagency-gateway/node_modules/express-gateway/lib/gateway/context.js:32:13)
  at Object.keys.forEach.addParam (/Users/daniel/Dev/selfagency-gateway/node_modules/express-gateway/lib/policies/request-transformer/transform-object.js:3:85)
  at Array.forEach (<anonymous>)
  at module.exports (/Users/daniel/Dev/selfagency-gateway/node_modules/express-gateway/lib/policies/request-transformer/transform-object.js:3:37)
  at /Users/daniel/Dev/selfagency-gateway/node_modules/express-gateway/lib/policies/request-transformer/index.js:14:9
  at router.use (/Users/daniel/Dev/selfagency-gateway/node_modules/express-gateway/lib/gateway/pipelines.js:138:11)
  at Layer.handle [as handle_request] (/Users/daniel/Dev/selfagency-gateway/node_modules/express/lib/router/layer.js:95:5)
```

From what I've been able to deduce through logging, [lib/policies/transform-request/transform-object.js](https://github.com/ExpressGateway/express-gateway/blob/master/lib/policies/request-transformer/transform-object.js) receives:
-  `{ Authorization: 'Bearer eyJhbGciOiJ.etc...' }` as `transformSpecs.add`
-  `egContext`, and
-  `obj`, which contains the following object:
```
{
  authorization: 'apiKey 6LMway.etc...',
  host: 'localhost:8084',
  connection: 'close',
  'user-agent': 'Paw/3.1.8 (Macintosh; OS X/10.14.3) GCDHTTPRequest',
  'eg-consumer-id': '52c503fd-05ad-4f9b-b603-3c14af5e572b'
}
```
It then passes these into:
```
Object.keys(transformSpecs.add).forEach(addParam => { obj[addParam] = egContext.run(transformSpecs.add[addParam]); });
```
After `egContext.run()` is called, and *this is the important part*, `obj` equals:

`Bearer eyJhbGciOiJ.etc...`

So the object being returned is no longer an object, it's a string with only the value of the key/value pair, and that's why the VM can't handle it.

If you remove the `egContext.run()` and just put `obj[addParam] = transformSpecs.add[addParam];`, then it works properly.

What might resolve the problem is perhaps something that evaluates whether the contents of `transformSpecs.add[addParam]` is JavaScript or not before it wraps it in `egContext.run()`, or something within `egContext.run()` itself that checks if it's executable or otherwise returns it as-is.

I tinkered with running [Acorn](https://github.com/acornjs/acorn), which is already a peer dependency, within `EgContextBase.prototype.run` (within [/lib/gateway/context.js](https://github.com/ExpressGateway/express-gateway/blob/master/lib/gateway/context.js)) in order to evaluate whether `code` was JS. This seemed to do the trick:

```
const acorn = require('acorn');
EgContextBase.prototype.run = function (code, ctx) {
  try {
    acorn.parse(code);
  } catch (err) {
    log.error(err);
    return code;
  }
  return vm.runInNewContext(code, ctx || this);
};
```

selfagency:
Resolves #892.

### DIFF
--- a/lib/gateway/context.js
+++ b/lib/gateway/context.js
@@ -30,12 +30,7 @@ EgContextBase.prototype.run = function (code, ctx) {
   // Do not construct new object each time
   // In that case it will have to build new contextified object
   // So subsequent calls will take longer time then agains some already contextified object
-  try {
-    acorn.parse(code);
-  } catch (err) {
-    return code;
-  }
-  return vm.runInNewContext(code, ctx || this);
+  return acorn.parse(code) instanceof Error ? code : vm.runInNewContext(code, ctx || this);
 };
 
 module.exports = EgContextBase;
--- a/lib/gateway/context.js
+++ b/lib/gateway/context.js
@@ -1,3 +1,4 @@
+const acorn = require('acorn');
 const vm = require('vm');
 
 const uuid62 = require('uuid62');
@@ -29,6 +30,11 @@ EgContextBase.prototype.run = function (code, ctx) {
   // Do not construct new object each time
   // In that case it will have to build new contextified object
   // So subsequent calls will take longer time then agains some already contextified object
+  try {
+    acorn.parse(code);
+  } catch (err) {
+    return code;
+  }
   return vm.runInNewContext(code, ctx || this);
 };
 
--- a/lib/gateway/context.js
+++ b/lib/gateway/context.js
@@ -1,4 +1,5 @@
 const acorn = require('acorn');
+const log = require('../logger').gateway;
 const vm = require('vm');
 
 const uuid62 = require('uuid62');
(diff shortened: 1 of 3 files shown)


XVincentX:
See my comment in the issue.

PR #899, opened 2019-04-05
fix multiple creds support for 'jwt' and 'key-auth'

nakardo:
See: https://github.com/ExpressGateway/express-gateway/issues/856

Also introduces changes on the `credentials.test.js` tests to make sure other policies work as intended.

### DIFF
--- a/lib/services/credentials/credential.service.js
+++ b/lib/services/credentials/credential.service.js
@@ -45,24 +45,23 @@ s.insertCredential = function (id, type, credentialDetails) {
     throw new Error(`Invalid credential type: ${type}`); // TODO: replace with validation error
   }
 
-  // TODO: not a good approach, new way TBD
-  const areMultipleCredsAllowed = ['key-auth', 'jwt'].includes(type);
-
-  let credentialExistsPromise;
-  if (areMultipleCredsAllowed) {
-    credentialExistsPromise = dereferencePromise;
-  } else {
-    // check if credential already exists
-    credentialExistsPromise = this.getCredential(id, type)
+  // check if credential already exists
+  const checkSingleCredExistence = () => {
+    return this.getCredential(id, type)
       .then(cred => {
         if (cred && cred.isActive) {
           throw new Error('Credential already exists and is active'); // TODO: replace with validation error
         }
-        return dereferencePromise;
       });
-  }
+  };
 
-  return credentialExistsPromise.then(resolvedSchema => {
+  // TODO: not a good approach, new way TBD
+  const areMultipleCredsAllowed = ['key-auth', 'jwt'].includes(type);
+  const flow = areMultipleCredsAllowed
+    ? dereferencePromise
+    : checkSingleCredExistence().then(() => dereferencePromise);
+
+  return flow.then(resolvedSchema => {
     const credentialConfig = resolvedSchema.properties[type];
     const newCredential = { isActive: 'true' };
     utils.appendCreatedAt(newCredential);
--- a/lib/services/credentials/credential.service.js
+++ b/lib/services/credentials/credential.service.js
@@ -45,84 +45,94 @@ s.insertCredential = function (id, type, credentialDetails) {
     throw new Error(`Invalid credential type: ${type}`); // TODO: replace with validation error
   }
 
-  return this.getCredential(id, type) // check if credential already exists
-    .then((cred) => {
-      if (cred && cred.isActive) {
-        throw new Error('Credential already exists and is active'); // TODO: replace with validation error
-      }
-      return dereferencePromise;
-    })
-    .then((resolvedSchema) => {
-      const credentialConfig = resolvedSchema.properties[type];
-      const newCredential = { isActive: 'true' };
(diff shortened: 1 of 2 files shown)


lib/services/credentials/credential.service.js
```diff
@@ -45,84 +45,94 @@ s.insertCredential = function (id, type, credentialDetails) {
...
-        throw new Error('Credential already exists and is active'); // TODO: replace with validation error
-      }
-      return dereferencePromise;
-    })
-    .then((resolvedSchema) => {
-      const credentialConfig = resolvedSchema.properties[type];
-      const newCredential = { isActive: 'true' };
-      utils.appendCreatedAt(newCredential);
-      utils.appendUpdatedAt(newCredential);
+  // TODO: not a good approach, new way TBD
```
XVincentX: I do not understand the meaning of this sentence.

lib/services/credentials/credential.service.js
```diff
@@ -45,84 +45,94 @@ s.insertCredential = function (id, type, credentialDetails) {
...
+  const areMultipleCredsAllowed = ['key-auth', 'jwt'].includes(type);
+
+  let credentialExistsPromise;
+  if (areMultipleCredsAllowed) {
+    credentialExistsPromise = Promise.resolve(dereferencePromise);
```
XVincentX: If this is a Promise there's no need to wrap it again
```suggestion
    credentialExistsPromise = dereferencePromise;
```

lib/services/credentials/credential.service.js
```diff
@@ -45,84 +45,94 @@ s.insertCredential = function (id, type, credentialDetails) {
...
-            newCredential.scopes = JSON.parse(newCredential.scopes);
-          }
-
-          if (typeof newCredential.isActive === 'string') {
-            newCredential.isActive = newCredential.isActive === 'true';
-          }
-          return newCredential;
-        });
-      }
+  return credentialExistsPromise.then((resolvedSchema) => {
```
XVincentX: If it's a single param we can remove parenthesis
```suggestion
  return credentialExistsPromise.then(resolvedSchema => {
```

test/services/credential-service/credentials.test.js
```diff
@@ -6,97 +6,120 @@ const credentialService = services.credential;
...
-    let credential;
-    const username = 'someUser';
+describe('Credential tests', () => {
+  const username = 'someUser';
+  const credential = {
+    secret: 'password'
+  };
 
-    before(() => db.flushdb());
+  function insertCredential(policy) {
```
XVincentX: Well technically it's not a policy, it's a credential type
```suggestion
  function insertCredential(type) {
```

test/services/credential-service/credentials.test.js
```diff
@@ -6,97 +6,120 @@ const credentialService = services.credential;
...
+    return credentialService.insertCredential(username, policy, credential);
```
XVincentX:
```suggestion
    return credentialService.insertCredential(username, type, credential);
```

lib/services/credentials/credential.service.js
```diff
@@ -45,84 +45,94 @@ s.insertCredential = function (id, type, credentialDetails) {
...
-    .then((resolvedSchema) => {
-      const credentialConfig = resolvedSchema.properties[type];
-      const newCredential = { isActive: 'true' };
-      utils.appendCreatedAt(newCredential);
-      utils.appendUpdatedAt(newCredential);
+  // TODO: not a good approach, new way TBD
+  const areMultipleCredsAllowed = ['key-auth', 'jwt'].includes(type);
+
+  let credentialExistsPromise;
+  if (areMultipleCredsAllowed) {
```
XVincentX: I think we can restructure the flow to be a little bit cleaner.

We could extract the checking function so that we can do (pseudo code)

```js
const checkSingleCredExistence = () => {};

const flow = areMultipleCredsAllowed ? 
  dereferencePromise.then(checkSingleCredExistence) : 
  dereferencePromise 
```

lib/services/credentials/credential.service.js
XVincentX: Ah it's actually not your fault :)

lib/services/credentials/credential.service.js
nakardo: @XVincentX do you mean something like this?

  ```javascript
  // check if credential already exists
  const checkSingleCredExistence = () => {
    return this.getCredential(id, type)
      .then(cred => {
        if (cred && cred.isActive) {
          throw new Error('Credential already exists and is active'); // TODO: replace with validation error
        }
      });
  };

  // TODO: not a good approach, new way TBD
  const areMultipleCredsAllowed = ['key-auth', 'jwt'].includes(type);
  const flow = areMultipleCredsAllowed ?
    dereferencePromise :
    checkSingleCredExistence().then(() => dereferencePromise);

  return flow.then(resolvedSchema => {
    ...
  });
  ```

XVincentX: Yeah exactly!

PR #901, opened 2019-04-05
Assignable user and application IDs

kouponmedia:
Updated validateAndCreateUser & validateAndCreateApp in user & application services. Changed order of object assign to allow id property to be used if present instead of generating a new uuid.

In my case, I need to integrate with an existing PKI where client_id/secret already exist for our customers. I want to be able to assign IDs to the client application instead of generating new IDs. See the following example:
```
{
  "$id": "http://express-gateway.io/models/applications.json",
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    "name": {
      "type": "string"
    }
  },
  "required": [
    "id",
    "name"
  ]
}
```
`eg apps create -u <user> -p "name=<name>" -p "id=<id>"`

### DIFF
--- a/lib/services/consumers/application.service.js
+++ b/lib/services/consumers/application.service.js
@@ -175,7 +175,7 @@ function validateAndCreateApp (appProperties, userId) {
 
       const baseAppProps = { isActive: 'true', id: uuidv4(), userId };
 
-      const app = Object.assign(appProperties, baseAppProps);
+      const app = Object.assign(baseAppProps, appProperties);
 
       utils.appendCreatedAt(app);
       utils.appendUpdatedAt(app);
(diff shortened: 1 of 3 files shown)


kouponmedia:
It wasn't obvious to me at first, but by switching the order of args in the Object.assign in validateAndCreateUser, the original passed in object is not being modified anymore and instead a new pointer is returned. This affected the test because the test tries to use the original pointer instead of one returned from the insert promise. This didn't affect actual behavior because the insert functions in both the user and application services always use objects returned from promises.

This does smell a bit because now a developer has to know that we're not just decorating pointers and instead we have to use the result of a promise because the pointer may have changed. Changing the order of Object.assign is not the only way this could be achieved. I really just want to be able to override the ID. Something like this would work too.

```
const baseUserProps = { isActive: 'true', username: _user.username, id: newUser.id || uuidv4() };
if (newUser) {
        user = Object.assign(newUser, baseUserProps); // return the original pointer
} else user = baseUserProps;
```

Thoughts?

XVincentX:
Ok, that makes sense.

PR #905, opened 2019-04-08
Test: test for proxy policy "hostRewrite" option

Andycharalambous:
currently "target" option is an object, when http-proxy expects it to be a string


Andycharalambous:
Have added a test for hostRewrite option on proxy policy - fails due to problem in node-http-proxy library dependency. Shall I leave this PR open for now, until the problem in node-http-proxy is resolved?

XVincentX:
Close until we figure it out.

PR #930, opened 2019-07-12
fix: ".return" is not a function of "db.del"

nvanheuverzwijn:
Whenever the token expiration code run, this error pops up.
```
{"error":"server_error","error_description":"this.remove(...).return is not a function"}
```


XVincentX:
Thanks @nvanheuverzwijn. It's really this part isn't properly tested; would you be open to add a minimal test for this functionality as well?

nvanheuverzwijn:
@XVincentX Just added a test. Tell me if this is what you have in mind.

XVincentX:
Thanks for this fix, I'll release a version now :)

Issue #941, opened 2019-09-03
[Question] OAuth 2 Introspection in 2 different pipelines


tagkiller:
Hi all,

I have an issue using OAuth 2 Introspection accross multiple pipelines.
So my first question is it even possible to use it that way ?
I configured 2 `apiEndpoints` and wants each one to use a specific `oauth2-introspect` endpoint. The thing it's whether I enter the first or the second pipeline, it's always the last `oauth2-introspect` endpoint configured that's hitting.

```
http:
  port: 8888
admin:
  port: 9876
  host: 0.0.0.0
apiEndpoints:
  managementAPI:
    host: localhost
    paths: '/a/management/*'
  usersAPI:
    host: localhost
    paths: '/a/*'
serviceEndpoints:
  ProvisioningService:
    url: 'http://provisioning.service:8080'
policies:
  - proxy
  - rewrite
  - request-transformer
  - oauth2-introspect
pipelines:
  usersservice:
    apiEndpoints:
      - usersAPI
    policies:
      - oauth2-introspect:
        - action:
            endpoint: 'http://security.service:8080/introspect'
            authorization_value: 'Bearer someToken'
            ttl: 60
      - request-transformer:
        - action:
            headers:
              add:
                x-auth-principal: "req.user.sub"
                x-auth-scope: "req.user.scope"
      - proxy:
          - action:
              serviceEndpoint: ProvisioningService
              changeOrigin: true
              stripPath: true
              proxyTimeout: 1000
  managementservice:
    apiEndpoints:
      - managementAPI
    policies:
      - oauth2-introspect:
        - action:
            endpoint: 'https://management.service:8080/introspect'
            ttl: 60
      - request-transformer:
        - action:
            headers:
              add:
                x-auth-principal: "req.user.sub"
                x-auth-scope: "req.user.scope"
      - rewrite:
        - condition:
            name: pathmatch
            match: /a/management/:code # Express Path
          action:
            rewrite: /a/:code
      - proxy:
          - action:
              serviceEndpoint: ProvisioningService
              changeOrigin: true
              stripPath: true
              proxyTimeout: 1000
``` 

Could you please help me with this matter ? Am I doing something wrong or is it a bug ?

Thanks in advance.

Best regards,

XVincentX:
I've just tried this configuration and I've correctly hit the two proposed endpoints. Can you provide a screenshot or something to investigate?

Feel free to reopen

tagkiller:
[...]

PR #942, opened 2019-09-05
OAuth2-Introspect plugin was always using the same passport strategy

tagkiller:
OAuth2 Introspection was hitting always the latest endpoint configured, in whichever pipeline.
Please have a look at [this](https://github.com/ExpressGateway/express-gateway/issues/941) for more info.
@XVincentX here is the related PR to the matter discussed in the above referenced link :-)


lib/policies/oauth2-introspect/oauth2-introspect.js
```diff
@@ -7,7 +7,7 @@ module.exports = function (actionParams) {
   const scannedTokens = [];
   const tokenIntrospection = tokenIntrospectionGenerator(actionParams);
 
-  passport.use('bearer-introspect', new BearerStrategy({ passReqToCallback: true }, (req, accessToken, done) => {
+  passport.use(`bearer-introspect_${actionParams.endpoint}`, new BearerStrategy({ passReqToCallback: true }, (req, accessToken, done) => {
```
XVincentX: For consistency, we might want to use the same we did for the `jwt` policy

`const strategyName = `jwt-${uuid()}`;`

PR #949, opened 2019-09-25
Adding `selfHandleResponse` option parameter to Proxy policy

josemf:
Adding `selfHandleResponse` and sending the option to `node-http-proxy` is important to enable processing responses from proxied backends.

Usually `node-http-proxy` pipes the proxied response through existing response stream. This can be disabled by `selfHandleResponse`. I supply an implementation suggestion for sending the proxied response through express `res.send()`.

Plugins might monkey patch this method to effectively intercept and handle express gateway responses. As I do in: https://github.com/josemf/express-gateway-plugin-errors.

### DIFF
--- a/lib/policies/proxy/index.js
+++ b/lib/policies/proxy/index.js
@@ -80,6 +80,11 @@ module.exports = {
         type: 'boolean',
         default: false,
         description: 'Adds x-forward headers to the proxied request'
+      },
+      selfHandleResponse: {
+        type: 'boolean',
+        default: false,
+        description: 'Self handles proxy response by sending headers and data through express'            
       }
     },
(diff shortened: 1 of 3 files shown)


XVincentX:
I'm sorry I haven't been able to jump on this yet.

josemf:
@XVincentX that's ok,

I'm not even sure this is the best way do implement this, I mean we could just pass the selfHandleResponse to the proxy and expose `http-proxy` instance so a specific implementation (ex:  https://github.com/josemf/express-gateway-plugin-errors) could hook to it and decide how responses should be handled.

Take your time, I just think `express-gateway` could offer some options for proxy data response handling.

🙏

XVincentX:
I'm jumping on this again now — I completely lost the context :(

Issue #827, opened 2018-09-28
proxy.js for websocket


zzzONzzz:
proxy.web(req, res, { target, headers, buffer: req.egContext.requestStream });

maybe modify

if (req.headers.upgrade && req.headers.upgrade == "websocket") {
      if (params.ws == true) {
        proxy.ws(req, req.socket, { target, headers, buffer: req.egContext.requestStream });
      }
      else {
        res.end();
      }
    }
    else {
      proxy.web(req, res, { target, headers, buffer: req.egContext.requestStream });
    }

XVincentX:
I wish that'd be so easy.

Unfortunately adding proxy support is something we'd need to do on server level — and it's a little bit difficult to make it happen on endpoint level. We'll get there, but we need to plan this accordingly.

Closing for now.

PR #954, opened 2019-10-11
Suggested change by @zzzONzzz

wywfalcon:
This was a [suggestion](https://github.com/ExpressGateway/express-gateway/issues/827) by @zzzONzzz

It seems to work just fine. I tested with both `wss` and `ws` connections.


XVincentX:
I wish that'd be so easy.

Unfortunately adding proxy support is something we'd need to do on server level — and it's a little bit difficult to make it happen on endpoint level. We'll get there, but we need to plan this accordingly.

Closing for now.

PR #965, opened 2019-11-19
Fix stripPath to stop it URL-decoding the path

### LINKED ISSUE #963: proxy stripPath incorrectly URL decodes the path
When using `stripPath` on a proxy policy, the path passed on to the proxy is URL decoded. This can cause the incorrect URL to be requested from the proxy. An example of when maintaining the URL encoding of path components is important is in the Gitlab API where requesting a project file takes the form `https://gitlab.example.com/api/v4/projects/backend-services%2Fauth` (projects are namespaced with a URL-encoded forward slash).

This seems to be caused by using `req.params[0]` in `stripPathFn` in `proxy.js`. It should instead strip the matched path from the beginning of `req.url` to ensure the encoding isn't messed with by express.

Example `gateway.config.yml` follows, run with `LOG_LEVEL=debug` and note the difference in the logged message when requesting `http://localhost:8080/test-path-encoding/path%2Fto%2Fresource` vs `http://localhost:8080/test-strip-path-encoding/path%2Fto%2Fresource`.
To avoid confusion please ignore the httpbin response given by this example, as httpbin has a similar issue in that it always decodes the `url` in the `/anything` response.

```
http:
  port: 8080
apiEndpoints:
  test_path_nostrip:
    host: localhost
    paths: '/test-path-encoding/*'
  test_path_strip:
    host: localhost
    paths: '/test-strip-path-encoding/*'
serviceEndpoints:
  httpbin:
    url: 'http://httpbin.org/anything'
policies:
  - proxy
pipelines:
  nostripPath:
    apiEndpoints:
      - test_path_nostrip
    policies:
      - proxy:
          - action:
              serviceEndpoint: httpbin 
  stripPath:
    apiEndpoints:
      - test_path_strip
    policies:
      - proxy:
          - action:
              serviceEndpoint: httpbin 
              stripPath: true
```

judev:
Fixes #963 

Instead of using `req.params[0]` it now strips the `originalUrl` up to the position of the first wildcard placeholder.

For example, when matching `/test/*` with `/test/foo%2Fbar`
    - Before this change, stripped URL is `foo/bar`
    - After this change, stripped URL is `/foo%2Fbar`

This also fixes a situation that could cause the incorrect path to be proxied onwards if there are multiple wildcards.

For example, when matching `/test/:foo/then/:bar` with `/test/foovar/then/barval`
    - Before this change, stripped URL is `fooval`
    - After this change, stripped URL is `/fooval/then/barval`

### DIFF
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -88,7 +88,16 @@ module.exports = function (params, config) {
 
   const stripPathFn = params.stripPath
     ? req => {
-      req.url = `${req.params[0] || ''}${req._parsedUrl.search || ''}`;
+      const wildcardPos = req.route.path.search(/[:*]/);
+      if (wildcardPos === -1) {
+        // no wildcard, strip the full path. i.e. change URL to / + query string
+        req.url = `/${req._parsedUrl.search || ''}`;
+      } else if (wildcardPos === 0) {
+        // full path is a wildcard match so there's nothing to strip
+      } else {
+        // else strip everything up to the first wildcard, ensuring path still begins with /
+        req.url = `/${req.originalUrl.substring(wildcardPos)}`;
+      }
     } : () => { };
 
   return function proxyHandler(req, res) {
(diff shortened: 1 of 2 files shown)


XVincentX:
Nice stuff, I'll review.

Issue #936, opened 2019-08-26
"eg gateway create" uses Express Gateway last version, not current installed version


mmadaria:
When executing "eg gateway create" command, the app generates an instance of Express Gateway using the last stored version in the public NPM repositories, and doesn't use the same version of the local installed app.

For example, if the installed version of Express Gateway used is 1.16.3 and the "eg gateway create" command is executed, internally it will execute "npm install express-gateway" when generating a new instance, so it's not considering the current installed version (e.g. 1.16.3), and it will install a newer and different version (e.g. 1.16.7). 

This logic can be found in "bin/generators/gateway/create.js" file:

  _install () {
    **this.npmInstall(['express-gateway'], { save: true });**
  }_ 

Note: when the instance is created, you can check the final installed version inside "package.json" file.

PR #971, opened 2020-01-02
fix: install the current installed version of while executing "eg gateway create"

MohammadWasi:
Fixes: #936 
when you execute the 'eg gateway command', this lets you install the currently installed version of express-gateway.

### DIFF
--- a/bin/generators/gateway/create.js
+++ b/bin/generators/gateway/create.js
@@ -128,7 +128,8 @@ module.exports = class extends eg.Generator {
   }
 
   install () {
-    this.npmInstall(['express-gateway'], { save: true });
+    const { version: installedVersion } = require('../../../package.json');
+    this.npmInstall([`express-gateway@${installedVersion}`], { save: true });
   }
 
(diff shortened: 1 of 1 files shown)


bin/generators/gateway/create.js
```diff
@@ -128,7 +128,8 @@ module.exports = class extends eg.Generator {
   }
 
   install () {
-    this.npmInstall(['express-gateway'], { save: true });
+    const { version: installedVersion } = require('../../../package.json');
```
XVincentX: How did you come up with this path? Did you test it?

MohammadWasi: @XVincentX   Yes, I have tested it. So I thought of picking the version from the package.json and pass it to express gateway so that the same version gets installed when generating the new instance of Express Gateway.

XVincentX: Right! My only concern is how to make sure that the path is effectively correct — I'll pull down the branch and check it out.

Issue #979, opened 2020-02-07
Bump express-rate-limit to 5.0.0


ioggstream:
### I expect

express-gateway to work with express-rate-limit to 5.0.0

### Instead

If I upgrade express-rate-limit, it does not work.

@sanyamdogra

sanyamdogra:
Yup, the schema needs to be updated by removing two deprecated functions `delayMs` and `delayAfter`.

sanyamdogra:
I would have made the PR with updated dependency and removing deprecated functions but just waiting for this [PR](https://github.com/nfriedly/express-rate-limit/pull/169) to merge so that the PR makes sense.

XVincentX:
@sanyamdogra Not too fast my friend — that would be a breaking change. We are going to need to keep such parameters and write a layer that would still use them somehow.

sanyamdogra:
@XVincentX Yes, you are right but how to should we go about it? If we update `express-rate-limit` to the new version, we'll lose the functions, if we don't we lose the headers.

XVincentX:
I believe that part of features has just been moved to another package; you'd need to install that into express gateway as well and use it.

sanyamdogra:
Yes, it's been moved to `express-slow-down`.

XVincentX:
All right, let’s melt them together.

sanyamdogra:
Yes sure, let's wait for [this](https://github.com/nfriedly/express-rate-limit/pull/169) to merge

kenime:
Can I suggest 2 things:

1) Support more parameters from the underlying module (e.g. `skipSuccessfulRequests`)
2) move slow-down to a new policy (while supporting rate-limit -> delayAfter / delayMs parameters for backward compatibility)

The reason of both suggestions is because for rate limit, the main objective is to prevent unfairness / service plan limitation, hence it should counted per person / IP without skipping successful nor failed requests.

For slow down, my usage is to prevent sudden burst of traffic causing underlying services not able to responding and result in 502 bad gateway. In such case, I would want to count **concurrently processing** calls instead of total call per API path. I can achieve this by using `req.baseUrl + req.path` as the key and enabling `skipSuccessfulRequests` and `skipFailedRequests` to decrement count when a request is completed.

With above suggestion, I could achieve my goals by setting:
```
- rate-limit:
        -
          action:                         # allow
            max: 100                       # max 100 request
            windowMs: 60000              # per 60 seconds
[...]

PR #981, opened 2020-02-14
Addition of Rate Limit standardization headers

sanyamdogra:
For #979 
* [x] Updated `express-rate-limit` to 5.1.1
* [x] Added `express-slow-down` 
* [ ] Add support for `delayMs` and `delayAfter` to not break stuff.
* [ ] Finally, make it work! 

cc @ioggstream


lib/policies/rate-limit/rate-limit.js
```diff
@@ -1,4 +1,5 @@
 const RateLimit = require('express-rate-limit');
+// const SlowDown = require('express-slow-down');
```
ioggstream: If we don't need this line, just remove it.

sanyamdogra: It's work in progress, we'll definitely need this line. `express-slow-down` contains the deprecated functions.

ioggstream:
@sanyamdogra try to propose a couple of solutions to this problem. Take *some time* to write down your ideas and expose them to Nathan, which has a deep knowledge of the plugins.

sanyamdogra:
> @sanyamdogra try to propose a couple of solutions to this problem. Take _some time_ to write down your ideas and expose them to Nathan, which has a deep knowledge of the plugins.

Yes, let me do some research and discuss it with him. Thanks!

PR #986, opened 2020-02-27
Implemented the passport-jwt secretOrKeyProvider

itsrennyman:
Hello, hope you are well :) 

In this PR i've implemented the secretOrKeyProvider method of the passport-jwt library.

The problem is, if i have many users, everyone with a jwt credential associated, is impossible to make requests, because jwt policy permit to set only one secret, or only one key file for all the requests, so if i sign every jwt with its secret, the gateway respond unauthorized.

This implementation permit to the jwt policy to check the secret for each user with the JWT credentials set.

So, now if there aren't set the params secretOrPublicKey and the secretOrPublicKeyFile, this check is triggered.

Renny.

### DIFF
--- a/lib/policies/jwt/index.js
+++ b/lib/policies/jwt/index.js
@@ -40,7 +40,7 @@ module.exports = {
         description: 'Value istructing the gateway whether verify the sub against the internal SOC'
       }
     },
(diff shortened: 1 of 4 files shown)


rennypoz:
Updated also test cases for this PR.

PR #999, opened 2020-07-15
Added algorithm option to jwt policy

Brady-Conn:
Hello,
Express-gateway is fantastic — it has become the backbone of our team’s api architecture. That being said, we have run into one small issue.
Problem: The JWT policy does not currently accept algorithms as an option to specify which algorithms can be used to verify tokens. Under the hood I noticed express-gateway uses `passport-jwt`, which uses `jsonwebtoken`. This is the same package we use. Both `passport-jwt` and `jsonwebtoken` support specifying algorithms. This is a good feature because otherwise jsonwebtoken will allow any algorithm to be used depending on what is decoded via the JWT header. Specifying algorithms allows enforcing of standards across our apis.
Solution: Expose the algorithms option in the JWT policy.
Here is the pr for the docs update: https://github.com/ExpressGateway/express-gateway.io/pull/336

### DIFF
--- a/lib/policies/jwt/index.js
+++ b/lib/policies/jwt/index.js
@@ -41,6 +41,7 @@ module.exports = {
       },
       algorithms: {
         type: 'array',
+        items: { type: 'string', enum: ['HS256', 'HS384', 'HS512', 'RS256', 'RS384', 'RS512'] },
         description: 'If defined, limits valid jwts to specified algorithms'
       }
     },
--- a/lib/policies/jwt/index.js
+++ b/lib/policies/jwt/index.js
@@ -38,6 +38,10 @@ module.exports = {
         type: 'boolean',
         default: true,
         description: 'Value istructing the gateway whether verify the sub against the internal SOC'
+      },
+      algorithms: {
+        type: 'array',
+        description: 'If defined, limits valid jwts to specified algorithms'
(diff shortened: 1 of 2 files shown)


Brady-Conn:
The tests failing for me failed once(with the same timeout error) locally but have since passed. Is there a way to re-try these pipelines?

XVincentX:
I can take care of this and make sure all tests pass. Can you rebase with latest master?

Brady-Conn:
Awesome, thank you. I just pushed up that rebase, let me know if there is anything else I can do.

lib/policies/jwt/index.js
```diff
@@ -38,6 +38,10 @@ module.exports = {
         type: 'boolean',
         default: true,
         description: 'Value istructing the gateway whether verify the sub against the internal SOC'
+      },
+      algorithms: {
+        type: 'array',
```
XVincentX, line 43: I think we should probably specify the `items` to be a string and then enum the possible choice. JWT has a fixed set of algorithms

Brady-Conn: That makes sense, I just pushed up a change to reflect that. I wasn't 100% sure about the syntax so let me know if I need to modify it at all.

PR #1002, opened 2020-07-17
OAuth2 introspection cannot identify 'scope' in token

BrunoVillanova:
OAuth2 Introspection policy was only expecting a 'scopes' in OAuth2 provider token response.
Some OAuth2 provider implementations use 'scope' field name causing the token scopes verification to fail.
I've added the possibility to work with OAuth2 providers that use both 'scope' and 'scopes' as name of the scopes field token.

### DIFF
--- a/lib/policies/oauth2-introspect/oauth2-introspect.js
+++ b/lib/policies/oauth2-introspect/oauth2-introspect.js
@@ -13,7 +13,8 @@ module.exports = function (actionParams) {
     const requestedScopes = req.egContext.apiEndpoint.scopes;
 
     const scopeCheck = (tokenData, done) => {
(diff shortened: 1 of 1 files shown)


XVincentX:
This is a tricky one, I'm going to have to look into this before merging.

As far as I remember, the standard is clear about the field name, that is why the field is only one and not both.

BrunoVillanova:
Hi Vicent,

I took a look and the doc says the correct is ‘scope’, please take a look: https://tools.ietf.org/html/rfc6749#section-3.3.

Thank you.

XVincentX:
Looks legit. Thanks for spotting this.

BrunoVillanova:
It was a pleasure to contribute.
Thank you for merging it.

PR #1012, opened 2021-01-06
Expose response body to response transformer policy.

73rhodes:
- Makes the reponse body accessible as egContext.res.body.
  - Supports setting optionally nested properties in the transformed response body.

Hello maintainers and thanks for taking the time to review this. We needed to be able to transform responses not only by adding and removing properties on the response body, but also by using properties in the original response body and modifying their original values. A simple way to do that was exposing the original response body in the egContext so that it can be referenced in the expressions of the responseTransformer actions. I also wanted to be able to set *nested* properties in the response body.

If you think this runs against the design intent of the original req/res modifier, it would be greatly appreciated if you could suggest an alternative approach to modifying properties within the original response body.

Note that this change only exposes the response body only, and not the request body, but it could be updated to support both if wanted.

### DIFF
--- a/test/policies/modifier/modifier.test.js
+++ b/test/policies/modifier/modifier.test.js
@@ -48,6 +48,7 @@ describe('@modifier policy', () => {
         should(res.header).not.have.property('x-test');
 
         should(res.body).have.property('hello', 'world');
+        should(res.body.foo).have.property('url', '/');
         should(res.header).have.property('r-test', 'baffino');
         should(res.header).have.property('res', 'correct');
       });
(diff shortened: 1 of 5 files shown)


test/policies/modifier/modifier.test.js
```diff
@@ -118,6 +120,14 @@ const setupGateway = () => {
                   remove: ['x-test']
                 }
               }
+            }, {
+              action: {
+                body: {
+                  add: {
+                    'foo.url': 'res.body.url'
```
73rhodes, line 127: This is demonstrating that a new nested property can be set on the body, using some property of the original response body.  Even though there is a responseTransformer action above that removes the `url` property from the response body, we can access the original response body here and use the value of the original `url` property to set a new `foo.url` property in the transformed response body.

XVincentX:
Hey @73rhodes, thank you very much for the PR.

Unfortunately the project has been halted and I currently have nor the time nor plans to resume it.

The time to resemble the context of all this would be significant and I do prefer to leave the project as it is right now.

I would suggest you to keep your fork and go from there.

PR #1024, opened 2021-05-30
Update dependencies

djordje:
Used following procedure:

- `npm outdated` - check outdated and audit data
- `npm update`- to update to newest version satisfying package file
- `npm audit` - there are two minor reports, but medium, high and critical reports have been fixed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,130 +1,13272 @@
 {
   "name": "express-gateway",
   "version": "1.16.10",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "1.16.10",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "ajv": "^6.10.2",
+        "ajv-keywords": "^3.4.1",
+        "bcryptjs": "^2.4.3",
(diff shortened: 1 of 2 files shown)


XVincentX:
I do not see the package.json file being updated — which is really the one that matters. Can you verify?

djordje:
@XVincentX It was `package-lock.json` that has been updated. According to my understanding package doesn't lock versions since you have `^`. It's in "files changed" tab, but not visible since diff is big.

XVincentX:
In such case, this PR is useless. The package-lock.json file does not get uploaded in the npm package anyway, so the user would not be able to see any difference.

I'm still willing to accept it, but I do see a number of new dependencies because of duplicated. Try to run `npm dedupe` and see if the size gets reduced.

djordje:
> In such case, this PR is useless. The package-lock.json file does not get uploaded in the npm package anyway, so the user would not be able to see any difference.
> 
> I'm still willing to accept it, but I do see a number of new dependencies because of duplicated. Try to run `npm dedupe` and see if the size gets reduced.

You're actually right. Lock doesn't get to the registry... Might be better to start fixing versions in package file. Main issue I have might be collision between express-gateway and other packages dependencies. I'll try to track it down.

BTW `dedupe` is done.

PR #1045, opened 2022-08-25
Version4.0.0

bhouser:
Remove the following policies:
- basic-auth
- jwt
- key-auth
- oauth2
- oauth2-introspect


bhouser:
This PR opened in error, not intended for this repo.

remvee:
Hi @bhouser, I see you're also maintaining a stripped down version of EG.  Have you considered using [jomco/express-gateway-lite](https://github.com/jomco/express-gateway-lite)? If you did, why did it not qualify for your use-case?  It seems we're trying to solve the same problem.  Cheers, Remco

PR #1068, opened 2023-10-08
Add retry mechanism

1204552371:
When the gateway proxy proxy sends a request error, there should be a retry mechanism. The same is true for forwarding including load balancing.
In this way, error 502 caused by socket hang up and connect ECONNREFUSED can be avoided.

maxAutoRetries can be configured in gateway.config.yml. The default is 1. 
such as:
```
      - proxy:
          - action:
              serviceEndpoint: lbsmart
              maxAutoRetries: 1
              changeOrigin: true
```

### DIFF
--- a/lib/policies/proxy/proxy.js
+++ b/lib/policies/proxy/proxy.js
@@ -61,11 +61,29 @@ module.exports = function (params, config) {
     proxyOptions.agent = new ProxyAgent(intermediateProxyUrl);
   }
 
+  const retryCountMap = {};
   const proxy = httpProxy.createProxyServer(Object.assign(params, proxyOptions));
+  //when http request suucess,clear map
+  proxy.on('end', (req,res)=>{
+    delete retryCountMap[req.egContext.requestID];
+  });
 
+
+  //retry function
+  const maxAutoRetries = params.maxAutoRetries || 1;
   proxy.on('error', (err, req, res) => {
     logger.warn(err);
 
+    const requestID = req.egContext.requestID;
+    let retryCount = retryCountMap[requestID] || 0;
+    if(retryCount < maxAutoRetries){
+      retryCount++;
+      retryCountMap[requestID] = retryCount;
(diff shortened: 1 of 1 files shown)


schizmeatarg (commented):
[EXPRESS/.ROOT](https://github.com/schizmeatarg/gateway/commit/1f9521a2ac49781aeb64a4a554340ce0b5471ea0)

lib/policies/proxy/proxy.js
```diff
@@ -61,11 +61,28 @@ module.exports = function (params, config) {
     proxyOptions.agent = new ProxyAgent(intermediateProxyUrl);
   }
 
+  const retryCountMap = {};
   const proxy = httpProxy.createProxyServer(Object.assign(params, proxyOptions));
+  // when http request suucess,clear map
```
diegobejardelaguila: Delete comments

lib/policies/proxy/proxy.js
```diff
@@ -61,11 +61,28 @@ module.exports = function (params, config) {
...
+  const maxAutoRetries = params.maxAutoRetries || 1;
   proxy.on('error', (err, req, res) => {
     logger.warn(err);
 
+    const requestID = req.egContext.requestID;
+    let retryCount = retryCountMap[requestID] || 0;
+    if (retryCount < maxAutoRetries) {
+      retryCount++;
+      retryCountMap[requestID] = retryCount;
+      // console.log(`retry count: ${retryCount} `, req.originalUrl);
```
diegobejardelaguila: Delete console.logs

lib/policies/proxy/proxy.js
1204552371: Okay, I will delete it

lib/policies/proxy/proxy.js
1204552371: Okay, I will delete it

1204552371:
I checked and found that the test error was due to the resources in the test case being currently inaccessible

